PR #57, opened 2020-07-01
Make the source file license annotations match LICENSE.agpl

natoscott:
Update the references in source files which refer to the GPLv3,
whereas the LICENSE.agpl file refers to this source as licensed
under AGPLv3.

@yossigo thank you for dual-licensing this code with AGPLv3!


yossigo:
@natoscott thank you!

PR #67, opened 2020-10-01
Add a create-cluster utility script for testing.

yossigo:
This tool is based on a the `utils/create-cluster` script that ships with Redis and serves the same purpose with Redis Cluster.


utils/create-cluster/create-cluster
```diff
@@ -0,0 +1,163 @@
...
+then
+    tail -f *.log
+    exit 0
+fi
+
+if [ "$1" == "call" ]
+then
+    while [ $((PORT < ENDPORT)) != "0" ]; do
+        PORT=$((PORT+1))
+        ${REDIS_PATH}/redis-cli -p $PORT $2 $3 $4 $5 $6 $7 $8 $9
```
gavrie, line 117: Why not use `shift` and then `$@`?

yossigo: This is copied from Redis, I had the same thought actually but ended up leaving it as is. I'll do that to get rid of the arguments count limit.

gavrie:
[foo](https://trello.com/c/ofgtASl2/4-foo)

gavrie (approved):

PR #68, opened 2020-10-01
Tests: remove clutter.

yossigo:
Log temporary files are now created in tests/tmp.

Every launched instance gets a GUID and a dedicated directory, so the
keepfiles trick is no longer needed when preserving data files (we
simply leave the dirs as-is).


gavrie (approved):
LGTM

PR #70, opened 2020-10-07
Fix #2: Check node id uniqueness on join

### LINKED ISSUE #2: Adding a node with the same id as a previously existing node should be forbidden
After creating a cluster with 2 nodes,
then running "RAFT.NODE remove 2"
Then running another server (with new log) and uid 2, when trying to run "RAFT.CLUSTER JOIN localhost:5001" the new server crashes with error:
"Assertion failed: (node), function raft_handle_append_cfg_change, file src/raft_server.c, line 1227."

gavrie:
- We keep a list of all node ids that were ever members of the cluster
- When adding a node, this list is checked
- If the node id has ever been used, the node will not be joined

### DIFF
--- a/node.c
+++ b/node.c
@@ -329,14 +329,16 @@ void handleNodeAddResponse(redisAsyncContext *c, void *r, void *privdata)
     } else if (reply->type != REDIS_REPLY_ARRAY || reply->elements != 2) {
         LOG_ERROR("RAFT.NODE ADD invalid reply.\n");
     } else {
-        LOG_INFO("Joined Raft cluster, node id: %lu, dbid: %.*s\n",
-                reply->element[0]->integer,
+        raft_node_id_t node_id = reply->element[0]->integer;
+
+        LOG_INFO("Joined Raft cluster, node id: %u, dbid: %.*s\n",
+                node_id,
                 reply->element[1]->len, reply->element[1]->str);
 
         strncpy(rr->snapshot_info.dbid, reply->element[1]->str, reply->element[1]->len);
         rr->snapshot_info.dbid[RAFT_DBID_LEN] = '\0';
 
-        rr->config->id = reply->element[0]->integer;
+        rr->config->id = node_id;
 
         HandleClusterJoinCompleted(rr);
     }
--- a/raft.c
+++ b/raft.c
@@ -557,7 +557,8 @@ static int raftNodeHasSufficientLogs(raft_server_t *raft, void *user_data, raft_
 void raftNotifyMembershipEvent(raft_server_t *raft, void *user_data, raft_node_t *raft_node,
         raft_entry_t *entry, raft_membership_e type)
 {
-    RaftCfgChange *cfgchange;
+    RedisRaftCtx *rr = (RedisRaftCtx *) user_data;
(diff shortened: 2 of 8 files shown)


tests/integration/test_membership.py
```diff
@@ -135,6 +140,29 @@ def test_full_cluster_remove(cluster):
         assert cluster.node(node_id).raft_info()['state'] == 'uninitialized'
 
 
+def test_remove_and_rejoin_node_with_same_id_fails(cluster):
```
yossigo, line 143: Full coverage here should consider the different persistence states, i.e. make sure we remember used Ids also if loaded from snapshot, restarted from log, etc.

raft.c
```diff
@@ -934,7 +943,7 @@ static raft_node_id_t makeRandomNodeId(RedisRaftCtx *rr)
     do {
         RedisModule_GetRandomBytes((unsigned char *) &id, sizeof(id));
         id &= ~(1<<31);
-    } while (!id || (rr->raft && raft_get_node(rr->raft, id) != NULL));
+    } while (!id || (rr->raft && raft_get_node(rr->raft, id) != NULL) || hasNodeIdBeenUsed(rr, id));
```
yossigo, line 946: Am I missing something here? `makeRandomNodeId()` would always get called by a node that is not yet part of the cluster, so `hasNodeIdBeenUsed()` should always be `false`.

node.c
```diff
@@ -329,14 +329,20 @@ void handleNodeAddResponse(redisAsyncContext *c, void *r, void *privdata)
...
+
+        LOG_INFO("Joined Raft cluster, node id: %u, dbid: %.*s\n",
+                node_id,
                 reply->element[1]->len, reply->element[1]->str);
 
         strncpy(rr->snapshot_info.dbid, reply->element[1]->str, reply->element[1]->len);
         rr->snapshot_info.dbid[RAFT_DBID_LEN] = '\0';
 
-        rr->config->id = reply->element[0]->integer;
+        rr->config->id = node_id;
```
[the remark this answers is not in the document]
gavrie: Sure, I agree -- my changes didn't modify this assumption. I just added a local node_id variable to improve readability.

node.c
```diff
@@ -329,14 +329,20 @@ void handleNodeAddResponse(redisAsyncContext *c, void *r, void *privdata)
...
+
+        /* TODO: Remove log message */
+        LOG_INFO("handleNodeAddResponse: addUsedNodeId %d\n", node_id);
+        addUsedNodeId(rr, node_id);
```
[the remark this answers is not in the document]
gavrie: Wasn't aware of it, sounds good!

raft.c
gavrie: Well, theoretically there could be a clash where the random id is equal to a previously used id, which we check for.
In practice the chance is low, but why not check for it?

tests/integration/test_membership.py
gavrie: 👍

raft.c
gavrie: Per out last discussion, we'll keep this.

yossigo (approved):

yossigo (approved):

Issue #65, opened 2020-09-14
Refresh configuration interface


yossigo:
The current configuration interface requires a refresh, and we need to do that as early as possible before approaching beta or GA releases.

Things to do:

- [x] Move from an `argument=value` to an `ARGUMENT <value>` format which is a bit easier/more intuitive and similar the the approach taken by other modules.
- [ ] Reconsider and improve names where applicable (and don't forget to update documentation).

gavrie:
As mentioned in #73, we'll define constants for the static strings defined in `config.c` as part of this issue.

PR #73, opened 2020-10-27
Add initial clustering and sharding support.

yossigo:
See `docs/Clustering.md` for more information.


cluster.c
```diff
@@ -38,15 +227,31 @@ static int addClusterSlotNodeReply(RedisRaftCtx *rr, RedisModuleCtx *ctx, raft_n
...
+    RedisModule_ReplyWithLongLong(ctx, sgn->addr.port);
+    RedisModule_ReplyWithCString(ctx, sgn->node_id);
+
+    return 1;
+}
+
 static void addClusterSlotsReply(RedisRaftCtx *rr, RaftReq *req)
 {
-    int i;
+    int i, j;
```
gavrie, line 608: It would be cleaner to declare these vars when they are first used, inside the `for`.

redisraft.h
```diff
@@ -215,6 +215,18 @@ extern raft_log_impl_t RaftLogImpl;
...
+#define REDIS_RAFT_HASH_MIN_SLOT                    0
+#define REDIS_RAFT_HASH_MAX_SLOT                    16383
+
+#define REDIS_RAFT_VALID_HASH_SLOT(h)   \
+    ((h) >= REDIS_RAFT_HASH_MIN_SLOT && (h) <= REDIS_RAFT_HASH_MAX_SLOT)
+
+#define REDIS_RAFT_VALID_HASH_SLOT_RANGE(a, b) \
+    (REDIS_RAFT_VALID_HASH_SLOT(a) && \
+     REDIS_RAFT_VALID_HASH_SLOT(b) && \
+     (a) <= (b))
```
gavrie, line 230: Do we really need to do this with macros in 2020? Why not use functions instead? The compiler should inline them anyway.

cluster.c
```diff
@@ -124,7 +124,7 @@ void addShardGroupFromArgs(RedisRaftCtx *rr, RedisModuleCtx *ctx, RedisModuleStr
         memcpy(sg_nodes[i].node_id, str, len);
         sg_nodes[i].node_id[len] = '\0';
 
-        str = RedisModule_StringPtrLen(argv[3 + i], &len);
+        str = RedisModule_StringPtrLen(argv[3+(i*2)], &len);
```
gavrie: Maybe we should clean up this number fiddling code a bit and avoid the magic numbers? Since it already introduced a bug once, it may do so again.

cluster.c
```diff
@@ -87,56 +87,61 @@ static RRStatus addShardGroup(RedisRaftCtx *rr, int start_slot, int end_slot, in
  *
  *  [start slot] [end slot] [node-uid node-addr:node-port] [node-uid node-addr:node-port...]
  *
- * !!! This is a temporary hack implemented just for demo/testing purposes. We should
- * !!! replace this by a mechanism similar to RAFT.CLUSTER JOIN, where the user provides
- * !!! host address and port to connect to and we periodically poll remote shardgroups
- * !!! for their configuration.
+ * If parsing errors are encountered, an error rpely is generated on the supplied RedisModuleCtx,
```
gavrie: Typo: `s/rpely/reply/`

cluster.c
```diff
@@ -47,6 +47,91 @@ unsigned int keyHashSlot(const char *key, int keylen) {
...
+ *      <node-uid>:<node host>:<node port>\n
+ *      ...
+ */
+
+/* Serialize a ShardGroup. Returns a newly allocated null terminated buffer
+ * that contains the serialized form.
+ */
+char *ShardGroupSerialize(ShardGroup *sg)
+{
+    size_t buf_size = 80 + (80 + sizeof(ShardGroupNode)) * sg->nodes_num; /* Over estimated */
```
gavrie, line 64: What's `80`? Defining some constants might help.

raft.c
```diff
@@ -112,6 +112,33 @@ static void populateReadonlyCommandDict(RedisModuleCtx *ctx)
     }
 }
 
+/* ------------------------------------ Common helpers ------------------------------------ */
+
+/* Set up a Raft log entry with an attached RaftReq. We use this when a user command provided
+ * in a RaftReq should keep the client blockde until the log entry is committed and applied.
```
gavrie, line 121: `s/blockde/blocked/`

docs/Development.md
```diff
@@ -29,9 +29,11 @@ to configure tests.
...
 For example, running a single test with live logging of all output:
 
-    PYTEST_OPTS="-v --log-cli-level=debug \
-        tests/integration/test_snapshots.py::test_snapshot_delivery" \
-    make integration-tests
+    pytest tests/integration/test_snapshots.py::test_snapshot_delivery -v --log-cli-level=debug
+
+To see a full list of custom test configuration options, use:
+
+    pytest tests/integration --help
```
gavrie, line 36: This is great!

common.c
```diff
@@ -78,9 +78,13 @@ void replyRaftError(RedisModuleCtx *ctx, int error)
  */
 RRStatus checkLeader(RedisRaftCtx *rr, RaftReq *req, Node **ret_leader)
 {
+    const char err_noleader[] = "NOLEADER No raft leader";
+    const char err_clusterdown[] = "CLUSTERDOWN No raft leader";
```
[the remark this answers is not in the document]
yossigo: No particular reason and you're right of course about the `static`, it was left out.

cluster.c
```diff
@@ -0,0 +1,102 @@
...
+
+    if (cmd->argc < 2) {
+        RedisModule_WrongArity(req->ctx);
+        goto exit;
+    }
+
+    size_t cmd_len;
+    const char *cmd_str = RedisModule_StringPtrLen(cmd->argv[1], &cmd_len);
+
+    if (cmd_len == 5 && !strncasecmp(cmd_str, "SLOTS", 5) && cmd->argc == 2) {
```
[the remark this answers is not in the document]
yossigo: `RedisModuleString` (which is really an `robj` with an `sds` value) doesn't guarantee it's going to be null terminated, hence the length variable.

I am OK with the `== 0` convention, but we'll need to be consistent at least in RedisRaft.

config.c
```diff
@@ -325,6 +332,10 @@ void handleConfigGet(RedisModuleCtx *ctx, RedisRaftConfig *config, RedisModuleSt
         len++;
         replyConfigStr(ctx, "loglevel", getLoglevelName(redis_raft_loglevel));
     }
+    if (stringmatch(pattern, "cluster-mode", 1)) {
+        len++;
+        replyConfigBool(ctx, "cluster-mode", config->cluster_mode);
+    }
```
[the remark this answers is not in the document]
yossigo: Sure, maybe do that as part of #65 ?

node.c
```diff
@@ -303,6 +303,22 @@ void NodeAddrListFree(NodeAddrListElement *head)
     }
 }
 
+static bool parseMovedReply(const char *str, NodeAddr *addr)
+{
+    if (strlen(str) < 15 || strncmp(str, "MOVED ", 6))
```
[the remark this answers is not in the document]
yossigo: Shortest form of `-MOVED` supported, adding a comment.

cluster.c
yossigo: Good idea, adding that

cluster.c
yossigo: Fixed already

redisraft.h
yossigo: All the reasons against using macros, except for habit. Will do.

TODO
```diff
@@ -0,0 +1,48 @@
...
+- It sends a `RAFT.SHARDGROUP GET` request and updates the local configuration
+  if anything changed (i.e. new / removed nodes).
+
+## Protocol compatibility
+
+Need to decide if we adopt Redis Cluster as the standard protocol for RedisRaft
+and always stick to it.
+
+Pros:
+- It's standard
```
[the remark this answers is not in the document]
yossigo: I agree, but we need to adjust the Jepsen tests for that. I will open an issue.

cluster.c
```diff
@@ -0,0 +1,1222 @@
...
+
+    /* Serialize */
+    char *payload = ShardGroupSerialize(sg);
+    if (!payload) {
+        return RR_ERROR;
+    }
+
+    raft_entry_t *entry = raft_entry_new(strlen(payload));
+    entry->type = type;
+    entry->id = rand();
```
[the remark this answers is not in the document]
yossigo: Yes, it's just for debugging purposes to make it easy to distinguish entries even if logs become garbled/offseted. We inherit this from the Raft library and it proved quite useful in debugging so far.

common.c
```diff
@@ -73,14 +73,45 @@ void replyRaftError(RedisModuleCtx *ctx, int error)
...
+{
+    size_t reply_maxlen = strlen(addr->host) + 40;
+    char *reply = RedisModule_Alloc(reply_maxlen);
+    int slot = 0;
+    if (req->type == RR_REDISCOMMAND && req->r.redis.hash_slot != -1)
+        slot = req->r.redis.hash_slot;
+    if (rr->config->cluster_mode) {
+        snprintf(reply, reply_maxlen, "MOVED %d %s:%u", slot, addr->host, addr->port);
+    } else {
+        snprintf(reply, reply_maxlen, "MOVED %s:%u", addr->host, addr->port);
```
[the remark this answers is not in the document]
yossigo: Yes, as noted above I'll open an issue.

gavrie (approved):

PR #87, opened 2021-04-11
Fix API compatibility with Redis unstable.

yossigo:
The fix introduced by redis/redis#8670 makes it mandatory for modules to
specify persistence callbacks for registered data types.


sjpotter (approved):
it looks fine to me as far as I can make it out, but I don't see where the struct its part of is used anywhere.  Is it not used directly by our code (i.e. itself or registering functions with the module interface), but used by the redis module interface directly (i.e. finds the symbol and uses it)

PR #103, opened 2021-05-11
add timeout to cluster joining

sjpotter:
add 2 config flags "connection-timeout" and "join-timeout" which control timeout on individual calls to hiredis as it tries to connect to individual nodes as well as how long the entire join operation can take in aggregate, before we error it out

### DIFF
--- a/connection.c
+++ b/connection.c
@@ -62,6 +62,10 @@ Connection *ConnCreate(RedisRaftCtx *rr, void *privdata, ConnectionCallbackFunc 
     conn->idle_callback = idle_cb;
     conn->free_callback = free_cb;
     conn->id = ++id;
+    if (rr->config->connection_timeout > 0) {
+        conn->timeout = RedisModule_Calloc(1, sizeof(struct timeval));
+        conn->timeout->tv_usec = rr->config->connection_timeout;
+    }
 
     CONN_TRACE(conn, "Connection created.");
 
@@ -83,6 +87,11 @@ static void ConnFree(Connection *conn)
     CONN_TRACE(conn, "Connection freed.");
 
     LIST_REMOVE(conn, entries);
+
+    if (conn->timeout != NULL) {
+        RedisModule_Free(conn->timeout);
+    }
+
     RedisModule_Free(conn);
 }
 
@@ -213,7 +222,15 @@ static void handleResolved(uv_getaddrinfo_t *resolver, int status, struct addrin
     if (conn->rc != NULL) {
         redisAsyncFree(conn->rc);
     }
-    conn->rc = redisAsyncConnect(conn->ipaddr, conn->addr.port);
+
+    /* copy default options setup from redisAsyncConnect with support for timeout */
+    redisOptions options = {0};
+    if (conn->timeout != NULL) {
+        options.timeout = conn->timeout;
+    }
+    REDIS_OPTIONS_SET_TCP(&options, conn->ipaddr, conn->addr.port);
+
+    conn->rc = redisAsyncConnectWithOptions(&options);
     if (conn->rc->err) {
         conn->state = CONN_CONNECT_ERROR;
         conn->connect_errors++;
@@ -232,6 +249,8 @@ static void handleResolved(uv_getaddrinfo_t *resolver, int status, struct addrin
     redisAsyncSetConnectCallback(conn->rc, handleConnected);
     redisAsyncSetDisconnectCallback(conn->rc, handleDisconnected);
 }
+
+
 
 RRStatus ConnConnect(Connection *conn, const NodeAddr *addr, ConnectionCallbackFunc connect_callback)
 {
--- a/connection.c
+++ b/connection.c
@@ -63,8 +63,7 @@ Connection *ConnCreate(RedisRaftCtx *rr, void *privdata, ConnectionCallbackFunc 
     conn->free_callback = free_cb;
     conn->id = ++id;
 
-    conn->timeout = RedisModule_Calloc(1, sizeof(struct timeval));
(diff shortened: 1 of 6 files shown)


sjpotter:
@yossigo redid this, we should decide on what the appropropriate timeout value is, I think some multiple of election timeout (2x?)

sjpotter:
I'd note that shard group connect doesn't retry and probably should in a similiar way to the join (as to an extent its the same and they could suffer from same issue of needing to wait for a new leader)

join.c
```diff
@@ -117,20 +118,27 @@ void joinFreeCallback(void *privdata)
  */
 void joinIdleCallback(Connection *conn)
 {
+    RedisRaftCtx *rr = ConnGetRedisRaftCtx(conn);
     JoinState *state = ConnGetPrivateData(conn);
 
+    time_t now;
+    time(&now);
+
+    // This value should probably be dynamic and be considered a multiple of election time period? 2x?
```
yossigo, line 127: @sjpotter Current convention is c-style `/**/` comments, let's stick with it.
I think x2 or x3 the election period seems right - and we should keep this constant defined in `redisraft.h` and not hidden here.

join.c
```diff
@@ -117,20 +118,27 @@ void joinFreeCallback(void *privdata)
...
+    // need to figure that out (i.e. perhaps     rr->config->election_timeout*2 ?
+    if (difftime(now, state->start) > 10) {
+        LOG_ERROR("timed out trying to connect");
```
yossigo, line 130: Consider something more detailed like "Cluster join: timed out trying to connect to <host>:<port>"

sjpotter: this isn't an individual connection timeout, so the only option would be to try to stringify the whole node addr list, I'm not sure that's right  put a message that it time out as too longer than the "join_timeout" # of seconds

connection.c
```diff
@@ -62,6 +62,10 @@ Connection *ConnCreate(RedisRaftCtx *rr, void *privdata, ConnectionCallbackFunc
     conn->idle_callback = idle_cb;
     conn->free_callback = free_cb;
     conn->id = ++id;
+    if (rr->config->connection_timeout > 0) {
+        conn->timeout = RedisModule_Calloc(1, sizeof(struct timeval));
+        conn->timeout->tv_usec = rr->config->connection_timeout;
+    }
```
yossigo, line 68: @sjpotter I think we can always assume there's a non-zero connection timeout (btw the configuration code already does that), keep the `struct timeval` part of the connection and always enforce it.

connection.c
```diff
@@ -233,6 +250,8 @@ static void handleResolved(uv_getaddrinfo_t *resolver, int status, struct addrin
     redisAsyncSetDisconnectCallback(conn->rc, handleDisconnected);
 }
 
+
+
```
yossigo, line 254: Redundant whitespace

docs/Deployment.md
```diff
@@ -267,6 +267,18 @@ This value should be sufficiently greater than `raft-interval` and
 
 *Default*: 1000
 
+### `connection-timeout`
+
+the number of milliseconds the cluster will wait for connections to other nodes to succeed before timing out.  We reccomend it be set if modifying `election-timeout` to be at least a 3x multiple of it.
```
yossigo, line 272:
```suggestion
The number of milliseconds the cluster will wait for connections to other nodes to succeed before timing out.  We recommend it be set if modifying `election-timeout` to be at least a 3x multiple of it.
```

Typo fix above. What is the rationale behind connection-timeout >= 3 * election-timeout?

docs/Deployment.md
```diff
@@ -267,6 +267,18 @@ This value should be sufficiently greater than `raft-interval` and
...
+
+*Defaults*: 3000
+
+### `join-timeout`
+
+the number of seconds the node will continue to try and connect to the cluster using the provided and discovered nodes, looping through them until a connection is made or the timeout is reached.
```
yossigo, line 278:
```suggestion
The number of seconds the node will continue to try and connect to the cluster using the provided and discovered nodes, looping through them until a connection is made or the timeout is reached.
```

Capitalization fix.

connection.c
sjpotter: ok, removing the if

docs/Deployment.md
sjpotter: fixed

redisraft.h
```diff
@@ -161,6 +161,7 @@ typedef struct Connection {
     long long last_connected_time;      /* Last connection time */
     unsigned long int connect_oks;      /* Successful connects */
     unsigned long int connect_errors;   /* Connection errors since last connection */
+    struct timeval *timeout;      /* Timeout to use if not null */
```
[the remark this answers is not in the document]
sjpotter: fixed

connection.c
```diff
@@ -62,10 +62,9 @@ Connection *ConnCreate(RedisRaftCtx *rr, void *privdata, ConnectionCallbackFunc
     conn->idle_callback = idle_cb;
     conn->free_callback = free_cb;
     conn->id = ++id;
-    if (rr->config->connection_timeout > 0) {
-        conn->timeout = RedisModule_Calloc(1, sizeof(struct timeval));
-        conn->timeout->tv_usec = rr->config->connection_timeout;
-    }
+
+    conn->timeout = RedisModule_Calloc(1, sizeof(struct timeval));
```
yossigo, line 66: @sjpotter Why not keep the `struct timeval` part of the connection structure?

sjpotter: oh, as in not allocated?

yossigo (approved):

PR #105, opened 2021-05-13
Add and report version info/git sha.

yossigo:


version.h
```diff
@@ -0,0 +1,2 @@
+#include "buildinfo.h"
+#define REDISRAFT_VERSION   "255.255.255"
```
sjpotter, line 2: is this what the default version is supposed to be?  I assume it be updated at release time, but still seems like a weird version

yossigo: Yes, it's an easy way to indicate a non-official build.

sjpotter (approved):

PR #106, opened 2021-05-16
Shardgroup fixes.

yossigo:
* Reduce logging verbosity of updates.
* Fix an issue resulting with `<redisraft> {conn:6} Failed to resolve '': unknown node or service` errors:

This PR is about storing the last attempted address in `conn_addr`. If the connection is successful, `use_conn_addr` is set to `true` so a subsequent reconnect will first attempt to reconnect to the same address before iterating nodes.

Before this fix `conn_addr` would not be initialized. Upon successful connection, `use_conn_addr` was set to true and if a reconnection was required it would use the uninitialized `conn_addr`.

### DIFF
--- a/cluster.c
+++ b/cluster.c
@@ -398,6 +398,7 @@ static void establishShardGroupConn(Connection *conn)
             sg->node_conn_idx = 0;
         }
         addr = &sg->nodes[sg->node_conn_idx++].addr;
+        sg->conn_addr = *addr;
     }
 
     LOG_DEBUG("Initiating shardgroup(%u) connection to %s:%u", sg->id, addr->host, addr->port);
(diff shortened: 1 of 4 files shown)


cluster.c
```diff
@@ -398,6 +398,7 @@ static void establishShardGroupConn(Connection *conn)
             sg->node_conn_idx = 0;
         }
         addr = &sg->nodes[sg->node_conn_idx++].addr;
+        sg->conn_addr = *addr;
```
sjpotter, line 401: based on the code / comment below this (where use_conn_addr is set to false) I'm wondering if this is correct?

yossigo: I will clarify (and will update the top comment if that makes sense).

This PR is about storing the last attempted address in `conn_addr`. If the connection is successful, `use_conn_addr` is set to `true` so a subsequent reconnect will first attempt to reconnect to the same address before iterating nodes.

Before this fix `conn_addr` would not be initialized. Upon successful connection, `use_conn_addr` was set to true and if a reconnection was required it would use the uninitialized `conn_addr`.

sjpotter (approved):

PR #108, opened 2021-05-20
Tests: make fsync optional.

yossigo:


sjpotter:
so I'm not 100% sure I understand this.  From my reading it looks like we now have control over passing the fsync flag to pytest which will be passed onto the redis instances that it brings up.  and we have an additional test within a test that tests if the commit id is up to date with whats expected.  seems simple enough, but want to make sure before I approve that I understand it.

sjpotter (approved):

PR #109, opened 2021-05-20
Fix crashes with large AE messages.

yossigo:
Before this commit, the Redis command was constructed on the stack which
could lead to overflow and crash. Now everything is heap based.

We should also address and optimize the root cause:
* Apply some limit on the number of entries per message
* Create a mechanism that would trigger a snapshot and deliver that
instead of a huge number of entries.

### DIFF
--- a/raft.c
+++ b/raft.c
@@ -399,13 +399,16 @@ static int raftSendAppendEntries(raft_server_t *raft, void *user_data,
     Node *node = (Node *) raft_node_get_udata(raft_node);
 
     int argc = 5 + msg->n_entries * 2;
-    char *argv[argc];
-    size_t argvlen[argc];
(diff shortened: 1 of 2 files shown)


raft.c
```diff
@@ -399,14 +399,17 @@ static int raftSendAppendEntries(raft_server_t *raft, void *user_data,
     Node *node = (Node *) raft_node_get_udata(raft_node);
 
     int argc = 5 + msg->n_entries * 2;
-    char *argv[argc];
-    size_t argvlen[argc];
+    char **argv = NULL;
+    size_t *argvlen = NULL;
```
sjpotter, line 403: why does argv need to be on the heap?

sjpotter (approved):

PR #110, opened 2021-05-20
convert cluster join to be synchronous

sjpotter:
instead of returning an immediate OK, will only return an OK (or ERR) when the join completes or times out.

as join can now fail, modify tests to handle that situation

### DIFF
--- a/join.c
+++ b/join.c
@@ -63,7 +63,7 @@ static void handleNodeAddResponse(redisAsyncContext *c, void *r, void *privdata)
                 LOG_VERBOSE("Join redirected to leader: %s:%d", addr.host, addr.port);
                 NodeAddrListAddElement(&state->addr, &addr);
             }
-        } else if (strlen(reply->str) > 12 && !strcmp(reply->str, "CLUSTERDOWN ")) {
+        } else if (strlen(reply->str) > 12 && !strncmp(reply->str, "CLUSTERDOWN ", 12)) {
             LOG_ERROR("RAFT.NODE ADD error: %s, retrying.", reply->str);
         } else {
             LOG_ERROR("RAFT.NODE ADD failed: %s", reply->str);
--- a/join.c
+++ b/join.c
@@ -30,8 +30,14 @@ typedef struct JoinState {
     NodeAddrListElement *addr;
     NodeAddrListElement *addr_iter;
     Connection *conn;
-    time_t start;
+    time_t start;                       /* Time we initiated the join, to enable it to fail if it takes too long */
+    RaftReq *req;                       /* Original RaftReq, so we can return a reply */
 } JoinState;
+
+void HandleClusterJoinFailed(RedisRaftCtx *rr, RaftReq *req) {
+    RedisModule_ReplyWithError(req->ctx, "ERR Failed to join cluster, check logs");
+    RaftReqFree(req);
+}
 
 /* Callback for the RAFT.NODE ADD command.
  */
@@ -71,7 +77,7 @@ static void handleNodeAddResponse(redisAsyncContext *c, void *r, void *privdata)
 
         rr->config->id = reply->element[0]->integer;
 
-        HandleClusterJoinCompleted(rr);
+        HandleClusterJoinCompleted(rr, state->req);
         assert(rr->state == REDIS_RAFT_UP);
 
         ConnAsyncTerminate(conn);
@@ -125,9 +131,9 @@ void joinIdleCallback(Connection *conn)
     time(&now);
 
     if (difftime(now, state->start) > rr->config->join_timeout) {
-        LOG_ERROR("timed out trying to join cluster");
+        LOG_ERROR("Cluster join: timed out, took longer than %d seconds", rr->config->join_timeout);
         ConnAsyncTerminate(conn);
-        HandleClusterJoinFailed(rr);
+        HandleClusterJoinFailed(rr, state->req);
         return;
     }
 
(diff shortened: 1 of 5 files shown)


sjpotter:
I saw a flakiness in test_membership.py::test_full_cluster_remove 

it does change when it gets an error

E           redis.exceptions.ResponseError: -ERR a voting change is already in progress

I am pretty sure this isn't due to this change and isn't a bug in the code, but a bug in the test being unreliable (i.e. a) expects a certain node to be leader and b) doesn't expect leader to change in course of test)

redisraft.h
```diff
@@ -619,8 +616,7 @@ void RaftRedisCommandArrayMove(RaftRedisCommandArray *target, RaftRedisCommandAr
 /* raft.c */
 RRStatus RedisRaftInit(RedisModuleCtx *ctx, RedisRaftCtx *rr, RedisRaftConfig *config);
 RRStatus RedisRaftStart(RedisModuleCtx *ctx, RedisRaftCtx *rr);
-void HandleClusterJoinCompleted(RedisRaftCtx *rr);
-void HandleClusterJoinFailed(RedisRaftCtx *rr);
+
```
yossigo, line 619: Whitespace

tests/integration/sandbox.py
```diff
@@ -130,9 +133,10 @@ def init(self):
         LOG.info('Cluster created: %s', dbid)
         return self
 
-    def join(self, addresses):
+    def join(self, addresses, expect_error=None):
```
yossigo, line 136: I think there is not enough separation of concerns here between utility functions and test cases. An alternative can be to simply expect the exception when calling `join`, and maybe pass an explicit `retries=1` to save time.

tests/integration/sandbox.py
```diff
@@ -107,20 +107,23 @@ def raftlogidx(self):
...
         retries = self.up_timeout
         if retries is not None:
             retries *= 10
         while True:
             try:
                 return self.client.execute_command('RAFT.CLUSTER', *args)
             except redis.exceptions.RedisError as err:
-                LOG.info(err)
+                if expect_error is not None and expect_error in str(err):
+                    LOG.info(f"found expected error {expect_error}")
```
yossigo, line 119: The test code style uses `"string".format()`.

tests/integration/sandbox.py
```diff
@@ -130,9 +133,10 @@ def init(self):
         LOG.info('Cluster created: %s', dbid)
         return self
 
-    def join(self, addresses):
+    def join(self, addresses, expect_error=None):
         self.start()
-        self.cluster('join', *addresses)
+        logging.info("In Join")
```
yossigo, line 138: Is this really needed or a leftover from earlier tests?

join.c
```diff
@@ -125,9 +131,9 @@ void joinIdleCallback(Connection *conn)
     time(&now);
 
     if (difftime(now, state->start) > rr->config->join_timeout) {
-        LOG_ERROR("timed out trying to join cluster");
+        LOG_ERROR("Cluster join: timed out, took longer than %d seconds", rr->config->join_timeout);
```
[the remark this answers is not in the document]
sjpotter: what's the practical difference?  it hit the limit, so if the limit is 60s and we only see it after 61s because of some hang, what difference does it make?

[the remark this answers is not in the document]
sjpotter: I'd figure that the limit would be even better as it would give a user something to figure out if they don't know where that limit is coming from, vs a random time doesn't give them insight into whats going on

tests/integration/sandbox.py
sjpotter: it's an artifact of me putting in logging to understand what was going on, need to remove.

tests/integration/sandbox.py
yossigo: "Standard is better than better". I don't mind this or the other but keep things consistent where possible.

join.c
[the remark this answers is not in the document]
yossigo: You're right, unless something unexpected happens and it blocks for a very long time... although the log timestamp would also tell the story.

tests/integration/sandbox.py
sjpotter: hmm, ok, that could make sense.  retries=1 and try/except the error in the caller.  I can do that.

tests/integration/test_membership.py
```diff
@@ -7,6 +7,7 @@
 (AGPLv3) or the Redis Source Available License (RSAL).
 """
 import logging
+import redis as redis
```
yossigo, line 10:
```suggestion
import redis
```

tests/integration/test_membership.py
```diff
@@ -167,13 +168,15 @@ def test_remove_and_rejoin_node_with_same_id_fails(cluster, use_snapshot):
...
-    # TODO: Detect the error immediately (by inspecting the logs?) instead of retrying until timeout
-    with raises(RedisRaftTimeout):
-        new_node.wait_for_node_voting()
+    try:
+        new_node = cluster.add_node(port=port, node_id=node_id, single_run=True)
+        logger.info("Waiting for timeout")
+
+        # TODO: Detect the error immediately (by inspecting the logs?) instead of retrying until timeout
+        with raises(RedisRaftTimeout):
+            new_node.wait_for_node_voting()
```
yossigo, line 177: This should be enough -- it will fail
* If `RedisRaftTimeout` is not raised
* If any other exception is raised

join.c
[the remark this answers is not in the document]
sjpotter: unsure if you are happy with this.  note, I added another failure case now.  we can short circuit retries if we get an unrecoverable error.  (speeds up testing a lot, no more retry on the try to readd node with old node id test)

tests/integration/sandbox.py
sjpotter: think I've done this now

tests/integration/sandbox.py
sjpotter: fixed all f"" strings

tests/integration/test_membership.py
sjpotter: this is now very changed

yossigo (approved):

PR #112, opened 2021-05-26
unify joining and shard group link code

sjpotter:
we know have a common set of connection handling code

+ python test string formating fix

### DIFF
--- a/common.c
+++ b/common.c
@@ -179,16 +179,12 @@ bool parseMovedReply(const char *str, NodeAddr *addr)
     return NodeAddrParse(p + 1, strlen(p), addr);
 }
 
-/* future types should remain 4 characters or stack array in common.c/joinLinkIdleCallback() has to expand */
-char * join_type = "join";
-char * link_type = "link";
-
 /* Invoked when the connection is not connected or actively attempting
  * a connection.
  */
 void joinLinkIdleCallback(Connection *conn)
 {
-    char err_msg[42];
+    char err_msg[50];
 
     RedisRaftCtx *rr = ConnGetRedisRaftCtx(conn);
     JoinLinkState *state = ConnGetPrivateData(conn);
@@ -225,7 +221,7 @@ void joinLinkIdleCallback(Connection *conn)
 exit_fail:
     ConnAsyncTerminate(conn);
 
-    snprintf(err_msg, 42, "ERR failed to %s cluster, please check logs", state->type);
+    snprintf(err_msg, 50, "ERR failed to %s cluster, please check logs", state->type);
     RedisModule_ReplyWithError(state->req->ctx, err_msg);
     RaftReqFree(state->req);
 }
@@ -235,6 +231,10 @@ exit_fail:
 void joinLinkFreeCallback(void *privdata)
 {
     JoinLinkState *state = (JoinLinkState *) privdata;
+
+    if (state->type != NULL) {
+        RedisModule_Free(state->type);
+    }
 
     NodeAddrListFree(state->addr);
     RedisModule_Free(state);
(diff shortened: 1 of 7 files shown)


redisraft.h
```diff
@@ -553,6 +553,9 @@ typedef struct SnapshotResult {
     char err[256];
 } SnapshotResult;
 
+extern char * join_type;
+extern char * link_type;
+
```
yossigo, line 558: I think there should be a better way to handle this than introduce these globals. If we're trying to keep this reusable and generic, maybe we should just have a callback function for unrecovarable error handling - which will be responsible for producing the apropriate error reply as well perform any other specific cleanup tasks we may need in the future.

common.c
```diff
@@ -177,4 +177,65 @@ bool parseMovedReply(const char *str, NodeAddr *addr)
...
+/* future types should remain 4 characters or stack array in common.c/joinLinkIdleCallback() has to expand */
+char * join_type = "join";
+char * link_type = "link";
+
+/* Invoked when the connection is not connected or actively attempting
+ * a connection.
+ */
+void joinLinkIdleCallback(Connection *conn)
+{
+    char err_msg[42];
```
yossigo, line 191: I'd make it just a bit bigger to accomodate for future error message changes, etc.

common.c
```diff
@@ -177,4 +177,65 @@ bool parseMovedReply(const char *str, NodeAddr *addr)
...
+
+    time_t now;
+    time(&now);
+
+    if (state->failed) {
+        LOG_ERROR("Cluster %s: unrecoverable error, check logs", state->type);
+        goto exit_fail;
+    }
+
+    if (difftime(now, state->start) > rr->config->join_timeout) {
```
yossigo, line 204: We need to either have a different timeout setting or document that the join applies to shardgroup linking as well.

common.c
```diff
@@ -177,4 +177,65 @@ bool parseMovedReply(const char *str, NodeAddr *addr)
...
+    /* Establish connection. We silently ignore errors here as we'll
+     * just get iterated again in the future.
+     */
+    ConnConnect(state->conn, &state->addr_iter->addr, state->connect_callback);
+    return;
+
+exit_fail:
+    ConnAsyncTerminate(conn);
+
+    snprintf(err_msg, 42, "ERR failed to %s cluster, please check logs", state->type);
```
yossigo, line 228:
```suggestion
    snprintf(err_msg, sizeof(err_msg), "ERR failed to %s cluster, please check logs", state->type);
```

Safer this way.

redisraft.h
```diff
@@ -580,7 +583,20 @@ static int inline checkInRedisModuleCall(void) {
     return redis_raft_in_rm_call;
 }
 
+typedef struct JoinLinkState {
```
yossigo, line 586: Now that's it's totally generic, maybe this should be named something like `NodeExecuteCommandState` or something similar?

common.c
sjpotter: I think documenting is fine, they are both "join" type operations

common.c
sjpotter: I guess its not the answer as is :)

redisraft.h
sjpotter: that's less descriptive to me.  i.e. this is state that relates to joining/linking redis raft nodes/clusters to each other.    I'm unsure what nodeExecuteCommand actually means

redisraft.h
sjpotter: ok, can get rid of the globals, I'd rather not add another callback yet if not understood what it really adds over the existign code.  I'll just add allocated data to state object and free in its normal callback

yossigo (approved):

PR #114, opened 2021-05-31
Only return to user after log commit

sjpotter:
we attach raft requests to a log entry to enable us to delay unblocking and returning to the caller once the log entry is actually committed.  

in addition, have a fix (perhaps a hack and needs to be done better) to allow leader to exit on its own removal.

### DIFF
--- a/raft.c
+++ b/raft.c
@@ -117,8 +117,8 @@ static void populateReadonlyCommandDict(RedisModuleCtx *ctx)
 
 /* ------------------------------------ Common helpers ------------------------------------ */
 
-static void handleApplyAllRet(RedisRaftCtx *rr, int i) {
-    if (i == RAFT_ERR_SHUTDOWN)  {
+static void handleApplyAllRet(RedisRaftCtx *rr, int ret) {
+    if (ret == RAFT_ERR_SHUTDOWN)  {
         LOG_INFO("*** NODE REMOVED, SHUTTING DOWN.");
 
         if (rr->config->raft_log_filename)
@@ -521,7 +521,8 @@ static int raftApplyLog(raft_server_t *raft, void *user_data, raft_entry_t *entr
 
     switch (entry->type) {
         case RAFT_LOGTYPE_DEMOTE_NODE:
-            if (raft_req != NULL && rr->state == REDIS_RAFT_UP && raft_is_leader(rr->raft)) {
+            LOG_INFO("DEMOTE");
+            if (rr->state == REDIS_RAFT_UP && raft_is_leader(rr->raft)) {
                 raft_entry_t *rem_entry = raft_entry_new(sizeof(RaftCfgChange));
 
                 entryAttachRaftReq(rr, rem_entry, raft_req);
@@ -539,6 +540,7 @@ static int raftApplyLog(raft_server_t *raft, void *user_data, raft_entry_t *entr
             }
             break;
         case RAFT_LOGTYPE_REMOVE_NODE:
+            LOG_INFO("REMOVE");
             req = (RaftCfgChange *) entry->data;
 
             entryDetachRaftReq(rr, entry);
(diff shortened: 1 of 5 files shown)


raft.c
```diff
@@ -496,12 +496,19 @@ static int raftPersistTerm(raft_server_t *raft, void *user_data, raft_term_t ter
...
 {
     RedisRaftCtx *rr = user_data;
+    RaftReq *raft_req = entry->user_data;
     RaftCfgChange *req;
 
     switch (entry->type) {
         case RAFT_LOGTYPE_DEMOTE_NODE:
             if (rr->state == REDIS_RAFT_UP && raft_is_leader(rr->raft)) {
                 raft_entry_t *rem_entry = raft_entry_new(sizeof(RaftCfgChange));
+
```
yossigo, line 506: Maybe it would be a good idea to skip this if `raft_req` is already NULL, there are different scenarios that could lead to that.

raft.c
```diff
@@ -496,12 +496,19 @@ static int raftPersistTerm(raft_server_t *raft, void *user_data, raft_term_t ter
...
+                entryAttachRaftReq(rr, rem_entry, raft_req);
+                // remove and detach from previous entry
```
yossigo, line 508:
```suggestion
                /* Remove and detach from previous entry */
```

raft.c
```diff
@@ -510,17 +517,42 @@ static int raftApplyLog(raft_server_t *raft, void *user_data, raft_entry_t *entr
...
+            rr->client_attached_entries--;
+
+            if (raft_req) {
+                RedisModule_ReplyWithSimpleString(raft_req->ctx, "OK");
+                RaftReqFree(raft_req);
+            }
+
             if (req->id == raft_get_nodeid(raft)) {
                 LOG_DEBUG("Removing this node from the cluster");
+                rr->raft_exit_node = true;
```
yossigo, line 537: Is this really needed?

Seems like `raft_apply_all()` would propagate the `RAFT_ERR_SHUTDOWN` all the way, but the problem is `raft.c` calls it in several cases without checking and acting on errors.

raft.c
```diff
@@ -510,17 +517,42 @@ static int raftApplyLog(raft_server_t *raft, void *user_data, raft_entry_t *entr
...
                 return RAFT_ERR_SHUTDOWN;
             }
             break;
+        case RAFT_LOGTYPE_ADD_NONVOTING_NODE:
+            entry->user_data = NULL;
+            rr->client_attached_entries--;
```
yossigo, line 543: Looks like this is repeating, maybe an `entryDetachRaftReq()` would be useful?

raft.c
sjpotter: as I understand this comment, what you are arguing is that it makes no sense to add a "REMOVE" entry (after demote) on any node besides the leader (where raft_req would exist)?

raft.c
sjpotter: ok, added code for each of them, that seemed to handle it.

raft.c
sjpotter: done

raft.c
sjpotter: though, my understanding of the comment can't be right, as we already have a check for "raft_is_leader(rr->raft)

sjpotter:
I've cleaned it up per your suggestion, but I'm seeing a problem.  I create N nodes (say 1 to N).  I go step by step removing leader.  They all exit, until I get to the last 2 nodes.  the N-1 node exits, but the Nth node ends up in a stuck voting state where it still thinks it should be connected to the N-1th node that was removed.

raft.c
yossigo: The leader could have been re-elected in between.

sjpotter: ok, now I understand.  can't handle raft_req issues in the if leader because the node with the req might not be the leader anymore.  which also explains why the break was within the if leader as the fall through was by design.

yossigo:
Replaced by #170 for node removal.

PR #115, opened 2021-06-15
Fix build issues with TRACE=1

yossigo:
This seems to have been broken with the connection/node refactoring.


sjpotter:
looks fine, but perhaps we should also do a github action build with TRACE=1 to make sure it doesn't break building in future?

yossigo:
@sjpotter I think we can take the risk, it's not likely to be broken and has no production value anyway.

sjpotter (approved):

PR #117, opened 2021-07-01
Prepare RedisRaft to run Redis test suite.

yossigo:
This fixes a few relatively minor issues and establishes the script and configuration required to run the Redis test suite on a RedisRaft cluster. At this point not all tests work yet but a `skip.txt` skips those that don't and documents why/what needs to be done - so it's a good scoping excercise.

One notable change here is the introduction of `commands.c` as a single unified to classify Redis commands into different categories:

* Read-only commands that need to go through quorum reads (but skip the log)
* Administrative commands that are not intercepted by RedisRaft
* Unsupported commands which should be blocked.

### DIFF
--- a/commands.c
+++ b/commands.c
@@ -1,0 +1,172 @@
+/*
+ * This file is part of RedisRaft.
+ *
+ * Copyright (c) 2021 Redis Labs
+ *
+ * RedisRaft is dual licensed under the GNU Affero General Public License version 3
+ * (AGPLv3) or the Redis Source Available License (RSAL).
+ */
+/* ------------------------------------ Command Classification ------------------------------------ */
+
+#include <string.h>
+#include <ctype.h>
+#include "redisraft.h"
+
+static RedisModuleDict *commandSpecDict = NULL;
+
+RRStatus CommandSpecInit(RedisModuleCtx *ctx)
+{
+    static CommandSpec commands[] = {
+            /* Core Redis Commands */
+            { "get",                    CMD_SPEC_READONLY },
+            { "strlen",                 CMD_SPEC_READONLY },
+            { "exists",                 CMD_SPEC_READONLY },
+            { "getbit",                 CMD_SPEC_READONLY },
+            { "getrange",               CMD_SPEC_READONLY },
+            { "substr",                 CMD_SPEC_READONLY },
+            { "mget",                   CMD_SPEC_READONLY },
+            { "llen",                   CMD_SPEC_READONLY },
+            { "lindex",                 CMD_SPEC_READONLY },
+            { "lrange",                 CMD_SPEC_READONLY },
+            { "scard",                  CMD_SPEC_READONLY },
+            { "sismember",              CMD_SPEC_READONLY },
+            { "srandmember",            CMD_SPEC_READONLY },
+            { "sinter",                 CMD_SPEC_READONLY },
+            { "sunion",                 CMD_SPEC_READONLY },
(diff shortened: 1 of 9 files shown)


commands.c
```diff
@@ -0,0 +1,167 @@
...
+            { "raft.entry",             CMD_SPEC_DONT_INTERCEPT },
+            { "raft.config",            CMD_SPEC_DONT_INTERCEPT },
+            { "raft.cluster",           CMD_SPEC_DONT_INTERCEPT },
+            { "raft.shardgroup",        CMD_SPEC_DONT_INTERCEPT },
+            { "raft.node",              CMD_SPEC_DONT_INTERCEPT },
+            { "raft.ae",                CMD_SPEC_DONT_INTERCEPT },
+            { "raft.requestvote",       CMD_SPEC_DONT_INTERCEPT },
+            { "raft.loadsnapshot",      CMD_SPEC_DONT_INTERCEPT },
+            { "raft.debug",             CMD_SPEC_DONT_INTERCEPT },
+            { "raft.info",              CMD_SPEC_DONT_INTERCEPT },
```
sjpotter, line 93: my personal asthetic doesn't tink the raft commands should be intermixed with the regular redis commands here.  while they are mostly their own block, it could make it a little hard to make sure all redis commands are covered?

sjpotter: also, should all write command be enumerated here (i.e. essentialy to enumerate all commands not assume something not listed here is not a write command)

yossigo:
Eventually - maybe yes. But for now I think that only handling specific commands and defaulting to pushing everything else through Raft makes sense. Also, this makes it possible to support additional modules on top of RedisRaft.

raft.c
```diff
@@ -1789,7 +1692,11 @@ static void handleRedisCommand(RedisRaftCtx *rr,RaftReq *req)
     /* If command is read only we don't push it to the log, but queue it
      * until we can confirm it's safe to execute (i.e. still a leader).
      */
-    if (checkReadOnlyCommandArray(&req->r.redis.cmds)) {
+    unsigned int cmd_flags = CommandSpecGetAggregateFlags(&req->r.redis.cmds);
+    if (cmd_flags & CMD_SPEC_UNSUPPORTED) {
+        RedisModule_ReplyWithError(req->ctx, "ERR not supported by RedisRaft");
+        goto exit;
+    } else if (cmd_flags & CMD_SPEC_READONLY) {
```
sjpotter, line 1699: I dont understand how this works.  it seems that if in a multi one has multiple commands and one if CMD_SPEC_READONLY, the flag would be set, even if there were other write commands mixed in?

skiptests.txt
```diff
@@ -0,0 +1,137 @@
+## No SYNC support
```
sjpotter, line 1: wondering what this whole file is used for?  i see skip.txt being used, but not this

raft.c
yossigo: Yes, that needs to be fixed. The reason it worked was that when processing commands from a user, there will always be just a single command. The only case where we currently have multiple commands batched in a single entry is when processing a log entry that is a result of `MULTI/EXEC`.

yossigo: @sjpotter The statement above is wrong, this *does* affect the leader as well. Fixing and adding a test.

skiptests.txt
yossigo: Left over, will remove.

PR #128, opened 2021-07-26
Move snapshot out of keyspace.

### LINKED ISSUE #111: Move metadata out of keyspace
Currently information about snapshot and shardgroups is stored as a meta-key:
* A way to store this info as part of a snapshot
* Guarantees about atomicity (in particular, making sure the last applied index is always correct for the specific snapshot saved).

This is not a clean solution and can probably be implemented in a nicer way using hooks, which were not available when it was originally developed.

yossigo:
Before this commit, snapshot information was persisted by creating a
module data type and having a fake key in the keyspace. The RDB save and
load callbacks would then be triggered to store or load the information.

With new versions of the module API that support RDB AUX saving and
loading, it is no longer necessary to have a fake key.

Fixes #111

### DIFF
--- a/join.c
+++ b/join.c
@@ -117,9 +117,6 @@ void handleClusterJoin(RedisRaftCtx *rr, RaftReq *req)
         goto exit_fail;
     }
 
-    /* Create a Snapshot Info meta-key */
-    initializeSnapshotInfo(rr);
-
     JoinLinkState *state = RedisModule_Calloc(1, sizeof(*state));
     state->type = RedisModule_Calloc(1, strlen(type)+1);
     strcpy(state->type, type);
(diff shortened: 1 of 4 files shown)


sjpotter:
i'm trying to understand what changed here.  I see the removal of the meta key initilization (and the function that did it), but otherwise it seems exactly the same except that it uses the aux callbacks and not the rdb callbacks?  nothing else had to change?

yossigo:
@sjpotter Not much needed to be changed because the serialization and deserialization of aux fields and keys uses the same `RedisModuleIO` primitives.

The main benefits here, if it's not clear, are:
* Don't rely on a fake key to trigger serialization of snapshot info (e.g. more resilient, someone could have removed it before this commit)
* Don't clutter the keyspace which is visible to users

tezc (approved):

dipAch:
I think this comment should have been also updated with this merge, just a FYI.

https://github.com/RedisLabs/redisraft/blob/29499d398d746dba2a62c7cc90229f6fabf636fa/redisraft.h#L220

PR #130, opened 2021-08-09
Update raft library

tezc:
- Updated raft library
- Removed `demote` related parts as `demote` is removed from the raft library.
- Made `test_full_cluster_remove` test not to expect a shutdown(process exit). 
- Future TODO : Shutdown feature remains to be fixed as a future improvement.

### DIFF
--- a/tests/integration/test_membership.py
+++ b/tests/integration/test_membership.py
@@ -126,6 +126,20 @@ def test_full_cluster_remove(cluster):
         expected_nodes -= 1
         leader.wait_for_num_nodes(expected_nodes)
 
+    # TODO: Currently, shutdown feature does not work reliably. This part of the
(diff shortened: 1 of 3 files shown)


tests/integration/test_membership.py
```diff
@@ -126,18 +126,6 @@ def test_full_cluster_remove(cluster):
         expected_nodes -= 1
         leader.wait_for_num_nodes(expected_nodes)
 
-    # make sure other nodes are down
```
yossigo, line 129: @tezc As we do plan to restore the shutdown behavior in the future, how about keeping this code disabled and leave a comment that reminds us to reinstate it when shutdown is implemented?

tezc: Done.

yossigo (approved):

PR #132, opened 2021-08-09
Cmake support

tezc:
- Added initial cmake support to make the project compatible with IDEs that expects Cmake primarily.
- All dependencies have cmake support already, code navigation and source code change detection should work on all IDEs.


yossigo:
@tezc I have really very basic experience with CMake, but it looks good to me.

One thing I noticed is `make test` fails, I presume it attempts to run imported tests from sub-directories (hiredis, libuv) which we probably want to skip. On the other hand, I guess we do want to preserve the unit and integration tests behavior under CMake right?

BTW I guess we can also do this in two steps if this becomes blocking due to IDE compatibility.

yossigo (approved):

tezc:
I didn't try to duplicate what we are doing with Makefile. I just added things that make IDE a better development environment.
Let me take another look, we can provide similar behavior.

tezc:
@yossigo added coverage and integration tests. At the top of the file, there is a comment with command examples

yossigo (approved):

PR #133, opened 2021-08-10
POC of integrating transfer leader into redisraft

sjpotter:
add 2 raft. commands to reddis
	raft.transfer_leader <node id>
	raft.timeout_now

user will issue a raft.transfer_leader command to existing leader
libraft itself will issue a callback at the redisraft callback handler to issue the raft.timeout_now at the right point in time

### DIFF
--- a/raft.c
+++ b/raft.c
@@ -199,7 +199,6 @@ static void executeLogEntry(RedisRaftCtx *rr, raft_entry_t *entry, raft_index_t 
     }
 }
 
-
 /* ------------------------------------ RequestVote ------------------------------------ */
 
 static void handleRequestVoteResponse(redisAsyncContext *c, void *r, void *privdata)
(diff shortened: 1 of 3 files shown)


raft.c
```diff
@@ -260,13 +259,15 @@ static int raftSendRequestVote(raft_server_t *raft, void *user_data,
...
-                node, "RAFT.REQUESTVOTE %d %d %d:%d:%d:%d",
+                node, "RAFT.REQUESTVOTE %d %d %d:%d:%d:%d:%d",
                 raft_node_get_id(raft_node),
                 raft_get_nodeid(raft),
                 msg->term,
                 msg->candidate_id,
                 msg->last_log_idx,
-                msg->last_log_term) != REDIS_OK) {
+                msg->last_log_term,
+                raft_is_timeout_now(raft)
```
tezc, line 269: I think you may set msg->transfer_leader in the library so lib user wouldn't need to call this function. It'll just use it as another variable here as msg->transfer_leader.

yossigo:
Closing as stale, implemented in #157

PR #135, opened 2021-08-28
Enable raft library logs only if log level is debug

tezc:
- Enable raft library logs only if log level is debug
- Accept case-insensitive config

### DIFF
--- a/raft.c
+++ b/raft.c
@@ -1029,6 +1029,12 @@ static void initRaftLibrary(RedisRaftCtx *rr)
     }
     raft_set_election_timeout(rr->raft, rr->config->election_timeout);
     raft_set_request_timeout(rr->raft, rr->config->request_timeout);
+
+    // To avoid performance hit, get library logs only if log level is debug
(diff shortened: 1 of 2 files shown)


raft.c
```diff
@@ -1029,6 +1029,12 @@ static void initRaftLibrary(RedisRaftCtx *rr)
     }
     raft_set_election_timeout(rr->raft, rr->config->election_timeout);
     raft_set_request_timeout(rr->raft, rr->config->request_timeout);
+
+    // To avoid performance hit, get library logs only if log level is debug
+    if (redis_raft_loglevel != LOGLEVEL_DEBUG) {
+        redis_raft_callbacks.log = NULL;
```
sjpotter, line 1035: I don't like this, as we have many different levels of logging.  why dont we just have a set_log_level() and have raft_log() check that?

tezc: We don't have any log level in libraft. It's just on/off.

sjpotter (approved):

PR #136, opened 2021-08-30
Leader informs removed node when its removal is committed

tezc:
While removing a node from the cluster, leader node might stop replicating entries to that node as soon as the removal entry is appended to the leader's log file. So, target node may not see its removal entry. To inform the node that it is removed, leader sends an out-of-band, best-effort message to that node.

### DIFF
--- a/raft.c
+++ b/raft.c
@@ -242,26 +242,14 @@ static void raftSendNodeShutdown(raft_node_t *raft_node)
         return;
     }
 
-    char tmp[32];
-    snprintf(tmp, sizeof(tmp), "%d", raft_node_get_id(raft_node));
-
-    char* argv[2] = {
-        "RAFT.NODESHUTDOWN",
-        tmp
-    };
-
-    size_t argvlen[2] = {
-        strlen("RAFT.NODESHUTDOWN"),
-        strlen(tmp)
-    };
-
     if (!ConnIsConnected(node->conn)) {
         NODE_TRACE(node, "not connected, state=%s", ConnGetStateStr(node->conn));
         return;
     }
 
-    if (redisAsyncCommandArgv(ConnGetRedisCtx(node->conn), NULL,
-                        node, 2, (const char **)argv, argvlen) != REDIS_OK) {
+    if (redisAsyncCommand(ConnGetRedisCtx(node->conn), NULL, NULL,
+                "%s %d", "RAFT.NODESHUTDOWN",
+                (int) raft_node_get_id(raft_node)) != REDIS_OK) {
         NODE_TRACE(node, "failed to send raft.nodeshutdown");
     }
 }
--- a/raft.c
+++ b/raft.c
@@ -231,6 +231,40 @@ static void executeLogEntry(RedisRaftCtx *rr, raft_entry_t *entry, raft_index_t 
     }
 }
 
+static void raftSendNodeShutdown(raft_node_t *raft_node)
+{
+    if (!raft_node) {
+        return;
+    }
+
+    Node *node = raft_node_get_udata(raft_node);
+    if (!node) {
+        return;
+    }
+
+    char tmp[32];
+    snprintf(tmp, sizeof(tmp), "%d", raft_node_get_id(raft_node));
+
+    char* argv[2] = {
+        "RAFT.NODESHUTDOWN",
+        tmp
+    };
+
+    size_t argvlen[2] = {
+        strlen("RAFT.NODESHUTDOWN"),
+        strlen(tmp)
+    };
+
+    if (!ConnIsConnected(node->conn)) {
+        NODE_TRACE(node, "not connected, state=%s", ConnGetStateStr(node->conn));
+        return;
+    }
+
+    if (redisAsyncCommandArgv(ConnGetRedisCtx(node->conn), NULL,
+                        node, 2, (const char **)argv, argvlen) != REDIS_OK) {
+        NODE_TRACE(node, "failed to send raft.nodeshutdown");
+    }
+}
 
 /* ------------------------------------ RequestVote ------------------------------------ */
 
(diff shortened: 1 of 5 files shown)


sjpotter:
I'm wondering if there's a better way to do this.

imagine either

a) we never delete a node struct, but when removed, we mark the idx of the entry that removed it.
b) we just keep a map of node -> idx removed (when we create the node, its mapped value is set to MAX_VAL)

then if/when the removed node sends a message to any other node and its remove has been committed (commit_idx >= recorded idx above), we are able to respond "in band" (not out of band or best effort) with a definitive message that "you have been removed"

i.e. I'm unconvinced that sending a message "out of band" will help, as in reality its not out of band, its the same tcp connection.  we could just as easily before removal, after commit_idx update send an appendentry to it, without an additional rpc.

tezc:
IMHO users should be fine even without this feature.  I see this as a "nice to have" one and a fire and forget message works all the time unless the node disconnects right at that moment.

Otherwise, you'll maintain a state and verify each message for that. Personally, I don't think it's worth any of this. Let me know if I'm missing something, if it is important to have.

sjpotter:
I'm unsure recording idx state of when a node was removed is so hard.  Right now we basically treat the nodes as a linked list that has to be scanned through each time (i.e. as its not a map, its an array, that isn't indexed by anything special, so has to be scanned 1 by 1, it probably be easier to treat it as a linked list and not do memmove() on it.  but that's an aside.

my gut says it be nice to do this right and say "your removal has been committed, bye bye".  But I'll leave that up to @yossigo

raft.c
```diff
@@ -437,6 +471,10 @@ static int raftApplyLog(raft_server_t *raft, void *user_data, raft_entry_t *entr
                 LOG_DEBUG("Removing this node from the cluster");
                 return RAFT_ERR_SHUTDOWN;
             }
+
+            if (raft_is_leader(raft)) {
+                raftSendNodeRemoved(raft_get_node(raft, req->id));
+            }
```
sjpotter, line 477: do we want to avoid sending if if we are removing self?

raft.c
```diff
@@ -2108,6 +2146,28 @@ void handleShardGroupGet(RedisRaftCtx *rr, RaftReq *req)
     RaftReqFree(req);
 }
 
+static void handleNodeRemoved(RedisRaftCtx *rr, RaftReq *req)
+{
+    if (req->r.node_removed.id != raft_get_nodeid(rr->raft)) {
+        LOG_ERROR("Received invalid noderemoved message with id : %d.",
+                  req->r.node_removed.id);
+        return;
+    }
```
sjpotter, line 2155: I'm not against this code, but if this happens, we probably have other issues too.

raft.c
```diff
@@ -199,6 +199,40 @@ static void executeLogEntry(RedisRaftCtx *rr, raft_entry_t *entry, raft_index_t
     }
 }
 
+static void raftSendNodeRemoved(raft_node_t *raft_node)
```
sjpotter, line 202: I'd call it raftSendNodeShutdown (and throughout).  Yes, its saying "you've been removed", but its really sending a shutdown command.

raft.c
tezc: has self check just above this line :)

raft.c
tezc: I see. I thought leader sends "node removed" message, what to do is up to the receiver, so we may change the behavior, decide not to shutdown and do something else without changing the protocol (old habit from protocol jobs). But I see no need this here. I'll change it to "node shutdown".  Thanks.

raft.c
sjpotter: ahh, missed it.

raft.c
tezc: Just it is a precaution but I think leader may send this message on replay or if another leader is elected and that leader applies this entry after becoming the leader.  If we remove a node and start another node at the same machine (ip:port), new node would shutdown if I'm not mistaken

tezc:
@yossigo @sjpotter Guys, you didn't forget this PR right? :) It'd great if you can take a look sometime.

raft.c
```diff
@@ -231,6 +231,40 @@ static void executeLogEntry(RedisRaftCtx *rr, raft_entry_t *entry, raft_index_t
...
+        strlen("RAFT.NODESHUTDOWN"),
+        strlen(tmp)
+    };
+
+    if (!ConnIsConnected(node->conn)) {
+        NODE_TRACE(node, "not connected, state=%s", ConnGetStateStr(node->conn));
+        return;
+    }
+
+    if (redisAsyncCommandArgv(ConnGetRedisCtx(node->conn), NULL,
```
yossigo, line 263: Any reason to use `redisAsyncCommandArgv` and not just `redisAsyncCommand` here? Might be shorter and simpler.

tezc: I don't remember, probably unfamiliarity with the hiredis API caused that. Fixed.

raft.c
```diff
@@ -231,6 +231,28 @@ static void executeLogEntry(RedisRaftCtx *rr, raft_entry_t *entry, raft_index_t
...
+        return;
+    }
+
+    if (!ConnIsConnected(node->conn)) {
+        NODE_TRACE(node, "not connected, state=%s", ConnGetStateStr(node->conn));
+        return;
+    }
+
+    if (redisAsyncCommand(ConnGetRedisCtx(node->conn), NULL, NULL,
+                "%s %d", "RAFT.NODESHUTDOWN",
```
yossigo, line 251: @tezc sorry for nit picking, but I see the pattern we're using elsewhere is `"RAFT.NODESHUTDOWN %d"`.

yossigo (approved):

raft.c
tezc: Right, that's unnecessary, fixed.

PR #137, opened 2021-08-30
Fix unsynchronized access on configuration change

tezc:
Access to raft library from Redis thread is not synchronized. Moved add/remove node id validation to the Raft thread.

### DIFF
--- a/raft.c
+++ b/raft.c
@@ -1327,12 +1327,17 @@ static void handleCfgChange(RedisRaftCtx *rr, RaftReq *req)
         goto exit;
     }
 
-    entry = raft_entry_new(sizeof(req->r.cfgchange));
-    entry->id = rand();
+    short type;
 
     switch (req->type) {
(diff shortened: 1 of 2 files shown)


sjpotter:
from what I can tell, this basically moves checking of node existence from redisraft (i.e. user of library) to libraft (library itself)?  seems reasonable, though related to what I proposed on other PR, about how to deal with unknown nodes.

tezc:
This is a thread-safety fix. Redis Main Thread calls `cmdRaftNode()`, inside it, while validating node ids, it reaches libraft objects without synchronization with the Raft Thread. That leads to undefined behavior/crash.

sjpotter:
ah, I see.  yea, that's bad.  no raft_ functions should be called from it.  it should all be handled by the dedicated thread.  cursory check, seems to be that you found the only usage in the prat of the module that runs in the main thread.  now I understand it better, so can look at it again

sjpotter (approved):

PR #139, opened 2021-09-07
Shutdown gracefully

tezc:
This PR is just for discussion, an attempt to implement graceful shutdown. There is one problem : 

1 - Registered to the shutdown event. Callback function will be called by Redis main thread. In callback function, it sends an event to the Redisraft thread to say go shutdown and it will wait until then.

```c
RRStatus RedisRaftShutdown(RedisModuleCtx *ctx, RedisRaftCtx *rr)
{
   // Send shutdown event
    uv_async_send(&rr->shutdown_sig);

   // Wait for RedisRaft thread to join
    if (uv_thread_join(&rr->thread) < 0) {
        RedisModule_Log(ctx, REDIS_WARNING, "Failed to stop redis_raft thread");
        return RR_ERROR;
    }
    ......
}
```

2- When Redis main thread calls this hook, it will be here : 

https://github.com/redis/redis/blob/74d9a35621d66a816e1c30de9cd84c461297495a/src/server.c#L4386

At this point, Redis main thread already called `afterSleep()` and acquired the module lock : 

```c
void afterSleep(struct aeEventLoop *eventLoop) {
    UNUSED(eventLoop);

    /* Do NOT add anything above moduleAcquireGIL !!! */

    /* Acquire the modules GIL so that their threads won't touch anything. */
    if (!ProcessingEventsWhileBlocked) {
        if (moduleCount()) moduleAcquireGIL();
    }
}
```

3 - When Redis main thread calls shutdown callback, RedisRaft thread might be here  : https://github.com/RedisLabs/redisraft/blob/befd50d88403c3f6c674a162c4c60795c9b7a5d3/raft.c#L181

Then it will be a deadlock. Redis main thread holds the lock and waits for Redisraft thread. Redisraft thread waits for Redis main thread to release the lock.  Do you see a way to avoid this?


yossigo:
@tezc I think it makes sense that the module GIL (and thus the dataset) is locked when `REDISMODULE_EVENT_SHUTDOWN` fires, because we don't modules modifying the dataset at this point.

I think we may be missing a `REDISMODULE_EVENT_PREPARE_FOR_SHUTDOWN` that can fire as soon as `prepareForShutdown` gets invoked and, unlike the other event, fires with the GIL unlocked to give background threads a chance to complete.

A module would then --

* Catch the `PREPARE_FOR_SHUTDOWN` hook
* Set termination flag for background threads
* Join all threads
* Return

Maybe @oranagra have some other thoughts about this as well.

oranagra:
I think i did suggest somewhere to issue two events during shutdown. one before the rdbSave and one after.
but back then someone convinced me that one at the end is enough.
maybe it has something to do with the fact we have the `REDISMODULE_SUBEVENT_PERSISTENCE_SYNC_RDB_START` event.

regarding the GIL, something similar to this was discussed in https://github.com/redis/redis/issues/6860#issuecomment-748835414 (note item 3 at the top).
Note that Meir's suggesting that a module can release the GIL, wait for the background tasks to finish and then re-acquire the GIL before returning.

the argument there against calling an event with a GIL unlocked is different, but it's still dangerous and odd (will require careful documentation), so maybe we can consider the same solution proposed there (an advanced module doing something special, rather than redis exposing a dangerous API)?

yossigo:
@oranagra The discussion you mentioned refers to another scenario, where the thread already holds the GIL and needs to preempt a fork (IIUC). The scenario @tezc refers to is different - we have a background thread trying to acquire the GIL while we're notified of a shutdown and would like to tell the thread to exit.

oranagra:
@yossigo in that other scenario, when redis calls `fork` for bgsave, it also holds the GIL, and we wanted to let the modules' threads drain their jobs before fork happens. so i think it's the same.
so i'm suggestion the module uses that shutdown signal to unlock the GIL and wait for the threads to join.

yossigo:
@oranagra Wouldn't having a pre-event that fires without the GIL be much safer?

oranagra:
we'll need to clearly document that this event is fired with GIL unlocked, and the documentation is often overlooked.
i think the approach of letting the module unlock the GIL itself, is more explicit, and puts the responsibility on the module.

also, maybe i'm not fully aware of what you wanna do in the pre-event, do you do anything that's gonna change the generated RDB file?
if not, then why do you need a pre-event?
if yes, then isn't `REDISMODULE_SUBEVENT_PERSISTENCE_SYNC_RDB_START` better?

yossigo:
@oranagra We already have a function that unlocks the GIL, but calling it from within a hook that never acquired the GIL to begin with would seem really wierd.

The pre-event would do something simple like
```
    thread_terminate = 1;
    join_all_threads().
```

oranagra:
@yossigo is there any reason why you need that join to happen in the pre-event?
what is it "pre" of? pre-save? pre-stopping-to-read-client-requests?

if the only reason you wanna add that pre-event is because you need another one in which the GIL is unlocked, and we can't define pre to what it is, maybe it should be named SHUTDOWN_NO_GIL or such.

anyway, my view is that it's safer to have the module explicitly unlock the GIL within the callback, than provide a callback that's called with the GIL unlocked. in that case the module is more explicit and is aware of the implications of what it's doing.

MeirShpilraien:
Why not thinking about it as an events-oriented? Like if Redis wants to shut down it tells all modules (that was asked to receive a shutdown event) that it wants to shut down, from here the modules should call Redis back when they are OK with the shutdown. Only when Redis got the shutdown OK from all the modules it knows it can perform the shutdown. I think this approach is safer with respect to deadlocks and also provides a cleaner API (like no need to assume that some events are fire with the lock and others are not), WDYT?

yossigo:
@oranagra There is already an API call to unlock the GIL, I just think that relying on a scheme where Actor 1 acquires a lock, Actor 2 releases it, does something re-acquires it before returning control to Actor 1 is an anti-pattern.

The "pre" hook is a "pre-shutdown" hook, i.e. last chance for the module to initiate cleanups before shutdown takes place. In a way it reflects the way shutdown_asap flag is used now in Redis itself.

@MeirShpilraien This can be an interesting alternative. It makes me think we may need to think about it in a bigger context, for example to expose things like `SHUTDOWN` vs `SHUTDOWN NOSAVE` as part of the contract with the module.

yossigo:
@oranagra and I discussed this briefly today and agreed on some observations:

* The idea proposed by @MeirShpilraien is a bit complex, addresses only the issue of shutdown but not other scenarios that may involve hooks and experience the same problem.
* Adding a pre-shutdown hook is also potentially a limited solution to a greater problem.
* Providing a mechanism where a module can unlock the GIL even though it did not acquire the lock in the first place can be too easy to abuse.

A safer and more generalized solution is to provide a `RM_SubscribeToServerEventsEx` function that can take an additional flags argument and request the event to be delivered with the GIL unlocked. Benefits are:

* A generalized solution for all events
* Not exposing GIL unlock
* Modules can still explicitly lock (and later unlock) the GIL if they need to, and will never implicitly get a callback with the GIL unlocked or need to rely on RTFM.

oranagra:
Maybe we don't want every event to support this new mode.
e.g. `RedisModuleEvent_ClientChange` and `RedisModuleEvent_CronLoop`.
please note that in most cases it would mean redis needs to explicitly release the GIL before calling, so it would be inefficient if then the module immediately locks it.

besides all that, we must expose a new `RedisModuleEvent_Fork` with PRE, POST_CHILD and POST_PARENT.
the reason we didn't do that so far is because we assumed that modules that need it will use `atfork` and unlock the GIL themselves if they want to.

oranagra:
FYI: https://github.com/redis/redis/pull/9872, maybe we should tap into that mechanism, or ask Viktor to add a module API in his PR.

PR #140, opened 2021-09-12
Fix recent redis-testsuite failures.

yossigo:
*  Improve error message compatibility.

Before this commit, a generic error was produced for RM_Call() failures.
Now we're evaluating the returned error code and produce an error
message that is more compatible with Redis.

* Skip recent tests that are incompatible.

### DIFF
--- a/raft.c
+++ b/raft.c
@@ -134,13 +134,32 @@ static void executeRaftRedisCommandArray(RaftRedisCommandArray *array,
         enterRedisModuleCall();
         RedisModuleCallReply *reply = RedisModule_Call(
                 ctx, cmd, "v", &c->argv[1], c->argc - 1);
+        int ret_errno = errno;
         exitRedisModuleCall();
 
         if (reply_ctx) {
             if (reply) {
                 RedisModule_ReplyWithCallReply(reply_ctx, reply);
             } else {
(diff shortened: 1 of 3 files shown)


raft.c
```diff
@@ -134,13 +134,32 @@ static void executeRaftRedisCommandArray(RaftRedisCommandArray *array,
...
+                    case EINVAL:
+                        snprintf(errmsg, errmsg_len, "ERR wrong number of arguments for '%.*s' command",
+                                trunc_cmdlen, cmd);
+                        break;
+                    default:
+                        snprintf(errmsg, errmsg_len, "ERR failed to execute command '%.*s'",
+                                trunc_cmdlen, cmd);
+                }
+                RedisModule_ReplyWithError(reply_ctx, errmsg);
+                RedisModule_Free(errmsg);
```
sjpotter, line 162: this seems reasonable, but should redis itself have a sprintf style ReplyWithErrorf() function?

sjpotter (approved):

raft.c
yossigo: Yes it could, or alternatively just have `RedisModule_ReplyWithErrorString` that takes a `RedisModuleString`. But I don't think it really matters.

PR #143, opened 2021-09-13
Fix raft_recv_entry error handling

tezc:
Currently :

```c
void handleRedisCommand(RedisRaftCtx *rr,RaftReq *req)
{
....
    entryAttachRaftReq(rr, entry, req);
    int e = raft_recv_entry(rr->raft, entry, &req->r.redis.response);
    raft_entry_release(entry);

    if (e != 0) {
        replyRaftError(req->ctx, e);
        goto exit;
    }
....
}
```

If `raft_recv_entry()` fails, `entry` will be freed in the next line.  It will also release `req` as we've attached `req` to the `entry`. Then, it will enter the next if check and try to use deallocated `req` there. This causes undefined behavior and crashes.

### DIFF
--- a/raft.c
+++ b/raft.c
@@ -49,17 +49,30 @@ static RedisModuleDict *multiClientState = NULL;
 
 /* ------------------------------------ Common helpers ------------------------------------ */
 
+static RaftReq *entryDetachRaftReq(RedisRaftCtx *rr, raft_entry_t *entry)
+{
+    RaftReq* req = entry->user_data;
+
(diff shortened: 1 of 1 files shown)


sjpotter:
I think I did something similiar before, but was part of a mega pr that was dropped.  seems fine, want to go over it again though, for some reason I am thinking the detach func can be used in other places as well to remove duplicate code, but perhaps I'm misremembering

sjpotter (approved):

PR #144, opened 2021-09-15
Increment term first before becoming leader in initCluster()

tezc:
Not a big problem but in log it shows "term 0" as the current term, it is slightly confusing : 

```
1246350:M 15 Sep 2021 07:28:24.088 * <redisraft> State change: Node is now a leader, term 0
```


sjpotter:
the question is if this should be in libraft, my gut is that it should be.  users shouldn't have to do this.

yossigo (approved):
Makes sense.

I agree with @sjpotter it may make sense to have some helper function to encapsulate these initialization details in the library, but it's a different topic.

PR #145, opened 2021-09-15
Remove unnecessary "-" from error reply messages

tezc:
RedisModule_ReplyWithError() already prepends "-".

### DIFF
--- a/common.c
+++ b/common.c
@@ -54,20 +54,20 @@ void replyRaftError(RedisModuleCtx *ctx, int error)
 
     switch (error) {
         case RAFT_ERR_NOT_LEADER:
-            RedisModule_ReplyWithError(ctx, "-ERR not leader");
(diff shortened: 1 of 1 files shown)


sjpotter:
I'm not sure this is correct, as its about the redis protocol

https://redis.io/topics/protocol

@yossigo would know the best

yossigo:
@sjpotter The protocol uses a `-` character, but modules don't need to specify it explicitly.

yossigo (approved):

dipAch:
I think @tezc is correct. The RM_ReplyWithError API prepends the **'-'** to the error text.

We can refer the redis module implementation here, https://github.com/redis/redis/blob/03fcc211de4cf7874c6eff58da3c6f67626ada5a/src/module.c#L1783

PR #150, opened 2021-09-22
Fix snapshot corruption

tezc:
Similar to https://github.com/RedisLabs/redisraft/pull/149, looks like snapshot data should be copied as well ?


yossigo:
@tezc Yes, spot on.
BTW Practically speaking this is less likely to get corrupted because larger `RedisModuleString` (i.e. `robj` pointing to an `sds`) are less likely to end up being compacted.

yossigo (approved):

PR #157, opened 2021-09-23
update redisraft to have libraft with timeoutnow

sjpotter:
this just pulls in latest libraft and adds the transfer_leader to requestvote send/recv rpc handlers.  Because there's no way to set it, they will currently always be 0/false until we add the redisraft support to handle it.


raft.c
```diff
@@ -671,7 +672,7 @@ raft_cbs_t redis_raft_callbacks = {
     .persist_vote = raftPersistVote,
     .persist_term = raftPersistTerm,
     .log = raftLog,
-    .log_get_node_id = raftLogGetNodeId,
+    .get_node_id = raftLogGetNodeId,
```
sjpotter, line 675: this is actually concerning.  did I somehow mess up libraft and change the name or in the midst of rebases, a change got reverted by me that we didn't catch?

tezc (approved):

PR #159, opened 2021-09-24
Copy Cmocka to the source tree

tezc:
Cmocka git server is down since yesterday, so build and tests are failing : https://git.cryptomilk.org/projects/cmocka.git

Looks like that's a personal address, not a reliable one. Should we copy Cmocka to source tree to avoid issues? Any downside here? (Personally, I'd copy all dependencies into to the project but others are on Github at least).


sjpotter:
just noting, its back up now
if we don't want to mirror it, this might be a better source https://gitlab.com/cmocka/cmocka/tree/master

yossigo:
@tezc We can copy it, possibly use git subtree to make updates easier.

PR #160, opened 2021-09-24
Fix undefined behavior while generating node id

tezc:
- (1 << 31)  Shifting 31 times causes integer overflow. (1u << 31) solves the issue.
- Also, added explicit cast to make compiler happy.


sjpotter:
seems like we are over complicaitng this.  why not just cast to int and use abs() ?

tezc:
@sjpotter Casting unsigned integers greater than INT_MAX to integer is undefined behavior as well.

sjpotter (approved):

PR #161, opened 2021-09-24
Prevent crash if leader is asked to send snapshot to itself

tezc:
Passing leader's own id as "sendsnapshot" target leads to crash. Extra check prevents that.

### DIFF
--- a/raft.c
+++ b/raft.c
@@ -2010,6 +2010,11 @@ static void handleDebugSendSnapshot(RedisRaftCtx *rr, RaftReq *req)
         goto exit;
     }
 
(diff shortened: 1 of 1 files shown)


sjpotter:
seems fine, I'm just wondering under what circumstances this would happen?  i.e. its only through RAFT.DEBUG so the root cause would be user error and we just want to prevent that? (that's fine, just wondering if I'm missing anything else)

tezc:
@sjpotter Yes, just to prevent user error on raft.debug command

sjpotter (approved):

PR #162, opened 2021-09-26
Better handle unsupported commands and fix tests.

yossigo:
* Streams and blocking commands are now explicitly rejected.
* Update Redis test suite configuration to skip unsupported commands,
  utilize test skipping by regex.


sjpotter:
why are blocking commands unsupported?  I'd think they would be easy to support?

yossigo:
@sjpotter It requires breaking the command implementation into a conditional blocking part and a non-blocking part.
For example, we can't simply push `BLPOP` into the Raft log and propagate it because if there's nothing to pop it will just block on all cluster nodes. Instead, we need to:

* Locally block the client until the list is non-empty
* When blocking condition is satisfied (list is non-empty), propagate a non-blocking version (`LPOP`)
* When committed and apply, return a reply to the user.

And of course we need to consider the case where multiple clients are blocking for the same condition, but only one can be satisfied.

I don't think it's terribly complex to handle that but unlike how most commands are supported (simply passed on to Redis on apply), blocking commands will require a more specific implementation.

sjpotter:
ok, that makes sense.

sjpotter (approved):

sjpotter:
per your second bullet, it would seem we would want redis to expose a block on condition command to make it easier to build the desired functionality?  (i.e. block, but don't modify underlying data structure)

PR #163, opened 2021-09-30
Remove submodules

yossigo:
This PR migrates the project from using git submodules to git subtrees. The main benefits are:

* Easier to build, no need to independently fetch submodules
* Easier to branch and test experiment with changes across components (mainly `deps/raft`)

After this is merged, new versions of `deps/raft` should be updated using the following `git` command:

```
git subtree pull --squash --prefix=deps/raft https://github.com/redislabs/raft <commit sha>
```

This operation essentially creates two commits: one commit with the new changes, and a merge commit. Because of this, I propose to adopt the following workflow:

* No direct changes to subtrees
* PRs with subtree changes should be rebased and not squash merged
* If possible, use a dedicated PR for subtree changes
* If not possible (due to dependencies/breaking changes), make sure the non-subtree commit(s) follow the squash-merge convention and directly reference the PR in the commit message. Because the PR is not squash merged, not doing this will make it hard to track commits to PRs.


tezc:
Just a note : 
https://github.com/jepsen-io/redis/blob/c23c7c125e431d58e85754986b3ce6fc86688f69/src/jepsen/redis/db.clj#L98

Jepsen test makes submodule calls. Looks like submodule commands return 0 when submodule does not exist. It should be okay for now. We can update Jepsen repo later sometime.

tezc (approved):

PR #164, opened 2021-10-04
libraft and redisraft changes needed for e2e leader transfer

sjpotter:


deps/raft/include/raft_private.h
```diff
@@ -118,8 +118,10 @@ typedef struct {
 
     raft_node_id_t node_transferring_leader_to; // the node we are targeting for leadership
     long transfer_leader_time; // how long we should wait for leadership transfer to take, before aborting
+    int sent_timeout_now; // if we've already sent a leadership transfer signal
 
     int timeout_now;
+    int timed_out;
```
tezc, line 124: Are you trying to prevent second election with this variable? Can you describe why do we need it?

raft.c
```diff
@@ -1339,6 +1411,91 @@ void RaftReqHandleQueue(uv_async_t *handle)
...
+{
+    raft_node_id_t transfer_id = raft_get_transfer_leader(raft);
+    raft_node_id_t leader = raft_get_leader_id(raft);
+
+    if (!redis_raft.transfer_req) {
+        LOG_ERROR("leader transfer update: but no req to correlate it to!");
+        return;
+    }
+
+    if (!transfer_id) {
```
tezc, line 1424: This will be RAFT_NODE_ID_NODE, not zero and check will not work. A bit more than that, I think we wouldn't call this callback if there is no transfer leader operation. So, we should be okay even without this check, maybe delete it?

raft.c
```diff
@@ -1339,6 +1411,91 @@ void RaftReqHandleQueue(uv_async_t *handle)
...
+
+    switch (state) {
+        case RAFT_STATE_LEADERSHIP_TRANSFER_EXPECTED_LEADER:
+            RedisModule_ReplyWithSimpleString(redis_raft.transfer_req->ctx, "OK");
+            break;
+        case RAFT_STATE_LEADERSHIP_TRANSFER_UNEXPECTED_LEADER:
+            RedisModule_ReplyWithError(redis_raft.transfer_req->ctx, "ERR: transferred to unexpected leader");
+            break;
+        case RAFT_STATE_LEADERSHIP_TRANSFER_TIMEOUT:
+            RedisModule_ReplyWithError(redis_raft.transfer_req->ctx, "ERR: transfer timed out");
```
tezc, line 1438: Just a minor thing, IIUC convention is "ERR error message". Looks like some clients strips "ERR" part and present rest of the error messages to users. So, we may want to use error message without ":"

raft.c
```diff
@@ -1339,6 +1411,91 @@ void RaftReqHandleQueue(uv_async_t *handle)
...
+        RedisModule_ReplyWithError(redis_raft.transfer_req->ctx, "ERR: no record of a leader transfer");
+        goto exit;
+    }
+
+    switch (state) {
+        case RAFT_STATE_LEADERSHIP_TRANSFER_EXPECTED_LEADER:
+            RedisModule_ReplyWithSimpleString(redis_raft.transfer_req->ctx, "OK");
+            break;
+        case RAFT_STATE_LEADERSHIP_TRANSFER_UNEXPECTED_LEADER:
+            RedisModule_ReplyWithError(redis_raft.transfer_req->ctx, "ERR: transferred to unexpected leader");
```
tezc, line 1435: This sounds a bit like we transferred to a wrong node by mistake. Is it better to say something like "ERR another node elected as leader"

raft.c
```diff
@@ -1339,6 +1411,91 @@ void RaftReqHandleQueue(uv_async_t *handle)
...
+                break;
+            case RAFT_ERR_LEADER_TRANSFER_IN_PROGRESS:
+                RedisModule_ReplyWithError(req->ctx, "ERR transfer already in progress");
+                break;
+            case RAFT_ERR_INVALID_NODEID:
+                e = RedisModule_CreateStringPrintf(req->ctx, "ERR invalid node id: %d", req->r.node_to_transfer_leader);
+                RedisModule_ReplyWithError(req->ctx, RedisModule_StringPtrLen(e, NULL));
+                break;
+            default:
+                e = RedisModule_CreateStringPrintf(req->ctx, "ERR unknown error transferring leader: %d", err);
```
tezc, line 1473: I think we need to deallocate these strings. I think you can use local buffer with snprintf instead.

raft.c
```diff
@@ -1339,6 +1411,91 @@ void RaftReqHandleQueue(uv_async_t *handle)
  * Implementation of specific request types.
  */
 
+static void handleTransferLeaderComplete(raft_server_t *raft, raft_transfer_state_e state)
+{
+    raft_node_id_t transfer_id = raft_get_transfer_leader(raft);
+    raft_node_id_t leader = raft_get_leader_id(raft);
```
tezc, line 1417: Unused variable, leftover?

raft.c
```diff
@@ -665,6 +726,15 @@ static char *raftMembershipInfoString(raft_server_t *raft)
     return buf;
 }
 
+static void handleTransferLeaderComplete(raft_server_t *raft, raft_transfer_state_e state);
+
+static void raftNotifyTransferEvent(raft_server_t *raft, void *user_data, raft_transfer_state_e state)
+{
+    if (redis_raft.transfer_req) {
+        handleTransferLeaderComplete(raft, state);
```
tezc, line 734: Same "if check" is also inside handleTransferLeaderComplete() with some error logic but it will not reach there. Btw, maybe just copy handleTransferLeaderComplete() contents into this function?

deps/raft/include/raft_private.h
sjpotter: yes.  its to not call election multiple times through, we can't reset timeout_now yet, as needed for the flag to requestvote.

raft.c
sjpotter: ack

raft.c
```diff
@@ -454,6 +457,64 @@ static int raftSendAppendEntries(raft_server_t *raft, void *user_data,
...
+                              node, argc, (const char **)argv, argvlen) != REDIS_OK) {
+        NODE_TRACE(node, "failed timeout now");
+    } else{
+        NodeAddPendingResponse(node, false);
+    }
+
+    RedisModule_Free(argv);
+    RedisModule_Free(argvlen);
+
+    return 0;
```
tezc, line 515:
```suggestion
    Node *node = raft_node_get_udata(raft_node);

    if (!ConnIsConnected(node->conn)) {
        NODE_TRACE(node, "not connected, state=%s", ConnGetStateStr(node->conn));
        return 0;
    }

    if (redisAsyncCommand(ConnGetRedisCtx(node->conn), handleTimeoutNowResponse,
                          node, "RAFT.TIMEOUT_NOW") != REDIS_OK) {
        NODE_TRACE(node, "failed timeout now");
    } else {
        NodeAddPendingResponse(node, false);
    }

    return 0;
```

raft.c
sjpotter: ok, done

raft.c
sjpotter: sort of, I wasn't sure if it should be included in the unexpected leader election message

raft.c
sjpotter: removed

sjpotter: ok, i'm unsure.  I like the fact that the raftreq "handlers" and the callback are separated, but I can combine that if desired.  but you are right, the check isn't neccessary (and in fact counterproductive)

tezc: Okay, however you want.

deps/raft/include/raft_private.h
tezc: I was thinking about this, how about in raft_recv_requestvote(); we make this change :

```c
    /* Reject request if we have a leader */
    if (vr->prevote &&                                      << -- prevote check here
        !vr->transfer_leader &&                  
        me->leader_id != RAFT_NODE_ID_NONE &&
        me->leader_id != vr->candidate_id &&
        me->timeout_elapsed < me->election_timeout) {
        goto done;
    }
```

We can do leader stickiness test only for prevotes. So, in this way, we can avoid timed_out flag. 

```c

if ((me->election_timeout_rand <= me->timeout_elapsed || me->timeout_now &&
        !raft_snapshot_is_in_progress(me_))
    {
        int e = raft_election_start(me_);   /// me->timeout_now will be used for prevotes.
        if (0 != e)
            return e;
    }
   
    me->timeout_now = 0;  // We can clear timeout_now
```

redisraft.c
```diff
@@ -129,7 +129,63 @@ static int cmdRaftNode(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
...
+
+/* RAFT.TIMEOUT_NOW
+ *   instruct this node to force an election
+ * Reply:
+ * ???
+ */
+
+static int cmdRaftTimeoutNow(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
+{
+    LOG_ERROR("cmdRaftTimeoutNow: enter");
```
tezc, line 174: leftover? Also there is one line before return

raft.c
tezc: But in general, local buffer is used to generate strings to avoid malloc/free in the codebase. If there is no downside, please consider doing something like : 
```c
char buf[128];
...
snprintf("ERR invalid node id: %d", req->r.node_to_transfer_leader);
RedisModule_ReplyWithError(req->ctx, buf);
```

deps/raft/include/raft_private.h
sjpotter: I'm not sure I agree with that logic change.  need to think it through.  i.e. I always viewed prevote as more of a connectivity check / "I could conceptually vote for you for leader, but not guaranteed", so the sticky leader is more important on the actual vote, not prevote.  but perhaps I'm wrong

raft.c
sjpotter: ok, understood

deps/raft/include/raft_private.h
sjpotter: with that said, perhaps the way to do it is not have prevote handler ignore sticky, but return the "transfer leader" flag in response, and when candidate gets responses from prevote, it uses that to send the right transfer_leader flag.  though, this would be a bit more invasive.  right now raft_send_requestvote() sets the transfer_leader flag based on me->timeout_now, what I'm proposing is that send_requestvote() func would take an extra arg that is transfer_leader and uses that to set the structs transfer_leader member.

so raft_become_candidate() will take it as a flag also (from raft_recv_requestvote_response) so it can pass it to raft_send_requestvote and raft_become_precandidate() will take it from me->timeout_now.

my only Q would be do we need some sort of assert in raft_recv_requestvote_response to ensure that all responses are transfer_leader or not or can we rely on just last one being accurate?

sjpotter: ok, I'm going to add a commit that is more complicated as I described above, but can go with your idea if you think its better.

tezc: It should work, hopefully :) I don't see a problematic scenario and I just find this better because it might be the smallest change among other options. No strong opinion though.

raft.c
```diff
@@ -665,6 +713,15 @@ static char *raftMembershipInfoString(raft_server_t *raft)
     return buf;
 }
 
+static void handleTransferLeaderComplete(raft_server_t *raft, raft_transfer_state_e state);
+
+static void raftNotifyTransferEvent(raft_server_t *raft, void *user_data, raft_transfer_state_e state)
+{
+    if (redis_raft.transfer_req) {
```
tezc, line 720: This check is also inside handleTransferLeaderComplete(). Forgot it? :)

tezc (approved):

Issue #107, opened 2021-05-17
Radis raft panic


zouyonghao:
```
103913:M 17 May 2021 13:17:50.850 # <redisraft> 

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
REDIS RAFT PANIC
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

Failed to update log header: raftlog2.db: No such file or directory


=== REDIS BUG REPORT START: Cut & paste starting from here ===
103913:M 17 May 2021 13:17:50.850 # Redis 255.255.255 crashed by signal: 6, si_code: -6
103913:M 17 May 2021 13:17:50.850 # Killed by PID: 103913, UID: 1000
103913:M 17 May 2021 13:17:50.850 # Crashed running the instruction at: 0x7f189b8cafb7

------ STACK TRACE ------
EIP:
/lib/x86_64-linux-gnu/libc.so.6(gsignal+0xc7)[0x7f189b8cafb7]

Backtrace:
/home/zyh/redis/src/redis-server *:5002(sigsegvHandler+0x18c)[0x63132c]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x12980)[0x7f189bea7980]
/lib/x86_64-linux-gnu/libc.so.6(gsignal+0xc7)[0x7f189b8cafb7]
/lib/x86_64-linux-gnu/libc.so.6(abort+0x141)[0x7f189b8cc921]
/home/zyh/redisraft/redisraft.so(updateLogHeader+0xab)[0x7f189819fe86]
/home/zyh/redisraft/redisraft.so(RaftLogSetTerm+0x35)[0x7f18981a11af]
/home/zyh/redisraft/redisraft.so(+0x28a42)[0x7f1898193a42]
/home/zyh/redisraft/redisraft.so(raft_set_current_term+0x31)[0x7f18981b0571]
/home/zyh/redisraft/redisraft.so(raft_become_candidate+0x5a)[0x7f18981addba]
/home/zyh/redisraft/redisraft.so(raft_periodic+0xa5)[0x7f18981b0385]
/home/zyh/redisraft/redisraft.so(+0x29d6c)[0x7f1898194d6c]
/home/zyh/redisraft/redisraft.so(uv__run_timers+0x58)[0x7f18981bc0d8]
/home/zyh/redisraft/redisraft.so(uv_run+0x92)[0x7f18981c0682]
[...]

PR #168, opened 2021-10-12
seek within log file handle, not open/close for seeing purposes

sjpotter:
try to remove flakiness that can occur with closing/reopening file to seek to beginning/end and just use fseek()


sjpotter:
motivated by the issue reported here: https://github.com/RedisLabs/redisraft/issues/107 where the file was failing to reopen

PR #170, opened 2021-10-13
make node removal synchronous

### LINKED ISSUE #56: Make membership operations synchronous
When calling `RAFT.CLUSTER JOIN` or `RAFT.NODE REMOVE`, ideally the user experience should be synchronous and return when the operation completes.

sjpotter:
client will block until entry for node removal is comitted

closes #56

### DIFF
--- a/raft.c
+++ b/raft.c
@@ -1608,8 +1608,7 @@ static void handleCfgChange(RedisRaftCtx *rr, RaftReq *req)
             // block until removed, so don't reply here
             return;
         default:
-            replyRaftError(req->ctx, e);
-            break;
+	    assert(0);
     }
 
 exit:


raft.c
```diff
@@ -1538,6 +1548,8 @@ static void handleCfgChange(RedisRaftCtx *rr, RaftReq *req)
     msg_entry_response_t response;
     int e;
 
+    printf("Hello world config change\n");
```
tezc, line 1551: Leftover?

raft.c
```diff
@@ -1565,6 +1577,8 @@ static void handleCfgChange(RedisRaftCtx *rr, RaftReq *req)
                 goto exit;
             }
 
+            printf("removing node %d\n", req->r.cfgchange.id);
```
tezc, line 1580: Leftover?

raft.c
```diff
@@ -1575,21 +1589,33 @@ static void handleCfgChange(RedisRaftCtx *rr, RaftReq *req)
...
             RedisModule_ReplyWithArray(req->ctx, 2);
             RedisModule_ReplyWithLongLong(req->ctx, req->r.cfgchange.id);
             RedisModule_ReplyWithSimpleString(req->ctx, rr->snapshot_info.dbid);
-        }
+            break;
+        case RR_CFGCHANGE_REMOVENODE:
+            // block until removed, so don't reply here
+            return;
+        default:
+            replyRaftError(req->ctx, e);
```
tezc, line 1615: I think it can't fall into "default" case. Even it will somehow, 'e' is 0 always. Should we use assert(0) instead or just if check as if before?

PR #171, opened 2021-10-20
Update license and copyright notices.

yossigo:


natoscott:
@yossigo if I can help in any way, please let me know - you can reach me directly via nathans at redhat.com

This change will unfortunately mean redisraft cannot become part of any of the major Linux distributions like Debian and Fedora, and all their various enterprise derivations and off-shoots.  This has a significant impact on people like myself: I can no longer use this code and I'm increasingly hesitant in recommending to my colleagues and our customers to use Redis at all because of the contempt shown to the open source community and loss of trust (today its all of your important modules, tomorrow...?).

I understand there's likely to be internal company consideration going on there so perhaps it might help for the management folks involved to consider the substantial benefit being gained by their own reliance on open source.  RedisLabs would not exist without the open source projects it uses *without contributing back to* - the Linux kernel, drivers, compilers, user space services and tools, profilers, libraries, and on and on - ironically the very "problem" the self-serving RedisLabs modules license claims to address.

Anyway, I truly hope this situation can be resolved - for redisraft, the other modules, and the health of Redis in general.

PR #179, opened 2021-11-09
Add CLUSTER NODES command support.

sjpotter:
Intended mainly for clients that still use `CLUSTER NODES` and not `CLUSTER SLOTS`, and not so much for cluster administration which should rely on `RAFT.INFO`.

Many fields are not applicable for RedisRaft, but the output should be parser-compatible with Redis Cluster.

### DIFF
--- a/cluster.c
+++ b/cluster.c
@@ -857,6 +857,134 @@ static int addClusterSlotShardGroupNodeReply(RedisRaftCtx *rr, RedisModuleCtx *c
     return 1;
 }
 
+/* doesn't handle mutiple slot ranges or importing/exports yet */
+RedisModuleString * generateSlots(RedisModuleCtx *ctx, ShardGroup *sg)
+{
+    if (sg->start_slot != sg->end_slot) {
+        return RedisModule_CreateStringPrintf(ctx, "%d-%d", sg->start_slot, sg->end_slot);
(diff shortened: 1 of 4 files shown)


yossigo (commented):
@sjpotter I've pushed a commit with some trivial changes (instead of lots of small comments here), hope all of that makes sense.

Some test coverage is missing.

cluster.c
```diff
@@ -886,16 +895,16 @@ static void appendClusterNodeString(RedisModuleString *ret, char node_id[41], No
                                                            pong_recv,
                                                            epoch,
                                                            link_state,
-                                                           RedisModule_StringPtrLen(slots, NULL));
+                                                           (int) slots_len, slots_str);
```
sjpotter, line 898: is the point of this change to protect against bugs in the creation of the slot_str? where the pointer we get out of it isn't null terminated where we expect it to be?

sjpotter:
the above question is my only Q, I'll do the tests tomorrow, mostly mimicing what we have for cluster slots

cluster.c
yossigo: Yes, `RedisModuleString` is not formally guaranteed to be null terminated.

yossigo (approved):

PR #182, opened 2021-11-16
Multiple Slot Ranges + Consistent Shard Group IDs

sjpotter:
This PR set out to do 2 things

1) Expand ShardGroup to enable users to specify multiple non contiguous ranges of shards
2) Enable each ShardGroup to be named persistently by the individual's RedisRaft cluster's DB ID.

1 is to enable a future where we can dynamically remap which RedisRaft clusters provide each slot to enable slot migrations
2 is to enable a future where an external orchestrator can more easily inform each RedisRaft cluster about other clusters providing access to other shards

- This also required changing the serialization format (both at the redis level and the internal RedisRaft entry level
- it removed the old hslot config option and includes a new slot-config option that can take multiple ranges
- enables slot ranges to be tagged with a stable/migrating/importing tag that uses Redis's nomenclature for a future where slot are migrating from one computer to another.


raft.c
```diff
@@ -2343,27 +2345,73 @@ void handleShardGroupAdd(RedisRaftCtx *rr, RaftReq *req)
...
+     */
+    RedisModule_ReplyWithArray(req->ctx, 3);
+    RedisModule_ReplyWithCString(req->ctx, redis_raft.snapshot_info.dbid);
+    RedisModule_ReplyWithArray(req->ctx, sg->slot_ranges_num);
+    for(int i = 0; i < sg->slot_ranges_num; i++) {
+        ShardGroupSlotRange * sr = &sg->slot_ranges[i];
+        RedisModule_ReplyWithArray(req->ctx, 3);
+        RedisModule_ReplyWithLongLong(req->ctx, sr->start_slot);
+        RedisModule_ReplyWithLongLong(req->ctx, sr->end_slot);
+        RedisModule_ReplyWithLongLong(req->ctx, sr->type);
```
yossigo, line 2410: Consider either more explicit mapping of type to the wire format, or at least explicitly assign values to the enum - the way it's done now is a bit fragile IMHO.

cluster.c
```diff
@@ -86,29 +100,70 @@ RRStatus ShardGroupDeserialize(const char *buf, size_t buf_len, ShardGroup *sg)
...
+    *nl = '\0';
 
-    if (sscanf(s, "%u:%u:%u:%u", &sg->id, &sg->start_slot, &sg->end_slot, &sg->nodes_num) != 4)
+    if (sscanf(s, "%u:%u", &sg->slot_ranges_num, &sg->nodes_num) != 2) {
         goto error;
+    }
     s = nl + 1;
 
+    /* Read in slot ranges */
+    sg->slot_ranges = RedisModule_Alloc(sizeof(ShardGroupSlotRange) * sg->slot_ranges_num);
```
yossigo, line 133: Suggest to use RedisModule_Calloc and let it handle the potential integer overflow (this is unsanitized external input).

cluster.c
```diff
@@ -86,29 +100,70 @@ RRStatus ShardGroupDeserialize(const char *buf, size_t buf_len, ShardGroup *sg)
...
+        *nl = '\0';
+
+        if (sscanf(s, "%u:%u:%u", &r->start_slot, &r->end_slot, &r->type) != 3) {
+            goto error;
+        }
+        s = nl + 1;
+    }
+
+    /* read in shard group nodes */
     sg->nodes = RedisModule_Alloc(sizeof(ShardGroupNode) * sg->nodes_num);
```
yossigo, line 151: Same as above (LIBTYFI).

cluster.c
```diff
@@ -197,31 +258,47 @@ int compareShardGroups(ShardGroup *a, ShardGroup *b)
...
+        reply->element[1]->type != REDIS_REPLY_ARRAY || /* slots array */
+        reply->element[2]->type != REDIS_REPLY_ARRAY) { /* nodes array */
         return RR_ERROR;
     }
 
-    /* Validate node arguments count is correct */
-    int num_nodes = (reply->elements - 2) / 2;
-    if ((reply->elements - 2) != num_nodes * 2) {
-        return RR_ERROR;
+    strncpy(sg->id, reply->element[0]->str, RAFT_DBID_LEN);
```
yossigo, line 273: Consider doing this safer and drop the strncpy.

cluster.c
```diff
@@ -325,13 +408,13 @@ static void handleShardGroupResponse(redisAsyncContext *c, void *r, void *privda
...
         } else {
-            LOG_DEBUG("Received shardgroup %u reply.", sg->id);
+            LOG_DEBUG("Received shardgroup %s reply.", recv_sg.id);
             sg->use_conn_addr = true;
             sg->last_updated = RedisModule_Milliseconds();
             sg->update_in_progress = false;
 
             /* Issue update */
-            recv_sg.id = sg->id;    /* Copy ID to allow correlation */
+            strncpy(recv_sg.id, sg->id, sizeof(recv_sg.id));    /* Copy ID to allow correlation */
```
yossigo, line 417: Also here, better avoid strncpy.

cluster.c
```diff
@@ -597,18 +699,34 @@ RRStatus ShardingInfoValidateShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
 
 RRStatus ShardingInfoUpdateShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
 {
-    ShardingInfo *si = rr->sharding_info;
+    if (!strncmp(rr->snapshot_info.dbid, new_sg->id, sizeof(new_sg->id))) {
+        /* TODO: This is our cluster, so will have to handle add and removing nodes */
```
yossigo, line 703: What do you mean?

cluster.c
```diff
@@ -597,18 +699,34 @@ RRStatus ShardingInfoValidateShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
 
 RRStatus ShardingInfoUpdateShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
 {
-    ShardingInfo *si = rr->sharding_info;
+    if (!strncmp(rr->snapshot_info.dbid, new_sg->id, sizeof(new_sg->id))) {
```
yossigo, line 702: memcmp is probably better, it's fixed length anyway.

cluster.c
```diff
@@ -597,18 +699,34 @@ RRStatus ShardingInfoValidateShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
...
+        sg->nodes_num = new_sg->nodes_num;
+        sg->nodes = RedisModule_Realloc(sg->nodes, sizeof(ShardGroupNode) * sg->nodes_num);
+        memcpy(sg->nodes, new_sg->nodes, sizeof(ShardGroupNode) * sg->nodes_num);
+    }
+
+    return RR_OK;
+}
 
-    ShardGroup *sg = si->shard_groups[new_sg->id - 1];
+ShardGroup * getShardGroupById(RedisRaftCtx *rr, char id[RAFT_DBID_LEN+1])
```
yossigo, line 718: Probably better to take `const char *id` and if we need a null-terminated version, make a local copy on the stack, no?

cluster.c
```diff
@@ -623,13 +741,22 @@ RRStatus ShardingInfoAddShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
...
+
+    unsigned int sg_idx = si->shard_groups_num++;
     si->shard_groups = RedisModule_Realloc(si->shard_groups, sizeof(ShardGroup *) * si->shard_groups_num);
 
-    ShardGroup *sg = si->shard_groups[si->shard_groups_num-1] = RedisModule_Alloc(sizeof(ShardGroup));
-    sg->id = si->shard_groups_num;
-    sg->start_slot = new_sg->start_slot;
-    sg->end_slot = new_sg->end_slot;
+    ShardGroup *sg = si->shard_groups[sg_idx] = RedisModule_Alloc(sizeof(ShardGroup));
+    strncpy(sg->id, new_sg->id, sizeof(sg->id));
```
yossigo, line 755: Same as above.

cluster.c
```diff
@@ -726,6 +872,39 @@ void ShardingInfoInit(RedisRaftCtx *rr)
...
+        sg->slot_ranges_num++;
+    }
+    sg->slot_ranges = RedisModule_Calloc(sg->slot_ranges_num, sizeof(ShardGroupSlotRange));
+
+    char * token = strtok(str, ",");
+    for(int i = 0; i < sg->slot_ranges_num; i++) {
+        unsigned long val;
+        if ((pos = strchr(token, ':'))) {
+            *pos = '\0';
+            val = strtoul(token, NULL, 10);
```
yossigo, line 891: Do we assume this has already been validated?

cluster.c
```diff
@@ -860,16 +1046,35 @@ static int addClusterSlotShardGroupNodeReply(RedisRaftCtx *rr, RedisModuleCtx *c
...
+    if (sg->slot_ranges[0].start_slot != sg->slot_ranges[0].end_slot) {
+        ret = RedisModule_CreateStringPrintf(ctx, "%d-%d", sg->slot_ranges[0].start_slot, sg->slot_ranges[0].end_slot);
     } else {
-        return RedisModule_CreateStringPrintf(ctx, "%d", sg->start_slot);
+        ret = RedisModule_CreateStringPrintf(ctx, "%d", sg->slot_ranges[0].start_slot);
+    }
+
+    for (int i = 1; i < sg->slot_ranges_num; i++) {
+        RedisModuleString * tmp;
+        const char * str;
```
yossigo, line 1063:
```suggestion
        RedisModuleString *tmp;
        const char *str;
```

yossigo:
@sjpotter Did you take a look at the latest comments here?

redisraft.h
```diff
@@ -452,7 +467,8 @@ typedef struct ShardGroup {
  */
 typedef struct ShardingInfo {
     unsigned int shard_groups_num;       /* Number of shard groups */
-    ShardGroup **shard_groups;           /* Shard groups array */
+    ShardGroup **shard_groups;
+    RedisModuleDict *shard_group_map;    /* shard group id -> x in shard_groups[x] */
```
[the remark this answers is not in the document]
sjpotter: pondering it, would make the hash_slots_map[REDIS_RAFT_HASH_SLOTS] slower.  as either it have to be dict as well.  its open for discussion

redisraft.h
```diff
@@ -511,6 +527,7 @@ typedef struct RaftReq {
             int hash_slot;
             RaftRedisCommandArray cmds;
             msg_entry_response_t response;
+            int multi;
```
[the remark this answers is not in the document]
sjpotter: yes.  removing

raft.c
```diff
@@ -2343,27 +2345,73 @@ void handleShardGroupAdd(RedisRaftCtx *rr, RaftReq *req)
...
+void handleShardGroupUpdate(RedisRaftCtx *rr, RaftReq *req)
+{
+    /* Must be done on a leader */
+    if (checkRaftState(rr, req) == RR_ERROR ||
+        checkLeader(rr, req, NULL) == RR_ERROR) {
+        RedisModule_ReplyWithError(req->ctx, "failed, please check logs.");
+        goto exit;
+    }
+
+    /* TODO: validate/handle slot range changes:
```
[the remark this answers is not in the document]
sjpotter: it seems something that is expanding scope.  It would seem to be a followup PR.  we already test for normal overlap, this is new logic for handling migrating/importing state correctly.

cluster.c
sjpotter: switched

cluster.c
sjpotter: note quite sure I understand

cluster.c
sjpotter: note quite sure I understand

raft.c
sjpotter: what do you mean by this?  that I should validate the values?  what would you recommend?

cluster.c
sjpotter: changed it to char *, unsure there would ever be a need to copy, from what I can tell, its always a char * c string (hence the +1)

cluster.c
sjpotter: how would we avoid the strncpy in this case?

redisraft.h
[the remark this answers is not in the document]
sjpotter: I'm thinking we can do this, just that the hash_slots_map would point to a ShardGroup, but once a ShardGroup is allocated, it will never be freed (which sort of makes sense, it would just become a ShardGroup without any slots assigned to it), so if another ShardGroup owns the slot, we would just change the pointer its pointing to.

cluster.c
sjpotter: added a function to validate

redisraft.h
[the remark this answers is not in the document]
yossigo: @sjpotter Maybe I'm missing something here, what I had in mind was that both `shard_group_map` and `hash_slots_map` would hold pointers to `ShardGroup` structures. I don't see what is the problem to free a `ShardGroup`, as long as you also make remove it from the dict and remove its slots reference.

A slot that is not assigned to a `ShardGroup` can have a `NULL`.

raft.c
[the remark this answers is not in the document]
yossigo: Is it more than just a single function?

raft.c
yossigo: I mean use specific values in the enum and have a comment that indicate this is not just an in-memory enum, but something that gets serialized and therefore must be stable (i.e. nobody should reorder elements).

cluster.c
yossigo: It's a fixed length buffer, no need to treat it as a null terminated string. And if it was null terminated, but longer than N, `strncpy` is a risky function that is best avoided because it will produce non-null-terminated strings in the destination.

sjpotter: in our case, it should never be non null terminated, as are are always strncpy(LEN) when the fixed sized buffer is LEN+1

In practice, I think it can be considered fix size, but I'm wondering if we should protect against users passing in wrong sized strings in.  perhaps better in those cases to verify up front?

raft.c
sjpotter: so basically a switch there over all the values of the enum as a form of validation? (or just have fallthrough for all valid enum values and a panic in default?

raft.c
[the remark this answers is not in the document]
sjpotter: I'm just unsure how we do this in a coordinated manner for all the slot types (stable/migrating/importing)

redisraft.h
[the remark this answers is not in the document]
sjpotter: ok, that makes sense.  I think I can do that.

cluster.c
```diff
@@ -463,19 +553,38 @@ void ShardingInfoRDBSave(RedisModuleIO *rdb)
...
+                RedisModule_SaveStringBuffer(rdb, n->addr.host, strlen(n->addr.host));
+                RedisModule_SaveUnsigned(rdb, n->addr.port);
+            }
+
         }
+        RedisModule_DictIteratorStop(iter);
+    }
+
+    for (int i = 1; i < si->shard_groups_num; i++) {
     }
```
yossigo, line 590: Leftover?

cluster.c
yossigo: @sjpotter ?

cluster.c
```diff
@@ -623,13 +755,18 @@ RRStatus ShardingInfoAddShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
     if (ShardingInfoValidateShardGroup(rr, new_sg) != RR_OK)
         return RR_ERROR;
 
+    unsigned int * idx = RedisModule_Alloc(sizeof(unsigned int));
```
yossigo, line 760: Leftover?

cluster.c
yossigo: Remember `strncpy` treats `n` as the size of the buffer and will not append a null byte if `source length == n`.

config.c
```diff
@@ -189,18 +217,13 @@ static RRStatus processConfigParam(const char *keyword, const char *value,
...
-            goto invalid_value;
-        target->sharding_start_hslot = (int)val;
-    } else if (!strcmp(keyword, CONF_SHARDING_END_HSLOT)) {
-        char *errptr;
-        unsigned long val = strtoul(value, &errptr, 10);
-        if (*errptr != '\0' || val > INT32_MAX || !HashSlotValid((int) val))
-            goto invalid_value;
-        target->sharding_end_hslot = (int)val;
+    } else if (!strcmp(keyword, CONF_SLOT_CONFIG)) {
+        // FIXME: verify slot config
```
yossigo, line 224: Leftover?

raft.c
yossigo: I think at least bound check should be done.

sjpotter: something like this?

```
        assert(sr->type == SLOTRANGE_STABLE ||
              sr->type == SLOTRANGE_MIGRATING ||
              sr->type == SLOTRANGE_IMPORTING);
```

cluster.c
sjpotter: the buffer is 64+1, the str should be capped at 64.  should we just throw an assert in somewhere verifying that?

sjpotter: would this be sufficient?

```
    if (strlen(new_sg) >= sizeof(sg->id)) {
        return RR_ERROR;
    }
    strncpy(sg->id, new_sg->id, sizeof(sg->id));
```

sjpotter:
so, the TODO above, I rewrote the comment (stubbing it out still as need to handle shard slot changes in future).

the other 2, I gave thoughts to see if I undertand the issue correctly.

sjpotter:
I'm seeing the error https://github.com/RedisLabs/redisraft/runs/4608081361?check_suite_focus=true here in many places, I'm unsure how its happening.

yossigo (approved):

PR #183, opened 2021-11-18
Enable RedisRaft to support older versions of Redis

sjpotter:
Older versions don't support the `0` format modifier in `RM_Call()`, which is used to automatically match the call's protocol level to the current client's select version.

This commit auto-detects the version of Redis and will a compatible format. As a result, older versions will not have proper RESP3 support.


yossigo:
@sjpotter I think it's better to handle this at init time and not every time we send a command to Redis, for performance reasons.

sjpotter:
ok, pushed it into the redis_raft configuration struct

yossigo (commented):
@sjpotter I think this would probably fit RedisRaftCtx more than the configuration structure, which only holds user configuration.

yossigo (approved):

yossigo:
@sjpotter I've updated the top comment, please use it when squash merging.

PR #186, opened 2021-11-22
Make removed node shutdown even it's the only node in the cluster

tezc:
Make removed node shutdown even it's the only node in the cluster

In a single node cluster, remove operation of the last node will only be appended to the log and it will not be applied. As configuration changes are effective immediately after appending to the log, the last node can do a special check and decide to shutdown in this case.

### DIFF
--- a/raft.c
+++ b/raft.c
@@ -1626,6 +1626,8 @@ static void handleCfgChange(RedisRaftCtx *rr, RaftReq *req)
             RedisModule_ReplyWithSimpleString(req->ctx, rr->snapshot_info.dbid);
             break;
         case RR_CFGCHANGE_REMOVENODE:
+            /* Block until removed, so don't reply here. */
+
             /* Special case, we are the only node and our node has been removed */
             if (req->r.cfgchange.id == raft_get_nodeid(rr->raft) &&
                 raft_get_num_voting_nodes(rr->raft) == 0) {
--- a/raft.c
+++ b/raft.c
@@ -912,6 +912,20 @@ static void handleLoadingState(RedisRaftCtx *rr)
     }
 }
 
+static void shutdownAfterRemoval(RedisRaftCtx *rr)
+{
+    LOG_INFO("*** NODE REMOVED, SHUTTING DOWN.");
+
+    if (rr->config->raft_log_filename) {
+        RaftLogArchiveFiles(rr);
+    }
+    if (rr->config->rdb_filename) {
+        archiveSnapshot(rr);
+    }
+
+    exit(0);
+}
+
 static void callRaftPeriodic(uv_timer_t *handle)
 {
     RedisRaftCtx *rr = (RedisRaftCtx *) uv_handle_get_data((uv_handle_t *) handle);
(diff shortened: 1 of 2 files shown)


sjpotter:
so this would work, I want to think about a little if this is the best way, I'm wondering why the apply never gets called, as the apply of this should cause it to exit.

tezc:
Once we append, it's zero node cluster. As there are no voters, raft cannot commit the entry and apply it. I think we can say this is the last moment of a valid Raft cluster.

sjpotter:
we have the concept of a non voting leader (i.e. removed from cluster).
we have the concept of a 1 node cluster where entries are automatically comitted without waiting for votes to come in (as they can't)

this is just the ultimate form of those.  I'm wondering if we need to special case it, or if we can make the code that handles them handle this.

basically in raft_recv_entry() we have

```
    if (raft_is_single_node_voting_cluster(me_)) {
        raft_set_commit_idx(me_, raft_get_current_idx(me_));
    }
```

2 thoughts I have

1) should we apply the entry (generally) right away as well in the above case? (probably need raft_apply_all(), to make sure we get previous entries)
2) should voter of a single voter cluster be expanded to include a zero voter cluster (at least in the case of if this entry is a removal)

sjpotter:
another thought to limit the trouble of the raft_append_all() proposal I made above, currently we prevent entries from being recv'd by a "non leader" with this

```
    if (!raft_is_leader(me_))
        return RAFT_ERR_NOT_LEADER;

    if (raft_get_transfer_leader(me_) != RAFT_NODE_ID_NONE)
        return RAFT_ERR_LEADER_TRANSFER_IN_PROGRESS;
```

should a non voting leader also refuse entries?  or at least a non voting leader of a 0 voter cluster?  then while we would still call raft_apply_all() in the situation I described above, it could only happen once (i.e. when removing the last node).

at the end of the day, I'm unsure if this is cleaner that what you propose in the PR, but figure worth discussing.

tezc:
> should a non voting leader also refuse entries?

I think it can accept entries, it's okay according to Raft algorithm

>should a non voting leader also refuse entries? or at least a non voting leader of a 0 voter cluster?

Zero voter cluster leader can append a new entry but it won't be able to commit, we are safe :) 

I'd be nice to handle this in a single place when we apply the entry but as you said requires extra care in libraft and adds that special case check/checks into it. I think question is where to do that special case handling. I suggest we follow Raft paper strictly for libraft (When we append removal for the last node, we can't commit the entry as there is no voter anymore) and leave that special case handling for applications. This is a rare situation, apps can implement as in this PR if they want.

sjpotter:
if there are zero voters and hence you can't commit an entry, what's the point of allowing an entry to be appended?

tezc:
You just implement Raft algorithm, all corner cases are automatically handled or avoided. This is one of them, no harm comes from that behavior so no special handling is defined for that in Raft algorithm AFAIK. There is no point on accepting or rejecting it according to Raft.

sjpotter:
I guess my point was that it would simplify the special case.  its no longer a special case. if an entry is accepted for appending, it can be committed, even by a zero voter cluster (as wasn't zero voter when accepted, only become zero voter upon append).  with that said, I don't have strong opinions either way.

tezc:
I rather have that single check in app, keep lib strictly follow Raft and avoid adding complexity into it for a once in a lifetime scenario.

raft.c
```diff
@@ -1618,7 +1626,13 @@ static void handleCfgChange(RedisRaftCtx *rr, RaftReq *req)
             RedisModule_ReplyWithSimpleString(req->ctx, rr->snapshot_info.dbid);
             break;
         case RR_CFGCHANGE_REMOVENODE:
-            // block until removed, so don't reply here
```
sjpotter, line 1621: i'm wondering if the original comment should remain in some shape or form to allow future readers to understand why there is no reply here.

tezc: Didn't realize it, reverted

sjpotter (approved):

PR #192, opened 2021-12-09
Use Redis event loop API

tezc:
Use Redis event loop API instead of using libuv. 

Motivation for this change is performance. When we use Redis event loop, RedisRaft becomes single threaded, operates in Redis main thread's context. There are still threads for blocking operations (for fsync() and DNS resolution) but rest of the processing happens in Redis main thread. This way, we are not congesting for the GIL and not spending time on context switches. As a bonus, we also avoid some memory copies as we don't need to pass data between module thread vs Redis main thread. All these helps to achieve better performance numbers.

Also, pulled latest version of libraft which contains this improvement : https://github.com/RedisLabs/raft/pull/81

**Changes:**
- Added Redis event loop adapter for hiredis.
- Added a threadpool, currently, just to call getaddrinfo().
- Added a fsync thread. 
- Removed libuv.

**Needs to be done in follow up PRs:**
- No need to create RaftReq in many cases. We can avoid some allocation/free pairs. Leaving this to follow up PRs not make this PR any bigger. 

--------------

**Benchmark :**
c5.4xlarge, 7 shards, 3 Nodes
fsync latency ~ 3 millisecond
latency between nodes 600 microsecond to 1 millisecond

memtier_benchmark:
-t = thread count
-c = client count

| Quorum_read = yes           | Throughput (op/s) | Avr Latency (ms) |
| ----------------------------------- | -------------------------| ---------------------- |
| -t 4 -c 50                             | 198284                  |   6.75                  |
| -t 1 -c 1                               | 3764                      | 1.54                    |
| -t 8 -c 50                             | 340107                  |  7.92                   |
| -t 8 -c 100                           | 487558                  | 10.82                  |
| -t 4 -c 50 --data-size=1024 | 178421                  | 7.56                    | 

| Quorum_read = no            | Throughput (op/s) | Avr Latency (ms) |  Read Latency(ms) | Write Latency(ms)   |
| ----------------------------------- | -------------------------| ---------------------- | ------------------------- | -------------------------- |
| -t 4 -c 50                             | 416713                  |   3.14                  |  2.59783                 | 8.58357                   |
| -t 1 -c 1                               | 5768                      | 0.83                    |  0.62215                 | 2.88907                   |
| -t 8 -c 10                             | 327325                  |  1.51                   | 1.07117                  | 5.87560                   |
| -t 4 -c 30 --ratio 50:50         | 166949                  | 4.28                    |  0.85828                 | 7.70102                  | 
| -t 4 -c 10 --data-size=4096 | 144004                  | 1.70                    |  0.90232                 | 9.65836                  | 

------------------------

c5.4xlarge, 1 shard, 3 Nodes
| Quorum_read = yes           | Throughput (op/s) | Avr Latency (ms) |
| ----------------------------------- | -------------------------| ---------------------- |
| -t 4 -c 100                           | 102148                  |   7.82                  |
| -t 8 -c 100                           | 99411                    |   8.02                  |
| -t 8 -c 50                             | 72922                    |  5.57                   |
| -t 4 -c 50 --data-size=1024 | 40522                    | 4.97                    |


yossigo:
@tezc Had a quick first look - very interesting! Left a few questions already.
There is one thing that concerns me regarding the transition to a single thread and the interaction with Redis. Currently, our interfaces are:

* Redis main thread commands -> queue -> libuv loop
* Replies from other nodes -> hiredis callbacks -> libuv loop
* Timer events -> libuv loop

Each of these flows may end up having to interact with Redis (do `RM_Call()`), but this is synchronized because of the thread safe context and GIL locking.

Now, we'll no longer synchronize our access - but we still have many different paths that may lead to interacting with Redis:

* Redis main thread commands -> this is easy, because we get a `RedisModuleCtx` and Redis basically expects a module to interact with it.
* Replies from other nodes -> ae -> hiredis callbacks :: Is it always safe and correct to do an RM_Call here? Some analysis may be required.
* Timer events -> Same question as above. AFAIR modules getting called in a hook can assume the GIL is locked and that keyspace access is permitted, but we need to validate this.

If we're able to validate this and confirm we don't have any gaps, I don't see any other major concern with this approach (which has a lot of value).

redisraft.c
```diff
@@ -809,6 +829,68 @@ static void interceptRedisCommands(RedisModuleCommandFilterCtx *filter)
     RedisModule_CommandFilterArgInsert(filter, 0, raft_str);
 }
 
+static void flush()
+{
+    int e;
+    RedisRaftCtx *rr = &redis_raft;
+
+    if (rr->state == REDIS_RAFT_UP) {
+        if (!raft_is_leader(rr->raft)) {
```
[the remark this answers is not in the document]
tezc: Yeah, leader is responsible for that. Leader sends in entries in batches, follower writes entries to the log, call fsync and return response to that message.

deps/raft/src/raft_server.c
```diff
@@ -751,6 +753,10 @@ int raft_recv_appendentries_response(raft_server_t* me_,
     raft_node_set_next_idx(node, r->current_idx + 1);
     raft_node_set_match_idx(node, r->current_idx);
 
+    if (!me->auto_flush) {
```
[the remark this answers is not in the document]
tezc: Yes but we do commit index update check in `flush()` call once on each iteration already. We receive client operations and appendreq responses etc. and then before going back to "sleep"(actually going back to epoll_wait), we check if we can increment commit, if we should send appendentries etc.  So, we'll increment commit index a moment later anyway.

log.c
```diff
@@ -494,6 +512,14 @@ RaftLog *RaftLogCreate(const char *filename, const char *dbid, raft_term_t snaps
     ftruncate(fileno(log->file), 0);
     ftruncate(fileno(log->idxfile), 0);
 
+    /* This is required for older linux kernels.
```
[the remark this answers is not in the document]
tezc: Don't know when it got better but realized that even on my local up to date linux, I see parallel fsync increases latency of write() time to time. I'll do more tests later, considering sharding setups as well.

tezc:
@yossigo 

Regarding using redis thread, I think we are safe. Redis main loop acquires GIL in `afterSleep()` as the first thing and releases it as the last thing in `beforeSleep()`. 

Sequence is like :
1- epoll_wait()
2- afterSleep()                      // Acquire GIL as the first thing
3- processNetworkEvents(); // Responses from other nodes are processed here
4- processTimerEvents();     // Our timer events are processed here
5- beforeSleep()                   // Release GIL as the last thing
1- epoll_wait();

- If there is one module using Redis main thread and other one using its own thread, Redis main thread will lock/unlock GIL in beforeSleep()/afterSleep(). Other module will need to acquire GIL anyway, we are safe.
- If there are two modules using redis main thread, they'll be safe as they are using same thread and they won't be accessing dataset at the same time. 
  - Concurrent access should be fine, right?. e.g Module A's timer is fired, then Module B's and then Module A's again in the same event loop iteration. Same example applies to network events as well. Module A's network socket callback is called, then module B's, then module A's again etc.  This is actually same as current timer API, that can trigger timers from different modules in the mixed order.

yossigo:
@tezc I think your analysis is correct. We can double check this with @oranagra as part of reviewing the additional Module API for event loop registration. We just need to make it explicit that event callbacks fire with the GIL locked and a RedisModuleCtx that has no restrictions on calling commands.

oranagra:
I only read the last two posts (not sure if i'm missing some context, or questions to answer).
the analysts in that last post with regards to the GIL is correct.
if you have any other question, please be explicit, or just call me.

ae.h
```diff
@@ -0,0 +1,26 @@
+#ifndef __AE_H__
+#define __AE_H__
+#include "redisraft.h"
+
+#define AE_NONE 0       /* No events registered. */
+#define AE_READABLE 1   /* Fire when descriptor is readable. */
+#define AE_WRITABLE 2   /* Fire when descriptor is writable. */
+
+
+/* Pretend like ae lib for hiredis */
```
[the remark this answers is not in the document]
tezc: Added.

connection.c
```diff
@@ -260,13 +261,15 @@ RRStatus ConnConnect(Connection *conn, const NodeAddr *addr, ConnectionCallbackF
...
     uv_req_set_data((uv_req_t *)&conn->uv_resolver, conn);
-    int r = uv_getaddrinfo(conn->rr->loop, &conn->uv_resolver, handleResolved,
+    int r = uv_getaddrinfo(conn->rr->loop, &conn->uv_resolver, NULL,
             conn->addr.host, NULL, &hints);
     if (r) {
         conn->state = CONN_CONNECT_ERROR;
         return RR_ERROR;
     }
 
+    handleResolved(&conn->uv_resolver, 0, conn->uv_resolver.addrinfo);
```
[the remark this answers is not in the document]
tezc: Added a threadpool similar to what libuv was doing.

raft.c
```diff
@@ -760,6 +756,18 @@ static void raftNotifyStateEvent(raft_server_t *raft, void *user_data, raft_stat
     RedisModule_Free(s);
 }
 
+static int raftBackpressure(raft_server_t *raft, void *user_data, raft_node_t *raft_node)
+{
+    RedisRaftCtx *rr = &redis_raft;
+    Node *node = raft_node_get_udata(raft_node);
+    if (node->pending_raft_response_num >= rr->config->max_append_req_in_flight) {
+        /* Don't send append req to this node */
```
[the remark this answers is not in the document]
tezc: There is an explanation in libraft part.

redisraft.c
```diff
@@ -809,6 +829,68 @@ static void interceptRedisCommands(RedisModuleCommandFilterCtx *filter)
...
+
+        if (!rr->config->raft_log_fsync) {
+            if (current > match) {
+                fflush(rr->log->file);
+                raft_node_set_match_idx(node, current);
+            }
+            goto out;
+        }
+
+        raft_index_t requestedIndex = fsyncRequestedIndex(&rr->fsyncThread);
```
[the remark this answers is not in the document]
tezc: Changed that a bit. Libraft is proving the index via `raft_get_index_to_sync()`. Libraft is updating the index when the node becomes the leader, so it should work correctly even some entries are deleted.

fsync.c
```diff
@@ -0,0 +1,104 @@
...
+static void* fsyncLoop(void *arg)
+{
+    int rc, fd;
+    raft_index_t request_idx;
+    FsyncThread *th = arg;
+
+    while (1) {
+        pthread_mutex_lock(&th->mtx);
+        while (th->need_fsync == 0) {
+            rc = pthread_cond_wait(&th->cond, &th->mtx);
```
yossigo, line 42: EINTR probably needs to be gracefully ignored here.

hiredis_redismodule.h
```diff
@@ -0,0 +1,97 @@
+#ifndef __HIREDIS_REDISMODULE_H__
```
yossigo, line 1: This should become a standard hiredis adapter at some point.

raft.c
```diff
@@ -2575,26 +2488,54 @@ static void handleConfig(RedisRaftCtx *rr, RaftReq *req)
...
+        if (!rr->log->fsync) {
+            raft_index_t next = raft_get_index_to_sync(rr->raft);
+            if (next > 0) {
+                fflush(rr->log->file);
+            }
+            flushed = next;
+            goto out;
+        }
+
+        raft_index_t next = raft_get_index_to_sync(rr->raft);
```
yossigo, line 2527: Everything here up to fsyncThreadAddTask() is identical, consider branching later?

redisraft.h
```diff
@@ -244,19 +257,63 @@ typedef struct RaftMeta {
...
+
+typedef struct ThreadPool {
+    int thread_count;
+    pthread_t *threads;
+
+    STAILQ_HEAD(tasks, Task) tasks;
+
+    pthread_mutex_t mtx;
+    pthread_cond_t cond;
+    int shutdown;
```
yossigo, line 275: Consider `bool`, this is used elsewhere.

redisraft.h
```diff
@@ -244,19 +257,63 @@ typedef struct RaftMeta {
...
+
+/* fsync.c */
+typedef struct FsyncThread
+{
+    pthread_t id;
+    pthread_cond_t cond;
+    pthread_mutex_t mtx;
+
+    int need_fsync;
+    int running;
```
yossigo, line 295: Also here, consider `bool`.

redisraft.h
```diff
@@ -292,22 +349,22 @@ extern RedisRaftCtx redis_raft;
...
+#define REDIS_RAFT_DEFAULT_RECONNECT_INTERVAL         100
+#define REDIS_RAFT_DEFAULT_PROXY_RESPONSE_TIMEOUT     10000
+#define REDIS_RAFT_DEFAULT_RAFT_RESPONSE_TIMEOUT      1000
+#define REDIS_RAFT_DEFAULT_LOG_MAX_CACHE_SIZE         (8*1000*1000)
+#define REDIS_RAFT_DEFAULT_LOG_MAX_FILE_SIZE          (64*1000*1000)
+#define REDIS_RAFT_HASH_SLOTS                         16384
+#define REDIS_RAFT_HASH_MIN_SLOT                      0
+#define REDIS_RAFT_HASH_MAX_SLOT                      16383
 #define REDIS_RAFT_DEFAULT_SHARDGROUP_UPDATE_INTERVAL 5000
+#define REDIS_RAFT_DEFAULT_MAX_APPENDENTRIES          1
```
yossigo, line 367: Is this really a good default?

yossigo (commented):
@tezc Overall, this look very good and really pretty compact compared to the scope of the change. A few more high level suggestions:

* Most new functions are missing top comments (what they do/expect, in/out and error handling).
* It might be a good idea to follow this PR with some restructuring of the code. The use of RaftReq everywhere, even when no long running/blocking requests are handled, is redundant. Maybe it's also time to break `raft.c` up, now that `redisraft.c` anyway invokes everything directly.
* Need to consider potential symbol collision, especially now that there lots of generic `handleSomething()` functions that are no longer static. Probably another reason to push the issue above - but really not a must as part of this PR.

fsync.c
tezc: `pthread_cond_wait()` will not return EINTR according to POSIX (https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_cond_wait.html) 


> These functions will not return an error code of [EINTR].

In our case, pthread_cond_wait() never returns an error (if it does, we have a bug).

hiredis_redismodule.h
tezc: You mean merging into hiredis? Maybe.

raft.c
tezc: Refactored that part

redisraft.h
tezc: done

redisraft.h
tezc: Increased to 4

tezc:
@yossigo Added some comments to functions. Regarding refactor, I suggest doing that after merging this PR not to make this one more complex.

fsync.c
```diff
@@ -3,6 +3,14 @@
...
 #include <errno.h>
 
+/* Add fsync task. requested_index should be the latest entry index in the file.
+ * This index will be reported on `on_complete` callback.
+ *
+ * If you call this function many times while fsyncThread is in fsync(), only
+ * the latest requested_index will be reported. This is some sort of batching.
+ * We'll call fsync() once and report the latest index that we called fsync()
+ * for.
+ * */
```
yossigo, line 13:
```suggestion
 */
```

fsync.c
yossigo: Wasn't aware of that!

fsync.c
```diff
@@ -73,6 +81,18 @@ static void* fsyncLoop(void *arg)
...
+ *
+ * `result` parameter will point to a FsyncThreadResult object:
+ *
+ * typedef struct FsyncThreadResult {
+ *    raft_index_t fsync_index;  // index parameter passed in fsyncThreadAddTask()
+ *    uint64_t time; // Time fsync() took in microseconds
+ * } FsyncThreadResult;
+ *
+ * User should free() the `result` object in on_complete() callback.
+ * */
```
yossigo, line 95:
```suggestion
 */
```

yossigo (approved):

yossigo (dismissed):

tezc:
@yossigo Sorry, need another approval. I had to squash-pull libraft.

yossigo (approved):

PR #194, opened 2021-12-11
Intercept "info" to spoof cluster_enabled and prevent starting with cluster_enabled flag

sjpotter:
As Redis Raft does the clustering itself, it can't be run the cluster_enabled yes, but clients expect to see the flag set, so have to spoof it.

### DIFF
--- a/raft.c
+++ b/raft.c
@@ -1823,13 +1823,13 @@ static bool handleMultiExec(RedisRaftCtx *rr, RaftReq *req)
 void handleInfoCommand(RedisRaftCtx *rr, RaftReq *req) {
     RaftRedisCommand *cmd = req->r.redis.cmds.commands[0];
 
-    const char * section = "all";
+    const char *section = "all";
 
     if (cmd->argc > 2) {
         /* Note: we can't use RM_WrongArity here because our req->ctx is a thread-safe context
          * with a synthetic client that no longer has the original argv.
          */
-        RedisModule_ReplyWithError(req->ctx, "ERR wrong number of arguments for 'cluster' command");
+        RedisModule_ReplyWithError(req->ctx, "ERR wrong number of arguments for 'info' command");
         goto exit;
     }
     if (cmd->argc == 2) {
@@ -1841,12 +1841,14 @@ void handleInfoCommand(RedisRaftCtx *rr, RaftReq *req) {
     const char *info = RedisModule_CallReplyProto(reply, &info_len);
 
     char * pos = strstr(info, "cluster_enabled:0");
-    pos += 16;
-    *pos = '1';
+    if (pos) {
+        pos += 16;
+        *pos = '1';
+    }
 
     RedisModule_ReplyWithStringBuffer(req->ctx, info, info_len);
 
-    exit:
+exit:
     RaftReqFree(req);
 }
 
--- a/raft.c
+++ b/raft.c
@@ -1220,6 +1220,7 @@ RRStatus RedisRaftInit(RedisModuleCtx *ctx, RedisRaftCtx *rr, RedisRaftConfig *c
     memset(rr, 0, sizeof(RedisRaftCtx));
     STAILQ_INIT(&rr->rqueue);
 
+
     /* Register an atexit handler to tell us we're exiting.  Redis offers no
      * other way and we need to be aware of this to avoid getting into execution
      * flows that involve libuv workers that self destructor using .dtors.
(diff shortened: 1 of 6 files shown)


raft.c
```diff
@@ -1819,6 +1820,36 @@ static bool handleMultiExec(RedisRaftCtx *rr, RaftReq *req)
...
+void handleInfoCommand(RedisRaftCtx *rr, RaftReq *req) {
+    RaftRedisCommand *cmd = req->r.redis.cmds.commands[0];
+
+    const char * section = "all";
+
+    if (cmd->argc > 2) {
+        /* Note: we can't use RM_WrongArity here because our req->ctx is a thread-safe context
+         * with a synthetic client that no longer has the original argv.
+         */
+        RedisModule_ReplyWithError(req->ctx, "ERR wrong number of arguments for 'cluster' command");
```
tezc, line 1832:
```suggestion
        RedisModule_ReplyWithError(req->ctx, "ERR wrong number of arguments for 'info' command");
```

raft.c
```diff
@@ -1819,6 +1820,36 @@ static bool handleMultiExec(RedisRaftCtx *rr, RaftReq *req)
     return false;
 }
 
+void handleInfoCommand(RedisRaftCtx *rr, RaftReq *req) {
+    RaftRedisCommand *cmd = req->r.redis.cmds.commands[0];
+
+    const char * section = "all";
```
tezc, line 1826:
```suggestion
    const char *section = "all";
```

raft.c
```diff
@@ -1819,6 +1820,36 @@ static bool handleMultiExec(RedisRaftCtx *rr, RaftReq *req)
...
+
+    RedisModuleCallReply *reply = RedisModule_Call(req->ctx, "INFO", "c", section);
+    size_t info_len;
+    const char *info = RedisModule_CallReplyProto(reply, &info_len);
+
+    char * pos = strstr(info, "cluster_enabled:0");
+    pos += 16;
+    *pos = '1';
+
+    RedisModule_ReplyWithStringBuffer(req->ctx, info, info_len);
```
tezc, line 1847:
```suggestion
    char *pos = strstr(info, "cluster_enabled:");
    if (pos) {
        pos += strlen("cluster_enabled:");
        *pos = rr->config->sharding ? '1' : '0';
    }

    RedisModule_ReplyWithCallReply(req->ctx, reply);
    RedisModule_FreeCallReply(reply);
```



1 - If the command is something like "info cpu", reply string will not contain "cluster_enabled".  So, we need a null check here.
2 - Should we set `cluster_enabled` according to our `sharding` config? Don't know why do we need `sharding` config though. Maybe we can remove it and run as sharding enabled by default.

raft.c
```diff
@@ -1819,6 +1820,36 @@ static bool handleMultiExec(RedisRaftCtx *rr, RaftReq *req)
...
+
+    exit:
```
tezc, line 1849:
```suggestion
exit:
```

redisraft.c
```diff
@@ -915,6 +915,16 @@ __attribute__((__unused__)) int RedisModule_OnLoad(RedisModuleCtx *ctx, RedisMod
         return REDISMODULE_ERR;
     }
 
+    /* Sanity check that not running with cluster_enabled */
+    RedisModuleServerInfoData *info = RedisModule_GetServerInfo(ctx, "cluster");
+    const char * val = RedisModule_ServerInfoGetFieldC(info, "cluster_enabled");
+    int cluster_enabled = strncmp("1", val, 1);
+    RedisModule_FreeServerInfo(ctx, info);
+    if (!cluster_enabled) {
```
tezc, line 923:
```suggestion
    int cluster_enabled = (int) RedisModule_ServerInfoGetFieldSigned(info, "cluster_enabled", NULL);
    RedisModule_FreeServerInfo(ctx, info);
    if (cluster_enabled) {
```

A nitpick, sounds like `cluster_enabled` value should be reverted.

raft.c
sjpotter: I think we are leaning to have sharding always enabled.  my thought is that sharding yes will remain as an option but will enable the "gossipy" current mechanism for transmitting sharding information between clusters, and that will default to off once we have an external mechanism to set the sharding info via an external orchestration mechanism

redisraft.c
sjpotter: well, before when dealing with as a string, it made sort of sense, but variable naming wasn't great for that, but treating it as a number now, everything is better.

raft.c
```diff
@@ -1819,6 +1820,38 @@ static bool handleMultiExec(RedisRaftCtx *rr, RaftReq *req)
...
+    }
+    if (cmd->argc == 2) {
+        section = RedisModule_StringPtrLen(cmd->argv[1], NULL);
+    }
+
+    RedisModuleCallReply *reply = RedisModule_Call(req->ctx, "INFO", "c", section);
+    size_t info_len;
+    const char *info = RedisModule_CallReplyProto(reply, &info_len);
+
+    char * pos = strstr(info, "cluster_enabled:0");
```
tezc, line 1843:
```suggestion
    char *pos = strstr(info, "cluster_enabled:0");
```

raft.c
tezc: Ok, currently should we set `cluster_enabled` according to `sharding` config? If we set it `1` unconditionally and if `sharding` is no, that will cause a problem with the clients?  e.g We won't send responses for `cluster nodes` etc.

tezc (approved):

PR #195, opened 2021-12-14
Execute Lua scripts in a Deteministic Manner

sjpotter:
Enable RedisRaft to intercept commands that are executed via Lua scripts.

This is to allow RedisRaft to execute commands sortable commands from within lua scripts and return them always in sorted order to enable the different nodes in a raft cluster to execute lua scripts deterministically.


sort.c
```diff
@@ -0,0 +1,109 @@
...
+    int entered_eval = redis_raft.entered_eval;
+    redis_raft.entered_eval = 0;
+    RedisModule_Log(redis_raft_log_ctx, REDIS_NOTICE, "Calling RM_Call for: %.*s", (int) cmd_len, cmd_str);
+    RedisModuleCallReply *reply = RedisModule_Call(redis_raft.ctx, cmd_str, "3v", &argv[1], argc-1);
+    exitRedisModuleCall();
+    redis_raft.entered_eval = entered_eval;
+
+    if (!reply) {
+        RedisModule_Log(redis_raft_log_ctx, REDIS_NOTICE, "Bad command or failed to execute: %.*s", (int) cmd_len, cmd_str);
+        RedisModule_ReplyWithError(ctx, "Bad command or failed to execute");
```
yossigo, line 90: We need to be more precise about the kind of error we're returning here, check out the follower proxy code which already handles some of that.

redisraft.h
```diff
@@ -579,6 +585,8 @@ typedef struct {
 #define CMD_SPEC_WRITE          (1<<2)      /* Command is a (potentially) write command */
 #define CMD_SPEC_UNSUPPORTED    (1<<3)      /* Command is not supported, should be rejected */
 #define CMD_SPEC_DONT_INTERCEPT (1<<4)      /* Command should not be intercepted to RAFT */
+#define CMD_SPEC_SORTABLE       (1<<5)      /* Command output should be sorted within a lua script */
```
yossigo, line 588: Suggest a more accurate name: `CMD_SPEC_SORT_REPLY`.

sort.c
```diff
@@ -0,0 +1,109 @@
...
+    if (len == 0) {
+        RedisModule_ReplyWithNullArray(ctx);
+        return;
+    }
+
+    RedisModuleDict * dict = RedisModule_CreateDict(ctx);
+    for(int i=0; i < len; i++) {
+        RedisModuleCallReply * entry = RedisModule_CallReplyArrayElement(reply, i);
+        size_t entry_len;
+        const char * entry_str = RedisModule_CallReplyStringPtr(entry, &entry_len);
```
yossigo, line 25:
```suggestion
    for (size_t i = 0; i < len; i++) {
        RedisModuleCallReply *entry = RedisModule_CallReplyArrayElement(reply, i);
        size_t entry_len;
        const char * entry_str = RedisModule_CallReplyStringPtr(entry, &entry_len);
```

Styling/spacing and type correctness.

sort.c
```diff
@@ -0,0 +1,109 @@
...
+        RedisModule_DictSetC(dict, (char *) entry_str, entry_len, NULL);
+    }
+
+    RedisModule_ReplyWithArray(ctx, len);
```
yossigo, line 29: Note that `ReplyWithArray` is a `long` while `CallReplyLength` is a `size_t`, perhaps we need to assert to avoid an overflow on 32-bit platforms.

sort.c
```diff
@@ -0,0 +1,109 @@
...
+    size_t len = RedisModule_CallReplyLength(reply);
+    RedisModuleDict * dict = RedisModule_CreateDict(ctx);
+
+    for(int i=0; i < len; i++) {
+        RedisModuleCallReply * key;
+        RedisModuleCallReply * value;
+
+        RedisModule_CallReplyMapElement(reply, i, &key, &value);
+        size_t key_len;
+        const char * key_str = RedisModule_CallReplyStringPtr(key, &key_len);
```
yossigo, line 53:
```suggestion
    RedisModuleDict *dict = RedisModule_CreateDict(ctx);

    for (size_t i = 0; i < len; i++) {
        RedisModuleCallReply *key;
        RedisModuleCallReply *value;

        RedisModule_CallReplyMapElement(reply, i, &key, &value);
        size_t key_len;
        const char *key_str = RedisModule_CallReplyStringPtr(key, &key_len);
```

Styling / spacing / type correctness.

sort.c
```diff
@@ -0,0 +1,109 @@
...
+
+void handleSort(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
+{
+    if (!sortableCommand(argv[0])) {
+        RedisModule_ReplyWithError(ctx, "ERR: not a sortable command");
+        return;
+    }
+
+    size_t cmd_len;
+    const char * cmd_str = RedisModule_StringPtrLen(argv[0], &cmd_len);
```
yossigo, line 78:
```suggestion
    const char *cmd_str = RedisModule_StringPtrLen(argv[0], &cmd_len);
```

tests/integration/test_random.py
```diff
@@ -0,0 +1,53 @@
...
+    create hash fields and set their values depending on the order of
+    keys.
+
+    Hash keys that exceed the 'hash-max-*' configuration settings
+    will be created as unordered hash tables and we want to confirm they
+    maintain the same order across the cluster.
+    """
+
+    cluster.create(3)
+    #cluster.config_set('lua-replicate-commands', 'no')
```
yossigo, line 24:
```suggestion
```

yossigo (commented):
@sjpotter Have several comments in different levels of importance. Most important points are:

* We need a non-unique array sort option, at least for `HVALS` (maybe others).
* Need much better test coverage.

sjpotter:
re your high level points

1) would changing the array dict to count, as well work for now?  the advantage of doing it that way, would be that for now wouldn't have to distinguish at all between array output types.  would slow it down in practical terms from the existing usage but would be same Big-O complexity.

2) agree in general.

sjpotter:
re your high level points

1) would changing the array dict to count, as well work for now?  the advantage of doing it that way, would be that for now wouldn't have to distinguish at all between array output types.  would slow it down in practical terms from the existing usage but would be same Big-O complexity.

I have a prototype that does, just that.

2) agree in general.

yossigo:
@sjpotter I think [1] makes sense, it also follows our current decision for a not-ideal-but-working approach.

sjpotter:
re #2.  would tests that directly test raft.sort be ok, or would you want them in lua context?  (the former would be easier for me to write)

yossigo:
@sjpotter I think Lua context would be better. Shouldn't be that more difficult to create I think.

sort.c
sjpotter: is this effectively a bug in redis module interface?

tests/integration/test_random.py
```diff
@@ -0,0 +1,53 @@
...
+
+Copyright (c) 2021 Redis Ltd.
+
+RedisRaft is licensed under the Redis Source Available License (RSAL).
+"""
+
+import time
+import logging
+
+def test_hash_deterministic_order(cluster):
```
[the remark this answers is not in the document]
sjpotter: added more, but a little difficult to do with resp3.  I'd also argue that if there was any difference, it be a bug in redis, I run the command (whatever it is) as resp3 always, and depend on redis to translate it out correctly.

sjpotter:
> @sjpotter I think Lua context would be better. Shouldn't be that more difficult to create I think.

will try to figure that out tomorrow.  probably useful to test raft._sort_reply directly as well.

sort.c
sjpotter: looking at the proxy code, this is what it does in the case of !reply

```
    if (!reply) {
        /* Connection have dropped.  The state of the request is unknown at this point
         * and this must be reflected to the user.
         *
         * Ideally the connection should be dropped but Module API does not provide for that.
         */
        ConnMarkDisconnected(req->r.redis.proxy_node->conn);
        RedisModule_ReplyWithError(req->ctx, "TIMEOUT no reply from leader");
        redis_raft.proxy_failed_responses++;
        goto exit;
    }
```

that string obviously isn't what we need, but it seems we are just returning a string, so I guess, a little confused, what would be a better string

tests/integration/test_random.py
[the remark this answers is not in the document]
sjpotter: still nothing with resp3, but test all the commands now both inside and outside of lua

redisraft.h
```diff
@@ -532,6 +534,10 @@ typedef struct RaftReq {
             raft_node_id_t id;
         } node_shutdown;
         raft_node_id_t node_to_transfer_leader;
+        struct {
+            RedisModuleString **argv;
+            int argc;
+        } sort_command;
```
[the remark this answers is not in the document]
sjpotter: agreed, no need anymore.  unsure why I missed this comment earlier.  artifact of first attempt

redisraft.h
sjpotter: ok.  I thought sortable was fine, as its a command whose output we can sort without breaking its semantics.  but I changed it to yours

sort.c
```diff
@@ -10,30 +10,52 @@
 
 #include "redisraft.h"
 
-static void replySortedArray(RedisModuleCtx *ctx, RedisModuleCallReply * reply)
+static void replySortedArray(RedisModuleCtx *ctx, RedisModuleCallReply * reply, int reply_type)
```
yossigo, line 13: Missing function documentation.
Also: Why pass reply_type explicitly, when it can be inferred from the RedisModuleCallReply?

sort.c
```diff
@@ -10,30 +10,52 @@
...
-    if (len == 0) {
-        RedisModule_ReplyWithNullArray(ctx);
-        return;
-    }
 
-    RedisModuleDict * dict = RedisModule_CreateDict(ctx);
-    for(int i=0; i < len; i++) {
-        RedisModuleCallReply * entry = RedisModule_CallReplyArrayElement(reply, i);
+    RedisModuleDict *dict = RedisModule_CreateDict(ctx);
+    for(size_t i = 0; i < len; i++) {
```
yossigo, line 18:
```suggestion
    for (size_t i = 0; i < len; i++) {
```

sort.c
```diff
@@ -10,30 +10,52 @@
...
+        long *val;
+        RedisModuleCallReply * entry;
```
yossigo, line 20:
```suggestion
        RedisModuleCallReply *entry;
```

sort.c
```diff
@@ -10,30 +10,52 @@
...
+        switch (reply_type) {
+            case REDISMODULE_REPLY_ARRAY:
+                entry = RedisModule_CallReplyArrayElement(reply, i);
+                break;
+            case REDISMODULE_REPLY_SET:
+                entry = RedisModule_CallReplySetElement(reply, i);
+                break;
+        }
```
yossigo, line 28: Consider being more defensive in case an unexpected type is specified (or passed).

sort.c
```diff
@@ -10,30 +10,52 @@
...
+
         size_t entry_len;
         const char * entry_str = RedisModule_CallReplyStringPtr(entry, &entry_len);
-        RedisModule_DictSetC(dict, (char *) entry_str, entry_len, NULL);
+        val = RedisModule_DictGetC(dict, (char *) entry_str, entry_len, NULL);
+        if (!val) {
```
yossigo, line 33: How about just casting the `void *` to `unsigned long` to keep this counter? A bit hackish but will be much more efficient.

sort.c
```diff
@@ -10,30 +10,52 @@
...
 
-    RedisModule_ReplyWithArray(ctx, len);
+    switch (reply_type)
+    {
+        case REDISMODULE_REPLY_ARRAY:
+            RedisModule_ReplyWithArray(ctx, len);
+            break;
+        case REDISMODULE_REPLY_SET:
+            RedisModule_ReplyWithSet(ctx, len);
+            break;
```
yossigo, line 47: Should probably pass the return value from RedisModule_ReplyWith...

sort.c
```diff
@@ -10,30 +10,52 @@
...
 
     char *key;
     size_t key_len;
-    void *data;
+    long *val;
     RedisModuleDictIter *iter = RedisModule_DictIteratorStartC(dict, "^", NULL, 0);
-    while((key = RedisModule_DictNextC(iter, &key_len, &data)) != NULL) {
-        RedisModule_ReplyWithStringBuffer(ctx, key, key_len);
+    while((key = RedisModule_DictNextC(iter, &key_len, (void **) &val)) != NULL) {
+        for(long i=0; i < *val; i++) {
```
yossigo, line 55:
```suggestion
        for (long i=0; i < *val; i++) {
```

sort.c
```diff
@@ -69,13 +91,15 @@ static void replySortedMap(RedisModuleCtx *ctx, RedisModuleCallReply * reply)
 
 void handleSort(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
 {
-    if (!sortableCommand(argv[0])) {
+    RedisModule_Log(redis_raft_log_ctx, REDIS_NOTICE, "in handle sort");
```
yossigo, line 94: Remove me.

tests/integration/test_random.py
```diff
@@ -21,33 +20,367 @@ def test_hash_deterministic_order(cluster):
...
 
     cluster.create(3)
-    #cluster.config_set('lua-replicate-commands', 'no')
 
     script = cluster.node(1).client.register_script("""
 -- Populate hash KEYS[1] with ARGV[1] fields, then assign each field
--- a value corresponding to its index in HKEYS
-local key = KEYS[1]
+-- a value corresponding to its index in result of cmd
+local key1 = "test"
```
yossigo, line 27: Well behaved Lua scripts need to only access keys listed in the KEYS array ("announced" keys).

tests/integration/test_random.py
```diff
@@ -5,9 +5,8 @@
...
 RedisRaft is licensed under the Redis Source Available License (RSAL).
 """
-
-import time
-import logging
+import random
+import string
 
 def test_hash_deterministic_order(cluster):
     """
```
yossigo, line 12: Update test description?

redisraft.c
```diff
@@ -830,15 +830,15 @@ static void interceptRedisCommands(RedisModuleCommandFilterCtx *filter)
...
-                size_t len;
-                const char *str = RedisModule_StringPtrLen(RedisModule_CommandFilterArgGet(filter, 0), &len);
-                RedisModule_Log(redis_raft_log_ctx, REDIS_NOTICE, "should sort: %.*s", (int) len, str);
-
-                RedisModuleString *raft_str = RedisModule_CreateString(NULL, "RAFT.SORT", 9);
+            const CommandSpec *cs = CommandSpecGet(cmd);
+            if (!cs) {
+                ;
+            } else if (cs->flags & CMD_SPEC_SORT_REPLY) {
+                RedisModule_Log(redis_raft_log_ctx, REDIS_NOTICE, "inserting RAFT._SORT_REPLY");
```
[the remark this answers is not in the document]
sjpotter: no, this was purposeful.  there are no returns in that if /else if / else if block, the return is outside of them. I just wanted a clear case where cs wasn't returned.  just my taste in code.  perhaps you'd prefer it with a comment there?

redisraft.c
```diff
@@ -830,15 +830,15 @@ static void interceptRedisCommands(RedisModuleCommandFilterCtx *filter)
...
             RedisModuleString *cmd = RedisModule_CommandFilterArgGet(filter, 0);
-            if (sortableCommand(cmd)) {
-                size_t len;
-                const char *str = RedisModule_StringPtrLen(RedisModule_CommandFilterArgGet(filter, 0), &len);
-                RedisModule_Log(redis_raft_log_ctx, REDIS_NOTICE, "should sort: %.*s", (int) len, str);
-
-                RedisModuleString *raft_str = RedisModule_CreateString(NULL, "RAFT.SORT", 9);
+            const CommandSpec *cs = CommandSpecGet(cmd);
+            if (!cs) {
+                ;
```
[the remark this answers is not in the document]
sjpotter: no, this was purposeful.  there are no returns in that if /else if / else if block, the return is outside of them. I just wanted a clear case where cs wasn't returned.  just my taste in code.  perhaps you'd prefer it with a comment there?

redisraft.c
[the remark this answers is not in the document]
sjpotter: ok, logging messages such as this can be removed

sort.c
sjpotter: ok re documenation

good Q. I was going to say I didn't know it could be inferred, but that's obviously wrong, as I infer it to get that value.  so only reason is simply to not infer it again, but probably not a good enough reason.

sort.c
sjpotter: you mean check?

tests/integration/test_random.py
sjpotter: ok.  easy enough to fix.  will do that.

tests/integration/test_random.py
sjpotter: ok will do.

sort.c
sjpotter: I dont understand this.

yossigo: Instead of allocating a `long`, storing a pointer to it in the dict and incrementing the value the pointer points to - do something like:

```c
unsigned long myval = (unsigned long) RM_DictGetC(...);
myval++;
RM_DictReplaceC(..., (void *) myval);
```

sort.c
yossigo: I think `replySortedArray` should bail out if a `ReplyWith...` returns an error, and itself return an error for callers to be able to handle it.

sort.c
```diff
@@ -45,34 +54,40 @@ static void replySortedArray(RedisModuleCtx *ctx, RedisModuleCallReply * reply,
         case REDISMODULE_REPLY_SET:
             RedisModule_ReplyWithSet(ctx, len);
             break;
+        default:
+            /* shouldn't get here, as should be protected above */
+            assert(0);
```
yossigo, line 59: Better use `RedisModule_Assert` (and drop `assert.h`).

sjpotter: 1) oh, I see.  trick of storing the value in the pointer.  also, that I was using Set wrong, it only worked as an artifact of the fact that I was getting the pointer and incrementing it.  with that said, that is an optimization in a way (the only Q would be, which is less efficient, walking the tree an extra time per key, or allocating memory. I guess allocating memory.

2) ok, I think I understand that

3).Easy enough change

sort.c
sjpotter: so looking at the docs, it says

```
RedisModule_ReplyWithArray

int RedisModule_ReplyWithArray(RedisModuleCtx *ctx, long len);

Reply with an array type of 'len' elements. However 'len' other calls to ReplyWith* style functions must follow in order to emit the elements of the array.

When producing arrays with a number of element that is not known beforehand the function can be called with the special count REDISMODULE_POSTPONED_LEN, and the actual number of elements can be later set with RedisModule_ReplySetArrayLength() (which will set the latest "open" count if there are multiple ones).

The function always returns REDISMODULE_OK.
```

is the point to be protective for the future where it could fail?

redisraft.c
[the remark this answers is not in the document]
yossigo: That would help.

sort.c
```diff
@@ -0,0 +1,145 @@
...
+        RedisModule_ReplyWithCallReply(ctx, value);
+    }
+}
+
+/* Calls Redis commands whose results can be sorted without semantically breaking them */
+void handleSort(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
+{
+    const CommandSpec *cs = CommandSpecGet(argv[0]);
+    if (!cs || !(cs->flags & CMD_SPEC_SORT_REPLY)) {
+        RedisModule_ReplyWithError(ctx, "ERR: not a sortable command");
```
yossigo, line 112:
```suggestion
        RedisModule_ReplyWithError(ctx, "ERR not a sortable command");
```

sort.c
```diff
@@ -0,0 +1,145 @@
...
+        unsigned long val = (unsigned long) RedisModule_DictGetC(dict, (char *) entry_str, entry_len, NULL);
+        val++;
+        RedisModule_DictReplaceC(dict, (char *) entry_str, entry_len, (void *) val);
+    }
+
+    switch (reply_type)
+    {
+        case REDISMODULE_REPLY_ARRAY:
+            if (RedisModule_ReplyWithArray(ctx, len) != REDISMODULE_OK) {
+                RedisModule_ReplyWithError(ctx, "Failed to generate sorted reply");
```
yossigo, line 48:
```suggestion
                RedisModule_ReplyWithError(ctx, "ERR Failed to generate sorted reply");
```

sort.c
```diff
@@ -0,0 +1,145 @@
...
+                goto early_exit;
+            }
+            break;
+        case REDISMODULE_REPLY_SET:
+            if (RedisModule_ReplyWithSet(ctx, len) != REDISMODULE_OK) {
+                RedisModule_ReplyWithError(ctx, "Failed to generate sorted reply");
```
yossigo, line 54:
```suggestion
                RedisModule_ReplyWithError(ctx, "ERR Failed to generate sorted reply");
```

sort.c
```diff
@@ -0,0 +1,145 @@
...
+    enterRedisModuleCall();
+    int entered_eval = redis_raft.entered_eval;
+    redis_raft.entered_eval = 0;
+    RedisModuleCallReply *reply = RedisModule_Call(redis_raft.ctx, cmd_str, "3v", &argv[1], argc-1);
+    exitRedisModuleCall();
+    redis_raft.entered_eval = entered_eval;
+
+    if (!reply) {
+        RedisModule_Log(redis_raft_log_ctx, REDIS_NOTICE, "Bad command or failed to execute: %.*s", (int) cmd_len, cmd_str);
+        RedisModule_ReplyWithError(ctx, "Bad command or failed to execute");
```
yossigo, line 128:
```suggestion
        RedisModule_ReplyWithError(ctx, "ERR Bad command or failed to execute");
```

sort.c
```diff
@@ -0,0 +1,145 @@
...
+             * have to add the proper ReplyWith function below
+             */
+            case REDISMODULE_REPLY_ARRAY:
+                entry = RedisModule_CallReplyArrayElement(reply, i);
+                break;
+            case REDISMODULE_REPLY_SET:
+                entry = RedisModule_CallReplySetElement(reply, i);
+                break;
+            default:
+                RedisModule_ReplyWithError(ctx, "Unknown type to sort");
```
yossigo, line 33:
```suggestion
                RedisModule_ReplyWithError(ctx, "ERR Unknown type to sort");
```

sort.c
yossigo: @sjpotter My bad, I was referring to `executeRaftRedisCommandArray()` where we handle `RM_Call` failure and try to return an error that is compatible with Redis.

redisraft.c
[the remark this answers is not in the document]
sjpotter: redid, to be cleaner

sort.c
sjpotter: ok, refactored out what the code did there into a function that is called in both places.

redisraft.c
```diff
@@ -830,16 +830,16 @@ static void interceptRedisCommands(RedisModuleCommandFilterCtx *filter)
...
-                RedisModule_Log(redis_raft_log_ctx, REDIS_NOTICE, "inserting RAFT._SORT_REPLY");
-                RedisModuleString *raft_str = RedisModule_CreateString(NULL, "RAFT._SORT_REPLY", 16);
-                RedisModule_CommandFilterArgInsert(filter, 0, raft_str);
-            } else if (cs->flags & CMD_SPEC_RANDOM) {
-                RedisModuleString *raft_str = RedisModule_CreateString(NULL, "RAFT._REJECT_RANDOM_COMMAND", 27);
-                RedisModule_CommandFilterArgInsert(filter, 0, raft_str);
+            const CommandSpec *cs;
+            if ((cs = CommandSpecGet(cmd)) != NULL) {
+                if (cs->flags & CMD_SPEC_SORT_REPLY) {
+                    RedisModule_Log(redis_raft_log_ctx, REDIS_NOTICE, "inserting RAFT._SORT_REPLY");
```
yossigo: Remove me?

yossigo (approved):
@sjpotter Pushed a minor commit (logging leftover), LGTM to me. Please don't forget to update the PR title and top comment to properly describe it and use it as part of the squash commit.

yossigo (approved):

PR #198, opened 2021-12-27
fail join and set state to unitialized if run out of nodes to join

### LINKED ISSUE #187: Join before init bricks joining redis-server
the "joining" node gets stuck in joining state

➜  ~ redis-cli -p 5002
127.0.0.1:5002> raft.cluster join 172.17.0.2:6379
(error) ERR failed to join cluster, please check logs
127.0.0.1:5002> 
➜  ~ redis-cli -p 5001
127.0.0.1:5001> raft.cluster init
OK 33530781e51b4003f9ab411695defc10
127.0.0.1:5001> 
➜  ~ redis-cli -p 5002
127.0.0.1:5002> raft.info
**<redacted>...state:joining...<redacted>** (and this will never change)
127.0.0.1:5002> raft.cluster join 172.17.0.2:6379
(error) ERR Already cluster member

sjpotter:
Before, once we start a join, the state stays in joining and keeps on retrying nodes that have failed

This means, that one can't issue another join, as the state is no longer uninitialized

this changes both things.

1) once we iterate through all given/discovered nodes, fail back to user with error message
2) when we fail back to user, reset state to unitialized

fixes #187

### DIFF
--- a/common.c
+++ b/common.c
@@ -246,6 +246,7 @@ void joinLinkIdleCallback(Connection *conn)
 
 exit_fail:
     ConnAsyncTerminate(conn);
+    rr->state = REDIS_RAFT_UNINITIALIZED;
 
     snprintf(err_msg, sizeof(err_msg), "ERR failed to %s cluster, please check logs", state->type);
     RedisModule_ReplyWithError(state->req->ctx, err_msg);
--- a/common.c
+++ b/common.c
@@ -233,7 +233,13 @@ void joinLinkIdleCallback(Connection *conn)
         state->addr_iter = state->addr_iter->next;
     }
     if (!state->addr_iter) {
-        state->addr_iter = state->addr;
+        if (!state->started) {
+            state->started = true;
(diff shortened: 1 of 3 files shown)


yossigo (commented):
@sjpotter LGTM, but I think we may want to have a test for that?

sjpotter:
I'm unhappy with it at the moment.  We need to discuss the semantics of join.  Currently, it basically tries forever (which is part of the problem faced, can't override the bad join given up front).  But sometimes that might be necessary (ex: cluster doesn't have quorum, so no leader).  Perhaps its fine to fail in those cases, i.e. we will redirect to leader, but once we fail, we fail and its up to the user to try again.

sjpotter:
small tweaks + tests added/changed

ushachar:
@sjpotter Can we land this? It's hurting us when doing large scale orchestration...

tezc (approved):

PR #205, opened 2021-12-30
Add RAFT.SHARDGROUP REPLACE command

sjpotter:
this replaces the whole shardgroup listing with the data contained within it.

it is similiar to RAFT.SHARDGROUP ADD, except it takes multiple shardgroups sets of data and has an initial count

RAFT.SHARDGROUP REPLACE [num shardgroups] ([shardgroup id] [num_slots] [num_nodes] ([start slot] [end slot] [slot type])* ([node-uid node-addr:node-port])*)*


cluster.c
```diff
@@ -379,6 +379,58 @@ RRStatus ShardGroupAppendLogEntry(RedisRaftCtx *rr, ShardGroup *sg, int type, vo
     return RR_OK;
 }
 
+RRStatus ShardGroupsAppendLogEntry(RedisRaftCtx *rr, int num_sg, ShardGroup **sg, int type, void *user_data)
+{
+    /* Make sure we're still a leader, could have changed... */
+    /* SJP: unsure this is true in this case */
+    if (!raft_is_leader(rr->raft)) {
```
tezc, line 386: There is a checkLeader() call before calling this function. So, I believe you can remove this check here.

cluster.c
```diff
@@ -379,6 +379,58 @@ RRStatus ShardGroupAppendLogEntry(RedisRaftCtx *rr, ShardGroup *sg, int type, vo
...
+        return RR_ERROR;
+    }
+
+    /* Serialize */
+    /* We reuse the existing shard group serialization
+     * Format:
+     * <num shard groups>:strlen(sg #1 payload):sg #1 pauload|...:strlen(sg #n payload):sg #n payload|
```
tezc, line 393:
```suggestion
     * <num shard groups>:strlen(sg #1 payload):sg #1 payload|...:strlen(sg #n payload):sg #n payload|
```

cluster.c
```diff
@@ -379,6 +379,58 @@ RRStatus ShardGroupAppendLogEntry(RedisRaftCtx *rr, ShardGroup *sg, int type, vo
...
+
+    size_t len;
+    const char * payload = RedisModule_StringPtrLen(str, &len);
+
+    raft_entry_t *entry = raft_entry_new(len);
+    entry->type = type;
+    entry->id = rand();
+    entry->user_data = user_data;
+    memcpy(entry->data, payload, len);
+    RedisModule_Free(str);
```
tezc, line 419: Would this make code a bit simpler ? 

```suggestion
    size_t buflen = 4096;
    char *buf = RedisModule_Alloc(buflen);

    buf = catsnprintf(buf, &buflen, "%d:", num_sg);
    for (int i = 0; i < num_sg; i++) {
        char *payload = ShardGroupSerialize(sg[i]);
        buf = catsnprintf(buf, &buflen, "%lu:%s:", strlen(payload), payload);
        RedisModule_Free(payload);
    }
    size_t payload_len = strlen(buf);
    raft_entry_t *entry = raft_entry_new(payload_len);
    entry->type = type;
    entry->id = rand();
    entry->user_data = user_data;
    memcpy(entry->data, buf, payload_len);
    RedisModule_Free(buf);
```

cluster.c
```diff
@@ -379,6 +379,58 @@ RRStatus ShardGroupAppendLogEntry(RedisRaftCtx *rr, ShardGroup *sg, int type, vo
...
+    }
+
+    size_t len;
+    const char * payload = RedisModule_StringPtrLen(str, &len);
+
+    raft_entry_t *entry = raft_entry_new(len);
+    entry->type = type;
+    entry->id = rand();
+    entry->user_data = user_data;
+    memcpy(entry->data, payload, len);
```
tezc, line 418: Are we copying here without NULL character? Should we copy it as well if there is no downside?

cluster.c
```diff
@@ -823,15 +879,17 @@ RRStatus ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
 
     id = RedisModule_StringPtrLen(argv[0], &len);
     if (len > RAFT_DBID_LEN) {
-        RedisModule_ReplyWithError(ctx, "ERR: shardgrup id is too long");
+        RedisModule_ReplyWithError(ctx, "ERR: invalid shard group message (shard group id length)");
```
tezc, line 882: My understanding is some clients strip ERR part before passing the error message. In this case, error message will start with ":". Maybe we shouldn't do this? 

```suggestion
        RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (shard group id length)");
```

cluster.c
```diff
@@ -868,7 +926,7 @@ RRStatus ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
         const char *str = RedisModule_StringPtrLen(argv[argidx++], &len);
 
         if (len != RAFT_SHARDGROUP_NODEID_LEN) {
-            RedisModule_ReplyWithError(ctx, "ERR invalid node id length");
+            RedisModule_ReplyWithError(ctx, "ERR: invalid shard group message (node id length)");
```
tezc, line 929:
```suggestion
            RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (node id length)");
```

cluster.c
```diff
@@ -877,7 +935,7 @@ RRStatus ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
 
         str = RedisModule_StringPtrLen(argv[argidx++], &len);
         if (!NodeAddrParse(str, len, &sg->nodes[i].addr)) {
-            RedisModule_ReplyWithError(ctx, "ERR invalid node address/port");
+            RedisModule_ReplyWithError(ctx, "ERR: invalid shard group message (node address/port)");
```
tezc, line 938:
```suggestion
            RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (node address/port)");
```

raft.c
```diff
@@ -2354,6 +2369,94 @@ void applyShardGroupChange(RedisRaftCtx *rr, raft_entry_t *entry)
     }
 }
 
+void replaceShardGroups(RedisRaftCtx *rr, raft_entry_t *entry)
+{
+    // 1. reset sharding info
+    ShardingInfo * si = rr->sharding_info;
```
tezc, line 2375:
```suggestion
    ShardingInfo *si = rr->sharding_info;
```

raft.c
```diff
@@ -2354,6 +2369,94 @@ void applyShardGroupChange(RedisRaftCtx *rr, raft_entry_t *entry)
...
+
+    ShardGroup *local = NULL;
+    if (si->shard_group_map != NULL) {
+        RedisModuleDictIter *iter = RedisModule_DictIteratorStartC(si->shard_group_map, "^", NULL, 0);
+
+        char * key;
```
tezc, line 2381:
```suggestion
        char *key;
```

raft.c
```diff
@@ -2354,6 +2369,94 @@ void applyShardGroupChange(RedisRaftCtx *rr, raft_entry_t *entry)
...
+                RedisModule_Free(data);
+            } else {
+                local = data;
+            }
+        }
+        RedisModule_DictIteratorStop(iter);
+        RedisModule_FreeDict(rr->ctx, si->shard_group_map);
+        si->shard_group_map = NULL;
+
+        RedisModule_Free(si->shard_group_map);
```
tezc, line 2397:
```suggestion
```

tezc: This has no effect as we freed it just above

raft.c
```diff
@@ -2354,6 +2369,94 @@ void applyShardGroupChange(RedisRaftCtx *rr, raft_entry_t *entry)
...
+        payload = pPayload + 1;
+
+        ShardGroup sg;
+        if (ShardGroupDeserialize(payload, payload_len, &sg) != RR_OK) {
+            LOG_ERROR("Failed to deserialize shardgroup payload: [%.*s]", (int) payload_len, payload);
+            return;
+        }
+
+        if (!memcmp(rr->snapshot_info.dbid, sg.id, sizeof(sg.id))) {
+            /* 2a. if payload is for local shardgroup, skip */
```
tezc, line 2440: sg will leak here if I'm not mistaken

raft.c
```diff
@@ -2354,6 +2369,94 @@ void applyShardGroupChange(RedisRaftCtx *rr, raft_entry_t *entry)
...
+        } else {
+            /* 2b. if payload is for remote shardgroup, deserialize and add with ShardingInfoAddShardGroup() */
+            RedisModule_Assert(ShardingInfoAddShardGroup(rr, &sg) == RR_OK);
```
tezc, line 2443: sg will leak here as well?

raft.c
```diff
@@ -2354,6 +2369,94 @@ void applyShardGroupChange(RedisRaftCtx *rr, raft_entry_t *entry)
...
+
+    /* 2. iterate over payloads
+     * payload structure
+     * "# shard groups:payload1 len:payload1:....:payload n len:payload n:"
+     */
+    char * payload = entry->data;
+    char * pPayload = strchr(payload, ':');
+    *pPayload = 0;
+
+    char * endptr;
```
tezc, line 2420:
```suggestion
    char *endptr;
```

raft.c
```diff
@@ -2354,6 +2369,94 @@ void applyShardGroupChange(RedisRaftCtx *rr, raft_entry_t *entry)
...
+        if (si->hash_slots_map[i] != local) {
+            si->hash_slots_map[i] = NULL;
+        }
+    }
+
+    /* 2. iterate over payloads
+     * payload structure
+     * "# shard groups:payload1 len:payload1:....:payload n len:payload n:"
+     */
+    char * payload = entry->data;
```
tezc, line 2416:
```suggestion
    char *payload = entry->data;
```

raft.c
```diff
@@ -2354,6 +2369,94 @@ void applyShardGroupChange(RedisRaftCtx *rr, raft_entry_t *entry)
...
+    char * pPayload = strchr(payload, ':');
```
tezc, line 2417:
```suggestion
    char *pPayload = strchr(payload, ':');
```

cluster.c
```diff
@@ -889,6 +947,59 @@ RRStatus ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
...
+    argc--;
+
+    int i;
+    for(i = 0; i < req->r.shardgroups_replace.len; i++) {
+        ShardGroup * sg = RedisModule_Alloc(sizeof(ShardGroup));
+        req->r.shardgroups_replace.shardgroups[i] = sg;
+
+        long long num_slots, num_nodes;
+        if (RedisModule_StringToLongLong(argv[1], &num_slots) != REDISMODULE_OK ) {
+            RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (num slots)");
```
tezc, line 969:
```suggestion
            RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (num slots)");
            goto fail;
```

cluster.c
```diff
@@ -889,6 +947,59 @@ RRStatus ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
...
+        }
```
tezc, line 970:
```suggestion
            RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (num slots)");
            goto fail;
        }
```

raft.c
```diff
@@ -2391,28 +2494,14 @@ void handleShardGroupUpdate(RedisRaftCtx *rr, RaftReq *req)
         goto exit;
     }
 
-    /* TODO: validate/handle slot range changes:
-     *  i.e. preventing stable overlap switching to migrating or adding as importing
-     */
-    ShardGroup *sg = getShardGroupById(rr, req->r.shardgroup_add.id);
-    if (sg == NULL) {
+    if (ShardGroupsAppendLogEntry(rr, req->r.shardgroups_replace.len, req->r.shardgroups_replace.shardgroups,
```
tezc, line 2497: In line 2493, `RedisModule_ReplyWithError(req->ctx, "failed, please check logs.");` is unnecessary as we already set response in  `checkLeader()` or `checkRaftState()`. Maybe we use this oppurtunity and remove that line as part of this PR.

cluster.c
```diff
@@ -823,15 +879,17 @@ RRStatus ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
...
         goto error;
     }
     memcpy(sg->id, id, len);
     sg->id[RAFT_DBID_LEN] = '\0';
 
-    if (RedisModule_StringToLongLong(argv[1], &num_slots) != REDISMODULE_OK ||
-            RedisModule_StringToLongLong(argv[2], &num_nodes) != REDISMODULE_OK) {
-        RedisModule_ReplyWithError(ctx, "ERR invalid shard group message");
+    if (RedisModule_StringToLongLong(argv[1], &num_slots) != REDISMODULE_OK) {
+        RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (num slots)");
```
tezc, line 889:
```suggestion
        RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (num slots)");
        goto error;
```

cluster.c
```diff
@@ -889,6 +947,59 @@ RRStatus ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
...
+        return RR_ERROR;
+    }
+    req->r.shardgroups_replace.shardgroups =
+            RedisModule_Calloc(req->r.shardgroups_replace.len, sizeof(ShardGroup *));
+    argv++;
+    argc--;
+
+    int i;
+    for(i = 0; i < req->r.shardgroups_replace.len; i++) {
+        ShardGroup * sg = RedisModule_Alloc(sizeof(ShardGroup));
```
tezc, line 964:
```suggestion
        ShardGroup *sg = RedisModule_Alloc(sizeof(ShardGroup));
```

cluster.c
sjpotter: so I realized that when writing, but my concern was that perhaps this function (much like add) might be called in the future from within rediraft itself and it miss the check.  If we dont want to worry about that now, easy enough to remove.

cluster.c
sjpotter: well, this is an artifact before of using RedisModule Strings which aren't guaranteed to be null terminated, with your suggestion above, that changes.

cluster.c
sjpotter: yes, agreed, have had to change that before, keep forgetting

cluster.c
sjpotter: thinking about it, I'm unsure in practice it gains us anything even with the change above, but am open to being convinced I'm wrong

raft.c
sjpotter: well, its an assert it wont really leak, right?  this is a crasher.

sjpotter: fix above should take care of main case, unsure this case matters.

cluster.c
sjpotter: similiar error above, also fixed.

raft.c
sjpotter: not 100% sure I understand the comment.  I see we dont need the reply with error in the `checkLeader()` / `CheckRaftState()` conditional failure, but I still think its needed here.

unless that's what you mean.

tezc: Yeah, I meant that. I see you already removed that line.

raft.c
tezc: Yeah, seems resolved

cluster.c
tezc: Even then, probably this check is not necessary. There is `raft_recv_entry` call inside this function, it will/should fail anyway if this node is not the leader.

cluster.c
```diff
@@ -889,6 +951,60 @@ RRStatus ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
...
+
+        argv += num_argv_entries;
+        argc -= num_argv_entries;
+    }
+
+    return RR_OK;
+
+fail:
+    for (; i >= 0; i--) {
+        RedisModule_Free(req->r.shardgroups_replace.shardgroups[i]);
```
tezc, line 999: I think we should also call ShardGroupFree() here. 

This issue aside, I think `ShardGroupFree()` creates a confusion because of its naming (It actually doesn't free ShardGroup itself).

Maybe we create these functions : 
```c
void ShardGroupInit(ShardGroup *sg) {
    memset(sg, 0, sizeof(ShardGroup));
}

void ShardGroupTerm(ShardGroup *sg)
{
    if (sg->conn) {
        ConnAsyncTerminate(sg->conn);
        sg->conn = NULL;
    }

    if (sg->slot_ranges) {
        RedisModule_Free(sg->slot_ranges);
        sg->slot_ranges = NULL;
    }

    if (sg->nodes) {
        RedisModule_Free(sg->nodes);
        sg->nodes = NULL;
    }
}

ShardGroup *ShardGroupCreate() {
    ShardGroup *sg = RedisModule_Alloc(sizeof(*sg));
    ShardGroupInit(sg);
    return sg;
}

void ShardGroupFree(ShardGroup *sg) {
    ShardGroupTerm(sg);
    RedisModule_Free(sg);
}
```

Now, we can pair `ShardGroupCreate()` with `ShardGroupFree()` and `ShardGroupInit()` with `ShardGroupTerm()`. We don't call manually RedisModule_Calloc() to create ShardGroup etc. All goes through these functions. We just make sure we pair these correctly. Does that make sense?

cluster.c
```diff
@@ -759,20 +814,24 @@ RRStatus ShardingInfoAddShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
...
-    if (RedisModule_DictSetC(si->shard_group_map, new_sg->id, strlen(new_sg->id), sg) != REDISMODULE_OK) {
+    if (ShardingInfoValidateShardGroup(rr, new_sg) != RR_OK) {
         return RR_ERROR;
     }
-    if (strlen(new_sg->id) >= sizeof(sg->id)) {
+
+    if (strlen(new_sg->id) >= sizeof(new_sg->id)) {
         return RR_ERROR;
     }
+    ShardGroup *sg = RedisModule_Calloc(1, sizeof(ShardGroup));
```
tezc, line 824: Similar to my previous comment, we need to call free this allocation if we goto `error` section?

raft.c
```diff
@@ -1349,6 +1354,14 @@ void RaftReqFree(RaftReq *req)
                 req->r.shardgroup_add.nodes = NULL;
             }
             break;
+        case RR_SHARDGROUPS_REPLACE:
+            if (req->r.shardgroups_replace.shardgroups != NULL) {
+                for(int i = 0; i < req->r.shardgroups_replace.len; i++) {
+                    RedisModule_Free(req->r.shardgroups_replace.shardgroups[i]);
```
tezc, line 1360: Also need to call ShardGroupFree() here?

raft.c
```diff
@@ -2346,6 +2361,89 @@ void applyShardGroupChange(RedisRaftCtx *rr, raft_entry_t *entry)
...
+        RedisModuleDictIter *iter = RedisModule_DictIteratorStartC(si->shard_group_map, "^", NULL, 0);
+
+        char *key;
+        size_t key_len;
+        ShardGroup *data;
+
+        while ((key = RedisModule_DictNextC(iter, &key_len, (void **) &data)) != NULL) {
+            /* local shardgroup will not have a filled in id */
+            if (*data->id != 0) {
+                RedisModule_Free(data);
```
tezc, line 2380: Similar here probably, need to call ShardGroupFree() ?

cluster.c
```diff
@@ -95,19 +95,16 @@ char *ShardGroupSerialize(ShardGroup *sg)
...
+    char *nl;
+    char *endptr;
     memset(sg, 0, sizeof(*sg));
 
-    /* Find and null terminate shard group id */
-    char *nl = strchr(s, '\n');
-    if (!nl) {
+    /* Find shard group id */
+    nl = strchr(s, '\n');
+    if (!nl || nl - buf > buf_len) {
```
tezc, line 105: Just fyi, you can use `memchr()` if you're worried buf may not have '\0' at the end. Because if there is no '\0' at the end, if `nl - buf > buf_len`  happens, means `strchr()` went beyond `buf_len`, it may trigger segfault.

raft.c
```diff
@@ -2346,6 +2361,89 @@ void applyShardGroupChange(RedisRaftCtx *rr, raft_entry_t *entry)
...
+                local = data;
+            }
+        }
+        RedisModule_DictIteratorStop(iter);
+        RedisModule_FreeDict(rr->ctx, si->shard_group_map);
+        si->shard_group_map = NULL;
+    }
+
+    /* if we didn't identify the local shard group, we have a problem */
+    RedisModule_Assert(local != NULL);
```
tezc, line 2391: So, this should go into "if check" above to set `local`. Do we always have `si->shard_group_map`? Can it be null if "replace" is the first command. I'm not sure, just asking, if user mistake can trigger an assert.

cluster.c
sjpotter: ok, will switch it to that.

cluster.c
sjpotter: ok, removed

cluster.c
sjpotter: I think its fine as is.  in general entry data can't be assumed to be a C string, so useful to always depend on data_len part of it, and not sometimes C string and sometimes not as then it gets confusing.

cluster.c
sjpotter: yes

raft.c
sjpotter: yes, we always have a shardgroup.  its created at node creation time (even before cluster init / join)

tezc: Okay, nice. So, probably null checks for it are leftover in the codebase, maybe better to remove them to avoid confusion.

cluster.c
sjpotter: added this.

raft.c
sjpotter: yes and another place too I found

tezc:
@sjpotter I couldn't suggest over Github, added a commit to simplify memchr checks, hope it's okay.

tezc (approved):

PR #206, opened 2021-12-30
enable RAFT.CLUSTER INIT to define the local cluster id string

sjpotter:
it is enforced to be 32 characters

### DIFF
--- a/redisraft.c
+++ b/redisraft.c
@@ -524,8 +524,9 @@ error:
     return REDISMODULE_OK;
 }
 
-/* RAFT.CLUSTER INIT
+/* RAFT.CLUSTER INIT <id>
  *   Initializes a new Raft cluster.
+ *   <id> is an optional 32 character string, if set, cluster will use it for the id
  * Reply:
  *   +OK [dbid]
  *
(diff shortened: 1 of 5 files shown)


redisraft.c
```diff
@@ -548,11 +549,14 @@ static int cmdRaftCluster(RedisModuleCtx *ctx, RedisModuleString **argv, int arg
...
     if (!strncasecmp(cmd, "INIT", cmd_len)) {
-        if (argc != 2) {
+        if (argc < 2 || argc > 3) {
             RedisModule_WrongArity(ctx);
             return REDISMODULE_OK;
         }
         req = RaftReqInit(ctx, RR_CLUSTER_INIT);
+        if (argc == 3) {
+            req->r.cluster_init.id = RedisModule_CreateStringFromString(ctx, argv[2]);
+        }
```
tezc, line 559: How about defining `cluster_init.id` as char[RAFT_DB_ID_LEN + 1] to avoid dynamic string handling? I didn't try the code but something like this may work : 
```suggestion
        char cluster_id[RAFT_DB_ID_LEN + 1];
        if (argc == 2) {
             RedisModule_GetRandomHexChars(cluster_id, RAFT_DBID_LEN);
             cluster_id[RAFT_DBID_LEN] = '\0';
        } else {
            size_t len;
            char *reqId = RedisModule_StringPtrLen(argv[2], &len);
            if (len != RAFT_DBID_LEN) {
                  RedisModule_ReplyWithError(ctx, "ERR cluster id must be 32 characters");
                  return REDISMODULE_OK;
            }
            memcpy(cluster_id, reqId, RAFT_DBID_LEN);
            cluster_id[RAFT_DBID_LEN] = '\0';
        } 
        
        req = RaftReqInit(ctx, RR_CLUSTER_INIT);
        mempy(req->r.cluster_init.id, cluster_id, RAFT_DBID_LEN + 1);
        
```

sjpotter:
note that the test I dded in the pr will fail right now as depends on shardgroup update pr for now (both would want to add a field th raft.info)

tezc (approved):

PR #212, opened 2022-01-04
fix raftNodeHasSufficientLogs assert

sjpotter:
raft_recv_entry (for config changes, which this does) can fail.  this assumed it couldn't.

the caller of this callback already expected it could fail.


tezc:
Just to note an example, if leader transfership is in progress when raftNodeHasSufficientLogs is called, `raft_recv_entry()` will reject the entry.

tezc (approved):

PR #215, opened 2022-01-06
Add fsync stats to raft.info

tezc:


yossigo (commented):
@tezc Redis has a monotonic time facility (see `monotonic.h`), which is not currently exposed to modules but might be useful here. Maybe it's an opportunity to add an API for it. WDUT?

tezc:
@yossigo. Is it really useful to have this in module API? It's basically a few lines of code, just a call to `clock_gettime()` and it has no dependency to Redis itself (we'll be just substituting a libc call with an API call). Doesn't this make API bigger without significant benefits?

yossigo:
@tezc To summarize our discussion, the answer is probably yes and the reason is the Redis monotonic timer mechanism supports different implementations including directly accessing a CPU timer on amd64 and arm64 platforms.

yossigo (dismissed):

yossigo (dismissed):

tezc:
Sorry I'll need another approval, CI is failed and need to rebase the PR.

yossigo (approved):

PR #218, opened 2022-01-06
Enable slot-config to be set with raft.config set before init/join

sjpotter:
This is to enable orchestrators to configure redisraft nodes not just via the command line.

### DIFF
--- a/config.c
+++ b/config.c
@@ -272,19 +272,19 @@ invalid_value:
 
 void handleConfigSet(RedisRaftCtx *rr, RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
 {
-    if (argc != 4) {
+    if (argc != 2) {
         RedisModule_WrongArity(ctx);
         return;
     }
 
     size_t key_len;
-    const char *key = RedisModule_StringPtrLen(argv[2], &key_len);
+    const char *key = RedisModule_StringPtrLen(argv[0], &key_len);
     char keybuf[key_len + 1];
     memcpy(keybuf, key, key_len);
     keybuf[key_len] = '\0';
 
     size_t value_len;
-    const char *value = RedisModule_StringPtrLen(argv[3], &value_len);
+    const char *value = RedisModule_StringPtrLen(argv[1], &value_len);
     char valuebuf[value_len + 1];
     memcpy(valuebuf, value, value_len);
     valuebuf[value_len] = '\0';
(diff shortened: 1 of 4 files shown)


redisraft.h
```diff
@@ -334,7 +334,7 @@ typedef struct RedisRaftConfig {
     bool raft_log_fsync;
     /* Cluster mode */
     bool sharding;                      /* Are we running in a sharding configuration? */
-    char *slot_config;                  /* Defining multiple slot ranges (# or #:#) that are delimited by ',' */
+    const char *slot_config;                  /* Defining multiple slot ranges (# or #:#) that are delimited by ',' */
```
tezc, line 337:
```suggestion
    const char *slot_config;            /* Defining multiple slot ranges (# or #:#) that are delimited by ',' */
```

config.c
```diff
@@ -229,11 +234,19 @@ static RRStatus processConfigParam(const char *keyword, const char *value,
...
+        if (!validSlotConfig(val)) {
+            RedisModule_Free(val);
+            snprintf(errbuf, errbuflen - 1, "invalid 'slot_config' value");
             return RR_ERROR;
         }
+        if (target->slot_config != DEFAULT_SLOT_CONFIG) {
+            RedisModule_Free((void *) target->slot_config);
+        }
+        target->slot_config = val;
+        if (redis_raft.sharding_info) {
```
tezc, line 247: redis_raft lives in raft thread context. So, probably need to create a req and handle config changes in raft thread.

config.c
```diff
@@ -112,12 +113,16 @@ static RRStatus processConfigParam(const char *keyword, const char *value,
...
-                !strcmp(keyword, CONF_SLOT_CONFIG) ||
                 !strcmp(keyword, CONF_EXTERNAL_SHARDING))) {
         snprintf(errbuf, errbuflen-1, "'%s' only supported at load time", keyword);
         return RR_ERROR;
     }
 
+    if (redis_raft.state != REDIS_RAFT_UNINITIALIZED && !strcmp(keyword, CONF_SLOT_CONFIG)) {
+        snprintf(errbuf, errbuflen-1, "'%s' only supported before cluster init/join", keyword);
+        return RR_ERROR;
+    }
```
tezc, line 124: Looks like this function doesn't know about redis_raft. Maybe we should add `before_join` parameter similar to `on_init`. We'll pass `state == REDIS_RAFT_UNINITIALIZED` into it. Then, reject slot config if `before_join` is false. That requires initializing shard info after this function(rather than calling `ShardingInfoReset()` here). Probably something like just before setting `state = REDIS_RAFT_UP`, we init sharding info if it is not initialized already. That way we'll have updated config if any.

config.c
sjpotter: added all config get/set to raft thread

config.c
sjpotter: what do you mean by doesn't know?

tezc: There is no usage of `redis_raft` here in this config.c file. Instead of checking `redis_raft.state != REDIS_RAFT_UNINITIALIZED`, this function accepts `on_init` parameter. Otherwise, it could do that check inside, it wouldn't need that function parameter. 
So, should we follow that convention? Can we avoid doing `redis_raft.state !=  REDIS_RAFT_UNINITIALIZED` and avoid calling `ShardingInfoReset()` here ?

sjpotter:
this pr is deprecated in favor of https://github.com/RedisLabs/redisraft/pull/228

PR #221, opened 2022-01-08
Add sanitizer support

tezc:
- Added address sanitizer to PR CI.
- Created daily CI, added address and undefined sanitizer.
- Fixed memory leaks.


.github/workflows/ci.yml
```diff
@@ -3,11 +3,8 @@ name: CI
 on: [push, pull_request]
 
 jobs:
-  build-ubuntu:
-    strategy:
-      matrix:
-        platform: [ubuntu-latest, ubuntu-18.04]
-    runs-on: ${{ matrix.platform }}
+  test-ubuntu:
+    runs-on: ubuntu-latest
```
yossigo, line 7: Just a thought, since the sanitizer is just a flag maybe we can use strategy here (sanitizer / non-sanitizer) and save the duplicate code? Only if it doesn't turn out to be over complicated.

yossigo (approved):

.github/workflows/ci.yml
tezc: For regular build, we pass gcov/Coverage flags and upload its result to codecov. I rather duplicate the code here to avoid maintaining these differences.

PR #222, opened 2022-01-09
Measure latency of many RedisRaft operations

sjpotter:
adds measurments using the Redis Latency sampling interface

### DIFF
--- a/raft.c
+++ b/raft.c
@@ -130,7 +130,7 @@ static void entryAttachRaftReq(RedisRaftCtx *rr, raft_entry_t *entry, RaftReq *r
 static void executeRaftRedisCommandArray(RaftRedisCommandArray *array,
     RedisModuleCtx *ctx, RedisModuleCtx *reply_ctx)
 {
-    uint64_t begin = timeMonotonicNanos();
+    mstime_t begin = RedisModule_Milliseconds();
 
     int i;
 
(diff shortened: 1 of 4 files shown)


raft.c
```diff
@@ -149,11 +152,13 @@ static void executeRaftRedisCommandArray(RaftRedisCommandArray *array,
             continue;
         }
 
+        uint64_t begin_rmcall = timeMonotonicNanos();
         enterRedisModuleCall();
         RedisModuleCallReply *reply = RedisModule_Call(
                 ctx, cmd, redis_raft.resp_call_fmt, &c->argv[1], c->argc - 1);
         int ret_errno = errno;
         exitRedisModuleCall();
+        RedisModule_LatencyAddSample("rm_call", (mstime_t) ((timeMonotonicNanos() - begin_rmcall) / 1000000));
```
tezc, line 161: Looks like `RedisModule_LatencyAddSample()` has millisecond precision only, so maybe we just use RedisModule_Milliseconds(). It has less precise timer call (gettimeofday(), less precise, better for performance).

raft.c
```diff
@@ -149,11 +152,13 @@ static void executeRaftRedisCommandArray(RaftRedisCommandArray *array,
             continue;
         }
 
+        uint64_t begin_rmcall = timeMonotonicNanos();
```
tezc, line 155: I see that inside RedisModule_Call(), it measures time passed. Maybe we avoid doing that here?

raft.c
```diff
@@ -1367,6 +1404,8 @@ void RaftReqFree(RaftReq *req)
 
         RedisModule_UnblockClient(req->client, NULL);
     }
+
+    RedisModule_LatencyAddSample(RaftReqTypeStr[req->type],  (mstime_t) ((timeMonotonicNanos() - req->begin) / 1000000));
```
tezc, line 1408: RedisModule_LatencyAddSample() does not seem thread-safe to me. So, I believe we should guarantee that we call it only in raft thread. Seems like we can call RaftReqFree() in redis thread itself? Hope I'm all wrong :)

sjpotter:
not happening

PR #226, opened 2022-01-13
Extend raft.shardgroup replace to change hash slot owneship

sjpotter:
- shardgroup replace will validate options before submitting to log
	a given slot, can either only be specified as stable or as (migrating and/or importing)

- ShardingInfoReset() no longer adds local shardgroup to sharding info.
	- its added at
		- cluster init - with default cluster slots (AddBasicLocalShardGroup())
		- cluster join - with default cluster slots
		- reloading rdb file on restart (if exists, with saved slot ranges)
		- reloading without rdb file (before log playback) with default cluste slots

	- the way cluster slots is modified is via a shardgroup replace log entry
		- can be called after init
		- on join will get log entry and change
		- on replay of log will change

- in practice, the local shardgroup struct will also contain the nodes passed to it in the command, but those are given to it by the external orchestrator, so aren't the real source of truth.  the real source of truth for its nodes is itself, so for cluster slots/nodes commands, it ignores those entries and uses its own view of its world.

lots of small changes
- in rdb snapshot serialize out local shardgroup
- rename/creation of new hash slot mappings
	for stable (rename), importing/exportng

todo: currently, as an artifact of older code, we treat the local shardgroup's id as an emptry string (as it's struct was allocated before dbid was specified).  In new model, that doesn't have to be the case.

the primary advantage of keeping it the way it is, is that it makes it easy to find the local shardgroup (*sg->id == 0).  However, it is a bit ugly and perhaps better to just memcmp() it in the future.

### DIFF
--- a/cluster.c
+++ b/cluster.c
@@ -876,37 +876,45 @@ RRStatus ShardingInfoAddShardGroup(RedisRaftCtx *rr, ShardGroup *sg)
  * and RR_ERROR is returned.
  */
 
-RRStatus ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc, ShardGroup *sg)
+int ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc, ShardGroup *sg, int base_argv_count)
 {
     const char *id;
     size_t len;
     long long num_slots, num_nodes;
-
-    ShardGroupInit(sg);
+    int argidx = 0;
 
     id = RedisModule_StringPtrLen(argv[0], &len);
     if (len > RAFT_DBID_LEN) {
+        LOG_ERROR("ShardGroupParse failed; argv[%d] shard_group id is too long", base_argv_count + argidx);
         RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (shard group id length)");
         goto error;
     }
+    argidx++;
+
     memcpy(sg->id, id, len);
     sg->id[RAFT_DBID_LEN] = '\0';
 
     if (RedisModule_StringToLongLong(argv[1], &num_slots) != REDISMODULE_OK) {
+        LOG_ERROR("ShardGroupParse failed; argv[%d]: couldn't parse num_slots", base_argv_count + argidx);
         RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (num slots)");
         goto error;
     }
+    argidx++;
+
     if (RedisModule_StringToLongLong(argv[2], &num_nodes) != REDISMODULE_OK) {
+        LOG_ERROR("ShardGroupParse failed; argv[%d]: couldn't parse num_nodes", base_argv_count + argidx);
         RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (num nodes)");
         goto error;
     }
+    argidx++;
 
     /* Validate node arguments count is correct */
-    if ((argc - 3) != (num_slots * 3) + (num_nodes * 2)) {
+    int num_argv_entries = (int) (3 + num_slots * 3 + num_nodes * 2);
+    if (num_argv_entries > argc) {
+        LOG_ERROR("ShardGroupParse failed; not enough args to parse: want %d have %d", num_argv_entries, argc);
         RedisModule_WrongArity(ctx);
         goto error;
     }
-    int argidx = 3; /* Next arg to consume */
 
     /* Parse slots */
     sg->slot_ranges_num = num_slots;
@@ -916,8 +924,9 @@ RRStatus ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
         if (RedisModule_StringToLongLong(argv[argidx++], &start_slot) != REDISMODULE_OK ||
             RedisModule_StringToLongLong(argv[argidx++], &end_slot) != REDISMODULE_OK ||
             RedisModule_StringToLongLong(argv[argidx++], &type) != REDISMODULE_OK ||
-            !HashSlotRangeValid(start_slot, end_slot) ||
+            !HashSlotRangeValid((int) start_slot, (int) end_slot) ||
             !SlotRangeTypeValid(type)) {
+            LOG_ERROR("ShardGroupParse failed; argv[%d]:argv[%d]: invalid parse slot_range", base_argv_count + argidx - 3,  base_argv_count + argidx - 1);
             RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (slot range)");
             goto error;
         }
@@ -930,36 +939,38 @@ RRStatus ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
     sg->nodes_num = num_nodes;
     sg->nodes = RedisModule_Calloc(num_nodes, sizeof(ShardGroupNode));
     for (int i = 0; i < num_nodes; i++) {
-        size_t len;
-        const char *str = RedisModule_StringPtrLen(argv[argidx++], &len);
+        const char *str = RedisModule_StringPtrLen(argv[argidx], &len);
 
         if (len != RAFT_SHARDGROUP_NODEID_LEN) {
+            LOG_ERROR("ShardGroupParse failed; argv[%d] node id is too long", base_argv_count + argidx);
             RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (node id length)");
             goto error;
         }
+        argidx++;
 
         memcpy(sg->nodes[i].node_id, str, len);
         sg->nodes[i].node_id[len] = '\0';
 
-        str = RedisModule_StringPtrLen(argv[argidx++], &len);
+        str = RedisModule_StringPtrLen(argv[argidx], &len);
         if (!NodeAddrParse(str, len, &sg->nodes[i].addr)) {
+            LOG_ERROR("ShardGroupParse failed; argv[%d] failed to parse node", base_argv_count + argidx);
             RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (node address/port)");
             goto error;
         }
+        argidx++;
     }
 
-    return RR_OK;
+    return num_argv_entries;
 
 error:
-    ShardGroupTerm(sg);
-    return RR_ERROR;
+    return -1;
 }
 
 /* parses all the shardgroups in a replace call and validates them */
 RRStatus ShardGroupsParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc, RaftReq *req)
 {
     if (RedisModule_StringToLongLong(argv[0], &req->r.shardgroups_replace.len) != REDISMODULE_OK) {
-        printf("ERR invalid shard group message(1)");
+        LOG_ERROR("Invalid shard group message: argv[0] (number of shard groups)");
         RedisModule_ReplyWithError(ctx, "ERR invalid shard group message(1)");
         return RR_ERROR;
     }
@@ -967,38 +978,27 @@ RRStatus ShardGroupsParse(RedisModuleCtx *ctx, RedisModuleString **argv, int arg
             RedisModule_Calloc(req->r.shardgroups_replace.len, sizeof(ShardGroup *));
     argv++;
     argc--;
+    int argv_count = 1;
 
     int i;
     for(i = 0; i < req->r.shardgroups_replace.len; i++) {
         ShardGroup *sg = ShardGroupCreate();
         req->r.shardgroups_replace.shardgroups[i] = sg;
 
-        long long num_slots, num_nodes;
-        if (RedisModule_StringToLongLong(argv[1], &num_slots) != REDISMODULE_OK ) {
-            printf("ERR invalid shard group message (num slots)");
-            RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (num slots)");
-            goto fail;
-        }
-        if (RedisModule_StringToLongLong(argv[2], &num_nodes) != REDISMODULE_OK) {
-            printf("ERR invalid shard group message (num nodes)");
-            RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (num nodes)");
-            goto fail;
-        }
-
-        int num_argv_entries = (int) (3 + num_slots * 3 + num_nodes * 2);
-        if (num_argv_entries > argc) {
-            printf("ERR invalid shard group message (num entries)");
-            RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (num entries)");
-            goto fail;
-        }
-
-        if (ShardGroupParse(ctx, argv, num_argv_entries, sg) != RR_OK) {
-            printf("ShardGroupParse failed!\n");
+        int num_argv_entries = 0;
+        if ((num_argv_entries = ShardGroupParse(ctx, argv, argc, sg, argv_count)) < 0) {
             goto fail;
         }
 
         argv += num_argv_entries;
         argc -= num_argv_entries;
+        argv_count += num_argv_entries;
+    }
+
+    if (argc != 0) {
+        LOG_ERROR("Invalid shard group message: extra unexpected arguments");
+        RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (extra arguments)");
+        goto fail;
     }
 
     /* basic validation, no shard groups have slotranges that overlap another */
--- a/cluster.c
+++ b/cluster.c
@@ -758,8 +758,19 @@ RRStatus ShardingInfoValidateShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
 
         for (int i = new_sg->slot_ranges[h].start_slot; i <= new_sg->slot_ranges[h].end_slot; i++) {
             if (si->stable_slots_map[i] != 0) {
-                LOG_ERROR("Invalid shardgroup: hash slot already mapped: %u", i);
+                LOG_ERROR("Invalid shardgroup: hash slot already mapped as stable: %u", i);
                 return RR_ERROR;
+            }
+            if (new_sg->slot_ranges[h].type == SLOTRANGE_TYPE_MIGRATING) {
+                if (si->migrating_slots_map[i] != 0) {
+                    LOG_ERROR("Invalid shardgroup: hash slot already mapped as migrating: %u", i);
+                    return RR_ERROR;
+                }
+            } else if (new_sg->slot_ranges[h].type == SLOTRANGE_TYPE_IMPORTING) {
+                if (si->importing_slots_map[i] != 0) {
+                    LOG_ERROR("Invalid shardgroup: hash slot already mapped as importing: %u", i);
+                    return RR_ERROR;
+                }
             }
         }
     }
--- a/cluster.c
+++ b/cluster.c
@@ -621,7 +621,6 @@ void ShardingInfoRDBSave(RedisModuleIO *rdb)
         return;
     }
 
-    /* When saving, skip shardgroup #1 which is the local cluster */
     RedisModule_SaveUnsigned(rdb, si->shard_groups_num);
     if (si->shard_group_map != NULL) {
         RedisModuleDictIter *iter = RedisModule_DictIteratorStartC(si->shard_group_map, "^", NULL, 0);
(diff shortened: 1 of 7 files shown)


cluster.c
```diff
@@ -621,20 +621,16 @@ void ShardingInfoRDBSave(RedisModuleIO *rdb)
         return;
     }
 
-    /* When saving, skip shardgroup #1 which is the local cluster */
-    RedisModule_SaveUnsigned(rdb, si->shard_groups_num - 1);
```
sjpotter, line 625: we know serialize out the local shardgroup as well as it can change and if if its snapshotted, we wont have a log entry for it.

cluster.c
```diff
@@ -621,20 +621,16 @@ void ShardingInfoRDBSave(RedisModuleIO *rdb)
...
         RedisModuleDictIter *iter = RedisModule_DictIteratorStartC(si->shard_group_map, "^", NULL, 0);
 
         size_t key_len;
         ShardGroup *sg;
 
         while (RedisModule_DictNextC(iter, &key_len, (void **) &sg) != NULL) {
-            if (*sg->id == 0) { /* skipping local cluster */
-                continue;
-            }
-
```
sjpotter, line 636: as noted above, no longer skip local shard group

cluster.c
```diff
@@ -678,15 +674,6 @@ void ShardingInfoRDBLoad(RedisModuleIO *rdb)
...
     unsigned int rdb_shard_groups_num = RedisModule_LoadUnsigned(rdb);
-    if (!rdb_shard_groups_num) {
-        /* No shardgroups. This could mean no sharding, or simply no shardgroups
-         * to read. If we have ShardingInfo we'll reset it.
-         */
-        if (si)
-            ShardingInfoReset(rr);
-
-        return;
-    }
```
sjpotter, line 689: this is not neccessary anymore as a) rdb_shard_group_num will always be at least 1 and the loop below will handle local fine (and would still handle it fine even if it was 0).  b) si can no longer be null (older change) and hence the loop below would also work fine and dont have to special case it.

cluster.c
```diff
@@ -783,7 +781,7 @@ RRStatus ShardingInfoValidateShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
 /* Update an existing ShardGroup in the active ShardingInfo.
  *
  * FIXME: We currently only handle updating nodes but don't support remapping
- *        hash slots.
+ *        hash slots, via this mechanism
```
sjpotter, line 784: in practice, this is only used for the internal gossipy sharding mechanism, where the thread can update individual shardgroups with an updated set of nodes, but isn't about replacing the whole set of shard groups as one gets with an external orchestrator.

The question becomes, should the internal mechanism be moved to replace (it would have to construct a whole list)

cluster.c
```diff
@@ -1026,40 +1088,6 @@ void ShardingInfoInit(RedisRaftCtx *rr)
...
-            sg->slot_ranges[i].start_slot = val;
-            sg->slot_ranges[i].end_slot = val;
-        }
-        sg->slot_ranges[i].type = SLOTRANGE_TYPE_STABLE;
-
-        token = strtok_r(NULL, ",", &saveptr);
-    }
-
-     RedisModule_Free(str);
-}
```
sjpotter, line 1061: moved to utils.c

cluster.c
```diff
@@ -1132,6 +1149,7 @@ RRStatus computeHashSlot(RedisRaftCtx *rr, RaftReq *req)
             } else {
                 if (slot != thisslot) {
                     RedisModule_Free(keyindex);
+                    RedisModule_ReplyWithError(req->ctx, "CROSSSLOT Keys in request don't hash to the same slot");
```
sjpotter, line 1152: moved ReplyWithError here, so we can return others errors as well in computeHashSlot in future.

redisraft.c
```diff
@@ -651,7 +651,7 @@ static int cmdRaftShardGroup(RedisModuleCtx *ctx, RedisModuleString **argv, int
         }
 
         req = RaftReqInit(ctx, RR_SHARDGROUP_ADD);
-        if (ShardGroupParse(ctx, &argv[2], argc - 2, &req->r.shardgroup_add) != RR_OK) {
+        if (ShardGroupParse(ctx, &argv[2], argc - 2, &req->r.shardgroup_add, 2) < 0) {
```
sjpotter, line 654: no longer an RR_STATUS, now it return -1 on error, or >= 0 (with number of args it consumed as the result) on success.  In practice, it might even >= 3 alwaysy

cluster.c
```diff
@@ -1088,22 +1116,11 @@ void ShardingInfoReset(RedisRaftCtx *rr)
...
-        .slot_ranges = NULL,
-        .nodes_num = 0,
-        .nodes = NULL
-    };
-
-    FillShardSlots(rr, &sg);
-
-    RRStatus ret = ShardingInfoAddShardGroup(rr, &sg);
-    RedisModule_Assert(ret == RR_OK);
-    ShardGroupTerm(&sg);
```
sjpotter, line 1106: this work is now done by AddBasicLocalShardGroup() if sharding info doesn't exist otherwise at the appropriate point in time (i.e. init / join / reload)

raft.c
```diff
@@ -1906,26 +1912,29 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr, RaftReq *req)
...
     if (req->r.redis.hash_slot == -1) return RR_OK;
 
     /* Make sure hash slot is mapped and handled locally. */
-    ShardGroup *sg = rr->sharding_info->hash_slots_map[req->r.redis.hash_slot];
+    ShardGroup *sg = rr->sharding_info->stable_slots_map[req->r.redis.hash_slot];
     if (!sg) {
-        RedisModule_ReplyWithError(req->ctx, "CLUSTERDOWN Hash slot is not served");
-        return RR_ERROR;
+        /* TODO: will have to be smarter with actual migration happening */
+        sg = rr->sharding_info->migrating_slots_map[req->r.redis.hash_slot];
```
sjpotter, line 1925: this might be too early here, as adding in some migrating slot logic

sjpotter:
I made a pass through over this, making notes to help reviewers.

I might have gone a step too far in checking the migrating slot list in one place for this PR, but besides that it seems to be good.

Besides that, what isn't done well, is handling cluster slots / nodes for migrating/importing slots.  not 100% sure I understand the syntax yet (it seems, as opposed to stable slots which can be slot ranges, importing/migrating have to be listed one at a time?)

util.c
```diff
@@ -394,3 +394,50 @@ void handleRMCallError(RedisModuleCtx *ctx, int ret_errno, const char *cmd, size
...
+    sg->slot_ranges = RedisModule_Calloc(sg->slot_ranges_num, sizeof(ShardGroupSlotRange));
+
+    char *saveptr = NULL;
+    char *token = strtok_r(str, ",", &saveptr);
+    for (int i = 0; i < sg->slot_ranges_num; i++) {
+        unsigned long val;
+        if ((pos = strchr(token, ':'))) {
+            *pos = '\0';
+            val = strtoul(token, NULL, 10);
+            sg->slot_ranges[i].start_slot = val;
```
yossigo, line 416: Maybe add a function comment that indicates we assume input slot config is valid?

yossigo, line 416: Maybe add a function comment that indicates we assume input slot config is valid?

cluster.c
```diff
@@ -230,7 +230,7 @@ void ShardGroupTerm(ShardGroup *sg)
     }
 }
 
-ShardGroup *ShardGroupCreate() {
+ShardGroup* ShardGroupCreate() {
```
yossigo, line 233: nit: this breaks the style.

cluster.c
yossigo: For now, this is useful for testing if not anything else. In the future it could become irrelevant.

cluster.c
```diff
@@ -882,37 +876,45 @@ RRStatus ShardingInfoAddShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
  * and RR_ERROR is returned.
```
yossigo, line 876: Update comment to match new function prototype?

cluster.c
```diff
@@ -936,70 +939,129 @@ RRStatus ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
...
+    }
+
+    if (argc != 0) {
+        LOG_ERROR("Invalid shard group message: extra unexpected arguments");
+        RedisModule_ReplyWithError(ctx, "ERR invalid shard group message (extra arguments)");
+        goto fail;
+    }
+
+    /* basic validation, no shard groups have slotranges that overlap another */
+    ShardGroup * stable[REDIS_RAFT_HASH_MAX_SLOT+1];
```
yossigo, line 1005: nit: pointer style (RedisRaft uses `ShardGroup *var` elsewhere).
Also, why not `ShardGroup *var[...] = { 0 }`?

cluster.c
yossigo: So maybe rename it to computeHashSlotOrReplyError?

redisraft.h
```diff
@@ -485,7 +486,9 @@ typedef struct ShardingInfo {
      * since a zero value indicates the slot is unassigned. The index
      * should therefore be adjusted before refering the array.
      */
-    ShardGroup *hash_slots_map[REDIS_RAFT_HASH_SLOTS];
+    ShardGroup *stable_slots_map[REDIS_RAFT_HASH_SLOTS];
```
yossigo, line 489: What is the rationale behind storing three maps vs. a single map and hold the state in the ShardGroup?

cluster.c
sjpotter: ok

cluster.c
sjpotter: I guess I didn't intuitively know that one can do that, changed the code.

redisraft.h
sjpotter: because in a migrating case, one will have migrating and importing shard groups for the same slot.

cluster.c
```diff
@@ -697,31 +684,33 @@ void ShardingInfoRDBLoad(RedisModuleIO *rdb)
...
-        RedisModule_Assert(len < sizeof(sg.id));
-        memcpy(sg.id, buf, len);
-        sg.id[len] = '\0';
+        RedisModule_Assert(len < sizeof(sg->id));
+        memcpy(sg->id, buf, len);
+        sg->id[len] = '\0';
         RedisModule_Free(buf);
+        if (RedisModule_LoadUnsigned(rdb)) {
+            sg->local = true;
+        }
```
tezc, line 697:
```suggestion
        sg->local = RedisModule_LoadUnsigned(rdb);
```

Would this be easier to understand?

cluster.c
```diff
@@ -770,10 +757,21 @@ RRStatus ShardingInfoValidateShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
         }
 
         for (int i = new_sg->slot_ranges[h].start_slot; i <= new_sg->slot_ranges[h].end_slot; i++) {
-            if (si->hash_slots_map[i] != 0) {
-                LOG_ERROR("Invalid shardgroup: hash slot already mapped: %u", i);
+            if (si->stable_slots_map[i] != 0) {
+                LOG_ERROR("Invalid shardgroup: hash slot already mapped as stable: %u", i);
```
tezc, line 761: Minor but just for type safety and silence linter, should `i` be unsigned here?

cluster.c
```diff
@@ -770,10 +757,21 @@ RRStatus ShardingInfoValidateShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
...
                 return RR_ERROR;
             }
+            if (new_sg->slot_ranges[h].type == SLOTRANGE_TYPE_MIGRATING) {
+                if (si->migrating_slots_map[i] != 0) {
```
tezc, line 765:
```suggestion
                if (si->migrating_slots_map[i] != NULL) {
```
Do you think it is better to compare against NULL here and in other if checks as well (just for clarity)?

cluster.c
```diff
@@ -936,70 +939,126 @@ RRStatus ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
...
+        goto fail;
+    }
+
+    /* basic validation, no shard groups have slotranges that overlap another */
+    ShardGroup *stable[REDIS_RAFT_HASH_MAX_SLOT+1] = {0};
+    ShardGroup *importing[REDIS_RAFT_HASH_MAX_SLOT+1] = {0};
+    ShardGroup *migrating[REDIS_RAFT_HASH_MAX_SLOT+1] = {0};
+
+    for (int j = 0; j < req->r.shardgroups_replace.len; j++) {
+        ShardGroup * sg = req->r.shardgroups_replace.shardgroups[j];
```
tezc, line 1010:
```suggestion
        ShardGroup *sg = req->r.shardgroups_replace.shardgroups[j];
```

cluster.c
```diff
@@ -936,70 +939,126 @@ RRStatus ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int argc
...
+        for (int k = 0; k < sg->slot_ranges_num; k++) {
+            ShardGroupSlotRange * sr = &sg->slot_ranges[k];
+            switch (sr->type) {
+                case SLOTRANGE_TYPE_STABLE:
+                    for(unsigned int l=sr->start_slot; l <= sr->end_slot; l++) {
+                        if (stable[l]) {
+                            printf("ERR Unable to validate shard groups: shard group already owns this slot\n");
```
tezc, line 1017: Are these `printf()` 's leftover?

cluster.c
```diff
@@ -1132,6 +1146,7 @@ RRStatus computeHashSlot(RedisRaftCtx *rr, RaftReq *req)
             } else {
                 if (slot != thisslot) {
                     RedisModule_Free(keyindex);
+                    RedisModule_ReplyWithError(req->ctx, "CROSSSLOT Keys in request don't hash to the same slot");
```
tezc, line 1149: Looks like there is typo in CROSSSLOT but not sure what is the correct way. Btw should we start with ERR, something like `ERR cross slot keys ... `

raft.c
```diff
@@ -2321,31 +2329,30 @@ void handleDebug(RedisRaftCtx *rr, RaftReq *req)
  */
 void applyShardGroupChange(RedisRaftCtx *rr, raft_entry_t *entry)
 {
-    ShardGroup sg;
+    ShardGroup * sg = ShardGroupCreate();
     RRStatus ret;
 
-    if (ShardGroupDeserialize(entry->data, entry->data_len, &sg) != RR_OK) {
+    if (ShardGroupDeserialize(entry->data, entry->data_len, sg) != RR_OK) {
```
tezc, line 2335: Should we call `ShardGroupFree()` on failure?

raft.c
```diff
@@ -2321,31 +2329,30 @@ void handleDebug(RedisRaftCtx *rr, RaftReq *req)
...
         LOG_ERROR("Failed to deserialize ADD_SHARDGROUP payload: [%.*s]",
                 entry->data_len, entry->data);
         return;
     }
 
     switch (entry->type) {
         case RAFT_LOGTYPE_ADD_SHARDGROUP:
-            if ((ret = ShardingInfoAddShardGroup(rr, &sg)) != RR_OK)
+            if ((ret = ShardingInfoAddShardGroup(rr, sg)) != RR_OK)
```
tezc, line 2343: Should we call `ShardGroupFree()` on failure?

raft.c
```diff
@@ -2321,31 +2329,30 @@ void handleDebug(RedisRaftCtx *rr, RaftReq *req)
...
                 LOG_ERROR("Failed to add a shardgroup");
             break;
         case RAFT_LOGTYPE_UPDATE_SHARDGROUP:
-            if ((ret = ShardingInfoUpdateShardGroup(rr, &sg)) != RR_OK)
+            if ((ret = ShardingInfoUpdateShardGroup(rr, sg)) != RR_OK)
                 LOG_ERROR("Failed to update shardgroup");
+            ShardGroupFree(sg);
```
tezc, line 2349: Should this be inside if check?

raft.c
tezc: Hmm, maybe not. If I'm not mistaken, `ShardingInfoAddShardGroup()` destroys the object in some cases itself but not in all cases. Is it possible to revert `sg` to be stack variable and and let these functions create/copy ShardGroup object itself if needed. It's just not obvious who is the owner of the object, where should we destroy the object. Maybe we should you take another look at this.

It'll be nice to have a test case that hits these parts just to catch leak. (If we don't have already and easy to add one)

raft.c
```diff
@@ -2418,19 +2416,18 @@ void replaceShardGroups(RedisRaftCtx *rr, raft_entry_t *entry)
         RedisModule_Assert(endptr == nl);
         payload = nl + 1;
 
-        ShardGroup sg;
-        if (ShardGroupDeserialize(payload, payload_len, &sg) != RR_OK) {
+        ShardGroup *sg = ShardGroupCreate();
```
tezc, line 2419: Similar issue here, `sg` will leak on early return if I'm not mistaken.

cluster.c
sjpotter: yes

cluster.c
sjpotter: yes, this is an artifact of the old slot map which mapped to an index.  it worked, but NULL is better

cluster.c
sjpotter: well, sort of.  implicit cast, but you're probably right.

cluster.c
sjpotter: done

cluster.c
yossigo: Seems like this is consistent with Redis, there's no typo but "cross" and "slot" combined:

```
src/cluster.c:        addReplyError(c,"-CROSSSLOT Keys in request don't hash to the same slot");
```

The first keyword is used as the error code. `-ERR` is the generic one, but certain errors have their own designated code which helps clients handle this / map it to individual exceptions.

raft.c
sjpotter: fixed, by moving the ShardGroupCreate() into ShardGroupDeserialize and returning it via normal return conventions.  instead of RR_OK/RR_ERR, we now test that the it's not NULL.

cluster.c
tezc: Ah right, thought there was one extra 'S' in the middle.

raft.c
sjpotter: so, I figured out how to test this, needs some changes to applyShardGroupChange to change correctly, working on it.

raft.c
sjpotter: decided to remove from this pr

cluster.c
```diff
@@ -871,48 +870,62 @@ RRStatus ShardingInfoAddShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
...
 
 /* Parse a ShardGroup specification as passed directly to RAFT.SHARDGROUP ADD.
  * Shard group syntax is as follows:
  *
  *  [shardgroup id] [num_slots] [num_nodes] ([start slot] [end slot] [slot type])* ([node-uid node-addr:node-port])*
  *
- * If parsing errors are encountered, an error reply is generated on the supplied RedisModuleCtx,
- * and RR_ERROR is returned.
+ * If parsing errors are encountered, an error reply is generated on the supplied RedisModuleCtx and -1 is returned
+ * If it was processed successfully, the number of argv elements consumed is returned
```
yossigo: Something doesn't add up here.
I'd expect to have only input validation here, but not validation against applied slot mapping configuration, because by the time we apply this it could change (due to other uncommitted changes). Am I missing something here?

yossigo (commented):
@sjpotter I pushed a commit with a few minor cleanups. Main question is about validation at parse time vs. apply time and how do we deal with multiple concurrent changes that have not yet been committed.

cluster.c
sjpotter: I don't understand.   That's what happens.  we do input validation here (are the values passed in sane)

Its called in 2 places

1) ShardGroupsParse (plural)

this is used by REPLACE, in here it validates that the set is sane (i.e. is internally consistent) as it will replace the whole thing.

2) RR_SHARDGROUP_ADD  RaftReqs where its passed to

where it will eventually be handled by handleShardGroupAdd() which because this isn't a whole world set of data, is validated relative to the state at apply time.

sjpotter: i.e. a REPLACE operation doesn't depend on the eventual state at apply time, its replacing the entire state.

PR #228, opened 2022-01-16
Allow node id te be set by raft.config before cluster init/join

sjpotter:
Previously, could only set on command line, now allow it to be set before cluster join/init.

change test to not test that id can't be set by raft.config set


config.c
```diff
@@ -106,18 +106,21 @@ int validSlotConfig(char *slot_config) {
...
-                !strcmp(keyword, CONF_RAFT_LOG_FILENAME) ||
+    if (!on_init && (!strcmp(keyword, CONF_RAFT_LOG_FILENAME) ||
                 !strcmp(keyword, CONF_SLOT_CONFIG) ||
                 !strcmp(keyword, CONF_EXTERNAL_SHARDING))) {
         snprintf(errbuf, errbuflen-1, "'%s' only supported at load time", keyword);
         return RR_ERROR;
     }
 
+    if (!uninitialized && !strcmp(keyword, CONF_ID)) {
+        snprintf(errbuf, errbuflen-1, "'%s' only supported at before cluster init/join", keyword);
```
tezc, line 121:
```suggestion
        snprintf(errbuf, errbuflen-1, "'%s' only supported at before cluster init/join", keyword);
        return RR_ERROR;
```

config.c
```diff
@@ -277,7 +280,8 @@ void handleConfigSet(RedisRaftCtx *rr, RedisModuleCtx *ctx, RedisModuleString **
     valuebuf[value_len] = '\0';
 
     char errbuf[256] = "ERR ";
-    if (processConfigParam(keybuf, valuebuf, rr->config, false,
+    if (processConfigParam(keybuf, valuebuf, rr->config,
+                           false, rr->state == REDIS_RAFT_UNINITIALIZED,
```
tezc, line 284: Redis thread will call this function and access is not synchronised with raft thread. (e.g rr->state is updated from raft thread). I think, we still need to create RaftReq and handle this change in raft thread.

tests/integration/test_config.py
```diff
@@ -47,7 +47,7 @@ def test_config_startup_only_params(cluster):
 
     r1 = cluster.add_node()
     with raises(ResponseError, match='.*only supported at load time'):
-        r1.raft_config_set('id', 2)
+        r1.raft_config_set('external-sharding', 'yes')
```
tezc, line 50: Should we add some tests?

config.c
sjpotter: right.  though isn't that in practice true for all set/get config variables?  or do we not care so much about the rest?

tezc: Yes, true for all. If we move this to raft thread, we are fixing the issue for all I guess. If it requires complex changes, we can create a follow-up PR if you want.

tezc (approved):

PR #230, opened 2022-01-24
 Prevent running Redis test suite on forks

tezc:
- Prevent running Redis test suite on forks
- Added manual dispatch options

### DIFF
--- a/.github/workflows/redis-suite.yml
+++ b/.github/workflows/redis-suite.yml
@@ -5,41 +5,31 @@ on:
     - cron: '0 0 * * *'
   workflow_dispatch:
     inputs:
-      redis_repo:
-        description: 'redis repo owner and name'
-        required: false
-        default: 'redis/redis'
-      redis_branch:
-        description: 'redis git branch or sha to use'
-        required: false
-        default: 'unstable'
-      redisraft_repo:
-        description: 'redisraft repo owner and name'
+      use_repo:
+        description: 'repo owner and name'
         required: false
         default: 'redislabs/redisraft'
-      redisraft_branch:
-        description: 'redisraft git branch or sha to use'
+      use_git_ref:
+        description: 'git branch or sha to use'
         required: false
         default: 'master'
 
 jobs:
   redis-suite:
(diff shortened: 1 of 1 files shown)


.github/workflows/redis-suite.yml
```diff
@@ -4,21 +4,51 @@ on:
...
+    - name: prep
+      if: github.event_name == 'workflow_dispatch'
+      run: |
+        echo "REDIS_REPOSITORY=${{github.event.inputs.redis_repo}}" >> $GITHUB_ENV
+        echo "REDIS_BRANCH=${{github.event.inputs.redis_branch}}" >> $GITHUB_ENV
+        echo "REDISRAFT_REPOSITORY=${{github.event.inputs.redisraft_repo}}" >> $GITHUB_ENV
+        echo "REDISRAFT_BRANCH=${{github.event.inputs.redisraft_branch}}" >> $GITHUB_ENV
     - uses: actions/checkout@v2
+      with:
+        repository: ${{ env.REDISRAFT_REPOSITORY }}
```
yossigo, line 41: @tezc Did you test this as a scheduled workflow? I'd expect these variables have to be defined (to hard coded value) in that case, but maybe I'm missing something.

tezc: @yossigo You're right, I didn't try with scheduled workflow. It should be fixed right now.

yossigo (commented):
@tezc I actually think the first version was good, having the ability to configure everything.

I believe the only thing that was missing was an `env:` section to set up the environment variables to the defaults even when the trigger is not workflow_dispatch. This is because inputs on workflow dispatch are simply not available on scheduled jobs.

tezc:
@yossigo Yeah, needed default values. Then, I thought maybe we won't need to configure Redis branch and commit any time soon, so, switched back to this. Looks like you liked that one better, I'll update the PR back to first version.

tezc:
@yossigo Updated the PR, made it configurable back again.

yossigo (approved):

PR #208, opened 2022-01-02
Avoid expensive fseek() call for the index file.

tezc:

PR #236, opened 2022-01-28
Fixed index file bug which becomes visible in BSDs

tezc:
Fixed index file bug and added MacOS to daily CI. 
Also, in tests, fixed a few issues that rely on timing to stabilize them in MacOS build.

The problem is while using FILE API, switching between fread to fwrite or vice versa 
requires calling file positioning functions, e.g fseek() according to C standard.
Example : 
```c
fwrite("abc", 4, 1, file); // Written 3 bytes, offset is 3. 
fseek(file, 0, SEEK_END);  // Set position to zero.
fread(buf, 1, 1, file);    // Read 1 byte, is position 1 ? 
fwrite(buf, "aa", 2, 1);   // <---- Undefined behavior, needs data positioning functions to be called before
```

The problem is implementations might be holding different position variables for read and write. 
Looks like this is not the case for glibc and there is no issue in Linux. 
On BSDs and MacOS, this issue is reproducible.

Related section from C standard : 

> When a file is opened with update mode ( '+' as the second or third character in the mode argument),
both input and output may be performed on the associated stream. However, the application shall 
ensure that output is not directly followed by input without an intervening call to fflush() or to a file
positioning function ( fseek(), fsetpos(), or rewind()), and input is not directly followed by output without 
an intervening call to a file positioning function, unless the input operation encounters end-of-file.

In RedisRaft, issue becomes visible when we read an index just to delete an entry and 
later when we try to write again for a new entry index. `test_log_delete` covers this case,
 not adding another test.

Introduced by https://github.com/RedisLabs/redisraft/pull/208

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -8,7 +8,7 @@ on:
       skipjobs:
         description: 'jobs to skip (delete the ones you want to keep, do not leave empty)'
         required: false
-        default: 'address,undefined'
+        default: 'address,undefined,macos'
       use_repo:
         description: 'repo owner and name'
         required: false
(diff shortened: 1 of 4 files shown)


yossigo (approved):
@tezc This is another argument for using POSIX I/O instead. Do you think it's worth switching?

tezc:
@yossigo Yes, it makes sense even just to have more control when to do buffering etc. Probably not a priority but we should do it sometime. This bug was quite scary, such a surprise. So, maybe we should accelerate :)

PR #247, opened 2022-02-07
add internode auth to redisraft

### LINKED ISSUE #238: Add some way to provide a client password
RedisRaft should use the redis AUTH mechanism when connecting to other redis instances (e.g. follower to leader). 

This means there has to be some way to pass the  password to RedisRaft (in the clear, not hashed or encrypted), and it should use this password when it needs to connect the other redis instances.

sjpotter:
currently, there is no authentication that is done between RedisRaft clusters

this adds the ability to define which acl user and what password should be used, as has redis authenticate them

the default user (if not specified) is default, and the default password is empty.

when used, on must ensure that redis itself is configured to accept such user/password authenticaiton.
If using the default user of default, one can just pass a --requirepass to redis, if one is changing the user, one must setup the acl user correctly in redis conf/acl files

Fixes #238

### DIFF
--- a/src/connection.c
+++ b/src/connection.c
@@ -403,7 +403,7 @@ void ConnMarkDisconnected(Connection *conn)
 
 bool ConnIsIdle(Connection *conn)
 {
-    return (conn->state == CONN_DISCONNECTED || conn->state == CONN_CONNECT_ERROR || (conn->flags | CONN_TERMINATING));
+    return (conn->state == CONN_DISCONNECTED || conn->state == CONN_CONNECT_ERROR || (conn->flags & CONN_TERMINATING));
 }
 
 bool ConnIsConnected(Connection *conn)
--- a/src/connection.c
+++ b/src/connection.c
@@ -185,22 +185,10 @@ static void handleAuth(redisAsyncContext *c, void *r, void *privdata)
         redisAsyncDisconnect(ConnGetRedisCtx(conn));
         ConnMarkDisconnected(conn);
     } else {
-        conn->state = CONN_CONNECTED;
-        conn->connect_oks++;
-        conn->last_connected_time = RedisModule_Milliseconds();
+        return connectionSuccess(conn);
     }
 
-    /* If connection was flagged for termination between connection attempt
-     * and now, we don't call the connect callback.
-     */
-    /* If we're terminating, abort now */
-    if (conn->flags & CONN_TERMINATING) {
-        return;
-    }
-
-    if (conn->connect_callback) {
-        conn->connect_callback(conn);
-    }
+    return connectionFailure(conn);
 }
 
 /* Connection with authentication
@@ -213,6 +201,11 @@ static void handleAuth(redisAsyncContext *c, void *r, void *privdata)
 static void handleConnectedWithAuth(Connection *conn, int status)
 {
     if (status == REDIS_OK) {
+        /* don't try to authenticate if terminating */
+        if (conn->flags & CONN_TERMINATING) {
+            return connectionSuccess(conn);
+        }
+
         if (redisAsyncCommand(ConnGetRedisCtx(conn), handleAuth, conn,
                               "AUTH %s %s",
                               conn->rr->config->cluster_user,
--- a/src/connection.c
+++ b/src/connection.c
@@ -185,10 +185,11 @@ static void handleAuth(redisAsyncContext *c, void *r, void *privdata)
         redisAsyncDisconnect(ConnGetRedisCtx(conn));
         ConnMarkDisconnected(conn);
     } else {
-        return connectionSuccess(conn);
+        connectionSuccess(conn);
+        return;
     }
 
-    return connectionFailure(conn);
+    connectionFailure(conn);
 }
 
 /* Connection with authentication
@@ -197,7 +198,6 @@ static void handleAuth(redisAsyncContext *c, void *r, void *privdata)
  * If connection status is not ok, will immediately transition connection to CONN_CONNECT_ERROR
  * User callback is called in all cases except when the connection has already been terminated externally
  */
-
 static void handleConnectedWithAuth(Connection *conn, int status)
 {
     if (status == REDIS_OK) {
@@ -227,12 +227,12 @@ fail:
  * Depending on status, state transitions to CONN_CONNECTED or CONN_CONNECT_ERROR.
  * User callback is called in all cases except when the connection has already been terminated externally
  */
-
-static void handleConnectedWithoutAuth(Connection *conn, int status) {
+static void handleConnectedWithoutAuth(Connection *conn, int status) 
+{
     if (status == REDIS_OK) {
-        return connectionSuccess(conn);
+        connectionSuccess(conn);
     } else {
-        return connectionFailure(conn);
+        connectionFailure(conn);
     }
 }
 
@@ -242,10 +242,9 @@ static void handleConnectedWithoutAuth(Connection *conn, int status) {
  * from handleResolved(), with state CONN_CONNECTING.
  *
  */
-
 static void handleConnected(const redisAsyncContext *c, int status)
 {
-    Connection *conn = (Connection *) c->data;
+    Connection *conn = c->data;
     CONN_TRACE(conn, "handleConnected: status=%d", status);
 
     if (conn->rr->config->cluster_user) {
--- a/src/connection.c
+++ b/src/connection.c
@@ -14,6 +14,7 @@
 #include <assert.h>
 #include <netdb.h>
 #include <arpa/inet.h>
+#include <string.h>
 
 #define CONN_LOG(level, conn, fmt, ...) \
     LOG(level, "{conn:%lu} " fmt, conn ? conn->id : 0, ##__VA_ARGS__)
@@ -126,13 +127,60 @@ void ConnAsyncTerminate(Connection *conn)
     conn->flags |= CONN_TERMINATING;
 }
 
+static void handleAuth(redisAsyncContext *c, void *r, void *privdata)
+{
+    Connection *conn = privdata;
+    JoinLinkState *state = ConnGetPrivateData(conn);
+
+    redisReply *reply = r;
+
+    if (!reply) {
+        LOG_ERROR("Redis connection authentication failed: connection died");
+        ConnMarkDisconnected(conn);
+    } else if (reply->type == REDIS_REPLY_ERROR) {
+        LOG_ERROR("Redis connection authentication failed: %s", reply->str);
+        state->failed = true;
+        ConnMarkDisconnected(conn);
+    } else if (reply->type != REDIS_REPLY_STATUS || strcmp("OK", reply->str) != 0) {
+        LOG_ERROR("Redis connection authentication failed: unexpected response");
+        state->failed = true;
+        ConnMarkDisconnected(conn);
+    }
+
+    if (conn->connect_callback) {
+        conn->connect_callback(conn);
+    }
+}
+
+static void doAuthentication(Connection *conn, int status)
+{
+    RedisRaftCtx *rr = ConnGetRedisRaftCtx(conn);
+
+    if (status == REDIS_OK) {
+        if (redisAsyncCommand(ConnGetRedisCtx(conn), handleAuth, conn,
+                              "AUTH %s %s",
+                              rr->config->cluster_user ? rr->config->cluster_user : "default",
+                              rr->config->cluster_password ? rr->config->cluster_password : "") != REDIS_OK) {
+            redisAsyncDisconnect(ConnGetRedisCtx(conn));
+            ConnMarkDisconnected(conn);
+        }
+    } else {
+        /* Call callback even when failed */
+        if (conn->connect_callback) {
+            conn->connect_callback(conn);
+        }
+
+    }
+}
+
 /* Connect callback (hiredis).
  *
  * This callback will always be called as a result of a prior redisAsyncConnect()
  * from handleResolved(), with state CONN_CONNECTING.
  *
  * Depending on status, state transitions to CONN_CONNECTED or CONN_CONNECT_ERROR.
- * User callback is called in both cases.
+ * If authentication is required, that will be attempted.
+ * User callback is called in all cases except when the connection has already been terminated externally
  */
 
 static void handleConnected(const redisAsyncContext *c, int status)
@@ -159,10 +207,7 @@ static void handleConnected(const redisAsyncContext *c, int status)
         return;
     }
 
-    /* Call explicit connect callback (even if failed) */
-    if (conn->connect_callback) {
-        conn->connect_callback(conn);
-    }
+    doAuthentication(conn, status);
 }
 
 /* Disconnect callback (hiredis).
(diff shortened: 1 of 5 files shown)


config.c
```diff
@@ -266,6 +268,16 @@ static RRStatus processConfigParam(const char *keyword, const char *value, Redis
         if (*errptr != '\0' || val < 0)
             goto invalid_value;
         target->max_appendentries_inflight = (int) val;
+    } else if (!strcmp(keyword, CONF_CLUSTER_PASSWORD)) {
+        if (target->cluster_password) {
+            RedisModule_Free(target->cluster_password);
+        }
+        target->cluster_password = RedisModule_Strdup(value);
+    } else if (!strcmp(keyword, CONF_CLUSTER_USER)) {
+        if (target->cluster_password) {
```
tezc, line 277:
```suggestion
        if (target->cluster_user) {
```

config.c
```diff
@@ -443,6 +455,14 @@ void handleConfigGet(RedisModuleCtx *ctx, RedisRaftConfig *config, RedisModuleSt
         len++;
         replyConfigInt(ctx, CONF_MAX_APPEND_REQ_IN_FLIGHT, config->max_appendentries_inflight);
     }
+    if (stringmatch(pattern, CONF_CLUSTER_USER, 1)) {
+        len++;
+        replyConfigStr(ctx, CONF_CLUSTER_USER, config->cluster_user);
+    }
+    if (stringmatch(pattern, CONF_CLUSTER_USER, 1)) {
```
tezc, line 462:
```suggestion
    if (stringmatch(pattern, CONF_CLUSTER_PASSWORD, 1)) {
```

connection.c
```diff
@@ -127,13 +128,63 @@ void ConnAsyncTerminate(Connection *conn)
...
+}
+
+static void doAuthentication(Connection *conn, int status)
+{
+    RedisRaftCtx *rr = ConnGetRedisRaftCtx(conn);
+
+    if (status == REDIS_OK) {
+        if (redisAsyncCommand(ConnGetRedisCtx(conn), handleAuth, conn,
+                              "AUTH %s %s",
+                              rr->config->cluster_user ? rr->config->cluster_user : "default",
```
tezc, line 166: If we always have username password,  you can just init in ConfigInit() like :
```c
config->cluster_username = RedisModuleStrdup("default");
config->cluster_password = RedisModuleStrdup("");
```

So, we'll avoid null checks here and above.

connection.c
```diff
@@ -127,13 +128,63 @@ void ConnAsyncTerminate(Connection *conn)
...
+        LOG_ERROR("Redis connection authentication failed: connection died");
+        ConnMarkDisconnected(conn);
+    } else if (reply->type == REDIS_REPLY_ERROR) {
+        LOG_ERROR("Redis connection authentication failed: %s", reply->str);
+        state->failed = true;
+        ConnMarkDisconnected(conn);
+    } else if (reply->type != REDIS_REPLY_STATUS || strcmp("OK", reply->str) != 0) {
+        LOG_ERROR("Redis connection authentication failed: unexpected response");
+        state->failed = true;
+        ConnMarkDisconnected(conn);
```
tezc, line 147: Marking here as disconnected, just down below setting it back to connected? Missing return? Do we still need to call connect_callback though? Similar issue in the above if checks as well.

connection.c
```diff
@@ -127,13 +128,63 @@ void ConnAsyncTerminate(Connection *conn)
...
+    Connection *conn = privdata;
+    JoinLinkState *state = ConnGetPrivateData(conn);
+
+    redisReply *reply = r;
+    if (!reply) {
+        LOG_ERROR("Redis connection authentication failed: connection died");
+        ConnMarkDisconnected(conn);
+    } else if (reply->type == REDIS_REPLY_ERROR) {
+        LOG_ERROR("Redis connection authentication failed: %s", reply->str);
+        state->failed = true;
```
tezc, line 142: Once we set failed, will it stay failed forever? Not just the usage here but in general, about marking a connection failed, should it work again once we fix the issue? For this password issue, after we change the password locally or remotely, should it work? Otherwise, admin must restart the node? No?

Another issue is logging, say I misconfigured the node, I'll take a look at the node an hour later, will I see a big log file with same lines over an over? Probably, with `unrecoverable error, check logs` lines which is printed in `joinLinkIdleCallback()`. (I don't know, I didn't try).

Probably, these are not important issues, maybe I'm wrong. Just saying, maybe we should take a look at Redis behavior and mimic it for these.

connection.c
sjpotter: so I actually did this originally, this was the error I was mentioning in slack about leak detector going of, hence why did this.  But I agree your would be better, will try and put it back and see what happens

connection.c
sjpotter: if it fails auth, we're in a bad state.  but yes, need to think about how it can recover from that.

sjpotter: so dug into this.  the failed boolean is only used by join/link.  it is not used by normal reconnection (for already linked nodes).  so I think its fine to leave them there, if the connection fails, join/link will fail and return to the user and this connection object will be killed.  If already connected, it will be ignored and will try to reconnect.

with that said, want to think about it a bit more to see if we can improve connection handling in general.

connection.c
```diff
@@ -160,10 +204,7 @@ static void handleConnected(const redisAsyncContext *c, int status)
         return;
     }
 
-    /* Call explicit connect callback (even if failed) */
-    if (conn->connect_callback) {
-        conn->connect_callback(conn);
-    }
+    doAuthentication(conn, status);
```
yossigo, line 207: Shouldn't this be conditional, only if authentication is configured?

config.c
```diff
@@ -443,6 +455,14 @@ void handleConfigGet(RedisModuleCtx *ctx, RedisRaftConfig *config, RedisModuleSt
...
         replyConfigInt(ctx, CONF_MAX_APPEND_REQ_IN_FLIGHT, config->max_appendentries_inflight);
     }
+    if (stringmatch(pattern, CONF_CLUSTER_USER, 1)) {
+        len++;
+        replyConfigStr(ctx, CONF_CLUSTER_USER, config->cluster_user);
+    }
+    if (stringmatch(pattern, CONF_CLUSTER_PASSWORD, 1)) {
+        len++;
+        replyConfigStr(ctx, CONF_CLUSTER_PASSWORD, "***");
+    }
```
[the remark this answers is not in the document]
sjpotter: sure, this was more cause I didn't know exactly what to do, but am ok with removing it.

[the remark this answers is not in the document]
yossigo: I think it is better, because it makes no assumptions about what the default user is.

connection.c
sjpotter: ok, its removed, but don't think it really mattered in this context

connection.c
sjpotter: as noted, failed is gone, still not 100% believing that I handle the connection correctly, but think its correct.

connection.c
sjpotter: ok done

connection.c
```diff
@@ -160,12 +199,70 @@ static void handleConnected(const redisAsyncContext *c, int status)
...
+/* Connection with authentication
+ *
+ * If the connection status is ok, will attempt to authenticate and only then call the connection callback
+ * If connection status is not ok, will immediately transition connection to CONN_CONNECT_ERROR
+ * User callback is called in all cases except when the connection has already been terminated externally
+ */
+
+static void handleConnectedWithAuth(Connection *conn, int status)
+{
+    if (status == REDIS_OK) {
```
yossigo, line 216: If we have CONN_TERMINATING, we probably want to avoid an authentication attempt no?

connection.c
```diff
@@ -127,29 +128,67 @@ void ConnAsyncTerminate(Connection *conn)
...
+    } else if (reply->type == REDIS_REPLY_ERROR) {
+        LOG_ERROR("Redis connection authentication failed: %s", reply->str);
+        redisAsyncDisconnect(ConnGetRedisCtx(conn));
+        ConnMarkDisconnected(conn);
+    } else if (reply->type != REDIS_REPLY_STATUS || strcmp("OK", reply->str) != 0) {
+        LOG_ERROR("Redis connection authentication failed: unexpected response");
+        redisAsyncDisconnect(ConnGetRedisCtx(conn));
+        ConnMarkDisconnected(conn);
+    } else {
         conn->state = CONN_CONNECTED;
```
yossigo, line 189: Isn't this back to connectionSuccess()?

yossigo (changes requested):

connection.c
sjpotter: sort of, unsure about asthetics, the issue is that it falls through to after the else to shared code for both success/failure.  pondering

src/connection.c
```diff
@@ -160,12 +163,99 @@ static void handleConnected(const redisAsyncContext *c, int status)
...
+    }
+}
+
+/* Connect callback (hiredis).
+ *
+ * This callback will always be called as a result of a prior redisAsyncConnect()
+ * from handleResolved(), with state CONN_CONNECTING.
+ *
+ */
+
```
tezc, line 246:
```suggestion
```

src/connection.c
```diff
@@ -160,12 +163,99 @@ static void handleConnected(const redisAsyncContext *c, int status)
...
+static void handleConnected(const redisAsyncContext *c, int status)
+{
+    Connection *conn = (Connection *) c->data;
```
tezc, line 249:
```suggestion
    Connection *conn = c->data;
```

src/connection.c
```diff
@@ -160,12 +163,99 @@ static void handleConnected(const redisAsyncContext *c, int status)
...
+ *
+ * Depending on status, state transitions to CONN_CONNECTED or CONN_CONNECT_ERROR.
+ * User callback is called in all cases except when the connection has already been terminated externally
+ */
+
+static void handleConnectedWithoutAuth(Connection *conn, int status) {
+    if (status == REDIS_OK) {
+        return connectionSuccess(conn);
+    } else {
+        return connectionFailure(conn);
```
tezc, line 236:
```suggestion
        connectionSuccess(conn);
    } else {
        connectionFailure(conn);
```

return with a void function is illegal according C as far as I know. We need to add -pedantic -Wextra warnings to compiler sometime(after cleaning up all warnings in the codebase), probably these options would warn us. 
```suggestion
        connectionSuccess(conn);
    } else {
        connectionFailure(conn);
```

src/connection.c
```diff
@@ -160,12 +163,99 @@ static void handleConnected(const redisAsyncContext *c, int status)
...
+fail:
+    connectionFailure(conn);
+}
+
+/* Connection without authentication
+ *
+ * Depending on status, state transitions to CONN_CONNECTED or CONN_CONNECT_ERROR.
+ * User callback is called in all cases except when the connection has already been terminated externally
+ */
+
```
tezc, line 231:
```suggestion
```

src/connection.c
```diff
@@ -160,12 +163,99 @@ static void handleConnected(const redisAsyncContext *c, int status)
...
+ * If connection status is not ok, will immediately transition connection to CONN_CONNECT_ERROR
+ * User callback is called in all cases except when the connection has already been terminated externally
+ */
+
+static void handleConnectedWithAuth(Connection *conn, int status)
+{
+    if (status == REDIS_OK) {
+        /* don't try to authenticate if terminating */
+        if (conn->flags & CONN_TERMINATING) {
+            return connectionSuccess(conn);
```
tezc, line 207: Why we are calling connectionSuccess(); here? Shouldn't we just return?

src/connection.c
```diff
@@ -160,12 +163,99 @@ static void handleConnected(const redisAsyncContext *c, int status)
...
+        ConnMarkDisconnected(conn);
+    } else if (reply->type != REDIS_REPLY_STATUS || strcmp("OK", reply->str) != 0) {
+        LOG_ERROR("Redis connection authentication failed: unexpected response");
+        redisAsyncDisconnect(ConnGetRedisCtx(conn));
+        ConnMarkDisconnected(conn);
+    } else {
+        return connectionSuccess(conn);
+    }
+
+    return connectionFailure(conn);
```
tezc, line 192:
```suggestion
    connectionFailure(conn);
```

src/connection.c
```diff
@@ -160,12 +163,99 @@ static void handleConnected(const redisAsyncContext *c, int status)
...
+    } else if (reply->type == REDIS_REPLY_ERROR) {
+        LOG_ERROR("Redis connection authentication failed: %s", reply->str);
+        redisAsyncDisconnect(ConnGetRedisCtx(conn));
+        ConnMarkDisconnected(conn);
+    } else if (reply->type != REDIS_REPLY_STATUS || strcmp("OK", reply->str) != 0) {
+        LOG_ERROR("Redis connection authentication failed: unexpected response");
+        redisAsyncDisconnect(ConnGetRedisCtx(conn));
+        ConnMarkDisconnected(conn);
+    } else {
+        return connectionSuccess(conn);
```
tezc, line 189:
```suggestion
        connectionSuccess(conn);
        return;
```

src/connection.c
```diff
@@ -160,12 +163,99 @@ static void handleConnected(const redisAsyncContext *c, int status)
...
+    connectionFailure(conn);
+}
+
+/* Connection without authentication
+ *
+ * Depending on status, state transitions to CONN_CONNECTED or CONN_CONNECT_ERROR.
+ * User callback is called in all cases except when the connection has already been terminated externally
+ */
+
+static void handleConnectedWithoutAuth(Connection *conn, int status) {
```
tezc, line 232:
```suggestion
static void handleConnectedWithoutAuth(Connection *conn, int status) 
{
```

src/connection.c
sjpotter: to set the flags, as if we didn't do auth.  not sure it matters.

tezc: I think calling connectionSuccess() here gives wrong impression as if like it actually connected. Probably, we should just return. 

We still need to fix the issue with CONN_TERMINATING, maybe as part of another PR.

sjpotter: in looking at the code, it seems the fix is that ConnIsIdle has to be expanded for checking the terminating flag.

yossigo: @sjpotter I think we also need to consider the visibility of the connection state. For example, differentiate "connecting" from "authenticating" and "connected".

This also makes me realize we should probably track auth errors.

sjpotter: I think I addressed it all in the latest commit

sjpotter: ok, didn't see yossi's last comment before I updated, will look into addressing that too.

yossigo (approved):

gildaf:
> when used, on must ensure that redis itself is configured to accept such user/password authenticaiton. If using the default user of default, one can just pass a --requirepass to redis, if one is changing the user, one must setup the acl user correctly in redis conf/acl files

@sjpotter just to clarify: 
the client should call  `raft.config set cluster-user <SOME_USER_NAME>` 
and  `raft.config set cluster-password <SOME_PASSWORD>` (and\or provide these arguments to raft on server start)
and it should be done on every node (the info **is not** propagated from leader to followers) . 

right ?

PR #248, opened 2022-02-08
Refactor project structure

tezc:
- Moved source files to `src/ ` : Just to make search(function name, variable etc) in the source files easier. Otherwise, search under project directory includes results from other directories. 
- Deleted makefile.
- Added missing functionalities to CmakeFile, fixed minor compiler warnings.
- Added static library build to hiredis Cmake. (Missing in v1.0.0 GA version).
- Added TLS build option to cmake: `cmake .. -DBUILD_TLS=1`. 
   We define `HAVE_TLS` flag to add compile time checks to source file.
- `PYTEST_OPTS` parameter must be passed to cmake instead of make: 
    `cmake . -DPYTEST_OPTS="--redis-executable=redis/src/redis-server -v"`

Note: 
 - `redisraft.so` will be created under the project directory same as before.
 - Requires a PR to Jepsen repo to use cmake.
    - We are running Jepsen tests with Debian Strech images. Realized that CMake version in that image is older than we support. We can lower minimum supported version of CMake to 3.7.2 with a minor adjustment. Also, need to merge this libraft PR as well : https://github.com/RedisLabs/raft/pull/83  
    
       We can do this part in a follow-up PR.


sjpotter:
while you are doing these changes, can you also have hiredis install itself into build for ssl?  requirement for adding ssl support (ssl headers aren't added, maybe other things).  otherwise I'm fine with it as is, if you would want that all to be in an ssl pr.

Makefile
```diff
@@ -171,5 +182,7 @@ $(BUILDDIR)/.deps_installed:
 	mkdir -p $(BUILDDIR)
 	mkdir -p $(BUILDDIR)/lib
 	mkdir -p $(BUILDDIR)/include
+	mkdir -p $(BUILDDIR)/src
+	mkdir -p $(BUILDDIR)/tests
 	$(MAKE) -C deps PREFIX=$(BUILDDIR)
```
sjpotter, line 187: add USE_SSL=1 to enable ssl

tezc:
Sure, I'll add it.

tezc:
@sjpotter Added a commit for TLS build. TLS build is optional, we pass same parameter as Redis: 

 ` make -j 4 BUILD_TLS=yes`

Can you try it? Let me know if something doesn't work.

tezc:
Added HAVE_TLS flag to use like :

```c
#ifdef HAVE_TLS
      #include "hiredis/hiredis_ssl.h"
#endif
```

sjpotter:
USE_SSL=1 needs to be passed to build of hiredis, otherwise its not ssl enabled (and doesn't includes headers in install)

tezc:
@yossigo  @sjpotter Updated the PR and the top comment. Please take a look.

sjpotter (approved):

PR #250, opened 2022-02-12
Fix unsigned variable less than zero check

tezc:
- Fixed one bug (For loop with unsigned variable less than zero check). 
- Changed variable types to silence linter.


yossigo (approved):
LGTM.
BTW I think there's a ticket to clean up clang-tidy stuff, some of that could point to real issues.

PR #253, opened 2022-02-14
TLS Inter-Node Connectivity

### LINKED ISSUE #134: TLS Support
Currently RedisRaft assumes it runs on top of Redis using a plain-text TCP port. Add support for cross-node communication using TLS.

sjpotter:
Adds support for TLS encrypted connection between nodes in a redisraft cluster.

Adds a RedisRaft module parameter `tls-enabled` that is marked yes if want to require tls encrypted connections between nodes.

leverages the tls configuration of redis-server itself to set the private key/cert/ca cert files that will be used for the client connections it makes to remote nodes

extends the integration test framework to support tls tests, including a python fixture to generate and cleanup certs needs for a tls test, as well as a test that run the cluster in tls mode.

Fixes #134


.github/workflows/ci.yml
```diff
@@ -33,7 +33,7 @@ jobs:
     - name: Run tests
       run: |
         cd build
-        make coverage
+        PYTEST_OPTS="--redis-executable=redis/src/redis-server -v" make coverage BUILD_TLS=yes
```
tezc, line 36:
```suggestion
        make coverage
```

We pass BUILD_TLS or PYTEST_OPTS to cmake command only. (This is how things are in cmake world, you pass parameters to cmake and it generates makefiles with these options (A bit annoying though)).

.github/workflows/ci.yml
```diff
@@ -70,4 +70,4 @@ jobs:
       - name: Run tests
         run: |
           cd build
-          make tests
+          PYTEST_OPTS="--redis-executable=redis/src/redis-server -v" make tests BUILD_TLS=yes
```
tezc, line 73:
```suggestion
          make tests
```
Same as above.

src/config.c
```diff
@@ -544,6 +559,36 @@ static char *getRedisConfig(RedisModuleCtx *ctx, const char *name)
...
+    config->raft_response_timeout = REDIS_RAFT_DEFAULT_RAFT_RESPONSE_TIMEOUT;
+    config->proxy_response_timeout = REDIS_RAFT_DEFAULT_PROXY_RESPONSE_TIMEOUT;
+    config->raft_log_max_cache_size = REDIS_RAFT_DEFAULT_LOG_MAX_CACHE_SIZE;
+    config->raft_log_max_file_size = REDIS_RAFT_DEFAULT_LOG_MAX_FILE_SIZE;
+    config->raft_log_fsync = true;
+    config->quorum_reads = true;
+    config->sharding = false;
+    config->external_sharding = false;
+    config->slot_config = "0:16383",
+            config->shardgroup_update_interval = REDIS_RAFT_DEFAULT_SHARDGROUP_UPDATE_INTERVAL;
```
tezc, line 584:
```suggestion
    config->slot_config = "0:16383",
    config->shardgroup_update_interval = REDIS_RAFT_DEFAULT_SHARDGROUP_UPDATE_INTERVAL;
```

src/config.c
```diff
@@ -266,6 +270,26 @@ static RRStatus processConfigParam(const char *keyword, const char *value, Redis
...
+            goto invalid_value;
+        target->tls_enabled = val;
+    } else if (!strcmp(keyword, CONF_TLS_CA_CERT)) {
+        if (target->tls_ca_cert) {
+            RedisModule_Free(target->tls_ca_cert);
+        }
+        target->tls_ca_cert = RedisModule_Strdup(value);
+    } else if (!strcmp(keyword, CONF_TLS_CERT)) {
+        if (target->tls_cert) {
+            RedisModule_Free(target->tls_ca_cert);
```
tezc, line 285:
```suggestion
            RedisModule_Free(target->tls_cert);
```

tezc (approved):

deps/hiredis/ssl.c
```diff
@@ -312,6 +312,7 @@ static int redisSSLConnect(redisContext *c, SSL *ssl) {
     }
 
     hi_free(rssl);
+    SSL_free(ssl);
```
yossigo, line 315: Is this a local change or are we pulling a new hiredis here?
If we do, it's probably best as a separate PR that will be merged with a merge commit to avoid issues with subtree down the road.

docs/TLS.md
```diff
@@ -0,0 +1,24 @@
...
+redis-server --tls-port <por> --port 0 --tls-cert-file <file> --tls-key-file <file> --tls-ca-cert-file <file> 
+```
+
+In addition, one has to pass the `tls-enabled yes` option to the redisraft module
+
+```
+--loadmodule ./redisraft.so <other options> tls-enabled yes
+```
+
+RedisRaft is smart enough to figure out which files to use for the local cert/key/ca by querying the underlying Redis server.  However, if a user wants to use different files for the client aspects vs server aspects, one can specify RedisRaft module options as well
```
yossigo, line 18: I think we should be explicit about what Redis configurations we use.

src/config.c
```diff
@@ -544,6 +559,36 @@ static char *getRedisConfig(RedisModuleCtx *ctx, const char *name)
...
+    config->raft_log_fsync = true;
+    config->quorum_reads = true;
+    config->sharding = false;
+    config->external_sharding = false;
+    config->slot_config = "0:16383";
+    config->shardgroup_update_interval = REDIS_RAFT_DEFAULT_SHARDGROUP_UPDATE_INTERVAL;
+    config->max_appendentries_inflight = REDIS_RAFT_DEFAULT_MAX_APPENDENTRIES;
+    config->tls_ca_cert = getRedisConfig(ctx, "tls-ca-cert-file");
+    config->tls_key = getRedisConfig(ctx, "tls-key-file");
+    config->tls_cert = getRedisConfig(ctx, "tls-cert-file");
```
yossigo, line 588: Redis supports `tls-client-cert-file` and `tls-client-key-file` to allow using different certificates on the client side and on the server side. I think this should also be supported.

Also, `tls-key-file-pass` and `tls-client-key-file-pass` should be supported.

src/connection.c
```diff
@@ -230,13 +227,17 @@ static void handleResolved(void *arg)
...
+        goto fail;
+    }
 
-        redisAsyncFree(conn->rc);
-        conn->rc = NULL;
-        return;
+#ifdef HAVE_TLS
+    if (conn->rr->config->tls_enabled) {
+        if (redisInitiateSSLWithContext(&conn->rc->c, conn->rr->ssl) != REDIS_OK) {
+            CONN_LOG_ERROR(conn, "SSL Error!");
```
yossigo, line 236: TLS issues are hard to debug, we need to make sure we provide more information here from OpenSSL.

tests/integration/conftest.py
```diff
@@ -18,6 +20,9 @@ def pytest_addoption(parser):
     parser.addoption(
         '--redis-executable', default='../redis/src/redis-server',
         help='Name of redis-server executable to use.')
+    parser.addoption(
```
yossigo, line 23: Do you see a good reason to pull this into the test? Not feeling so easy about this unnecessary coupling.

Also, generating certificates every time in a temporary directory is unnecessarily slow I think. After all, it's not certificate handling that we're testing here.

sjpotter: we aren't generating it for each test, we generate it once per test suite run.  the only thing we "slow down" is if you don't run a tls test (i.e. running a single test), it still generates it.  can probably avoid that with a "skip tls generation" flag for those cases.

deps/hiredis/ssl.c
sjpotter: it copied in the hiredis fix.

src/config.c
sjpotter: ok, can add those in a future PR.  will create a new ticket

tests/integration/conftest.py
yossigo: My main concern is we're coupling things here in a way that has little value, IMHO. It would seem better that the test configuration could point to a dir / specific files to use and pass as arguments, and that the user will be required to run the script if TLS tests are to be executed.

PR #258, opened 2022-02-24
Refactor RaftReq usage to block clients only when necessary

tezc:
Block / Unblock clients only when necessary. 

1- Clients are blocked if the request cannot be served immediately. 
- Raft operations will block clients as these operations need quorum: 
   Redis commands, ShardGroup operations, config changes (join or rnode removal) 
- Non-quorum reads don't have to block, we can process them immediately.
- Proxy commands will block as this requires sending message to another node.
- Leader transfer will block until operation is completed. 
     
2-  As part of the PR 
- Fixed some bugs  
- Fixed some TODOs
- Fixed some compiler warnings 

Regarding performance:
- Non-quorum reads are done without blocking clients, gives some performance boost.
- If RaftReq exists for the entry, deserialized commands are reused from RaftReq rather than deserializing from the entry while applying it.


src/raft.c
```diff
@@ -210,7 +204,7 @@ static void executeLogEntry(RedisRaftCtx *rr, raft_entry_t *entry, raft_index_t
     rr->snapshot_info.last_applied_term = entry->term;
     rr->snapshot_info.last_applied_idx = entry_idx;
 
-    RaftRedisCommandArrayFree(&entry_cmds);
+    RaftRedisCommandArrayFree(arr);
```
sjpotter, line 207: this is a really nitty nit, but at first glance I thought arr was &rr as we use rr a lot.

tezc: okay, renamed as "array"

sjpotter:
it be interesting to have documentation for when a raftreq is required and when its not.  I didnt get that instinctual understanding reading it through

tezc:
@sjpotter Added some documentation to the PR description and to RaftReq() description inside the code.

src/raft.c
```diff
@@ -1238,334 +1236,271 @@ RRStatus RedisRaftInit(RedisModuleCtx *ctx, RedisRaftCtx *rr, RedisRaftConfig *c
...
+    e = raft_recv_entry(rr->raft, entry, &response);
+    if (e != 0) {
+        raft_entry_release(entry);
+        replyRaftError(ctx, e);
+        return;
+    }
+
+    raft_entry_release(entry);
+
+    /* We don't have to block on add node, it's all through join which blocks itself */
```
yossigo, line 1452: Don't we have to handle this like any consensus operation? If we're generating the reply now, how can we be sure we don't end up unblocking the client and returning it even if not committed?

src/redisraft.c
```diff
@@ -247,26 +236,27 @@ static int cmdRaftRequestVote(RedisModuleCtx *ctx, RedisModuleString **argv, int
...
 
     int i;
     for (i = 0; i < argc - 1; i++) {
         cmd->argv[i] =  argv[i + 1];
-        RedisModule_RetainString(req->ctx, cmd->argv[i]);
+        RedisModule_RetainString(ctx, cmd->argv[i]);
     }
-    handleRedisCommand(&redis_raft, req);
+    handleRedisCommand(&redis_raft, ctx, &cmds);
+    RaftRedisCommandArrayFree(&cmds);
```
yossigo, line 258: Isn't the command array part of the RaftReq at this point and cannot be freed (or at least, the individual strings need to be retained again)?

yossigo (commented):
@tezc Had a quick initial review, and it generally looks good to me. I guess this would be easier to digest in smaller chunks, but it's probably(?) too late now - unless you have a clear idea on how to do that.

I think that this PR itself is also an interim step towards getting rid of many of the `handleXXX()` which don't do much and could be merged into the actual Redis Module command (and maybe split to different files). There's anyway not a lot of abstraction between them and the Redis Module stuff (as they produce the reply, etc.).

src/redisraft.c
tezc: These commands will be moved if needed inside `handleRedisCommand()`. Please see `RaftRedisCommandArrayMove()` calls inside `handleRedisCommand()`. Whether we moved or not, it'll be safe to free this array.

src/raft.c
tezc: This is existing code, I didn't change, just moved. I agree, it makes to block here but probably we should handle as part of another PR.

tezc: Opened an issue: https://github.com/RedisLabs/redisraft/issues/265

tezc:
@yossigo  Let me try to split this PR a bit then. I didn't do second part of refactor not make this PR bigger. As you mentioned, we can move handle...() function into redisraft.c but it'd make diff even bigger. 

I'll try to create smaller PR for this PR. After we merge them, we can move functions around, it'll be easier to review.

PR #259, opened 2022-02-24
Use config fsync option always

tezc:
RaftLog is now using RedisRaftCtx config which is reconfigurable from the command line. Enables to switch fsync option without restart.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -220,21 +220,6 @@ static int writeBegin(FILE *logfile, int length)
     return n;
 }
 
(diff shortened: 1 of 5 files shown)


yossigo (approved):
@tezc I seem to remember it was configurable also in the past, isn't that right?

tezc:
@yossigo It wasn't. We were accepting new config but wouldn't update `RaftLog`'s `fsync` variable. So, online config wasn't working. This change was part of POC of performance PR and forgot to merge this change.

PR #260, opened 2022-02-24
Add TLS test suite to daily jobs

tezc:
Add a test suite with TLS to daily job.
Require pytest option "--tls" to use tls.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     - name: Build
       run: |
         mkdir build && cd build
-        cmake .. -DCMAKE_BUILD_TYPE=Coverage -DPYTEST_OPTS="--redis-executable=redis/src/redis-server -v" -DBUILD_TLS=yes
+        cmake .. -DCMAKE_BUILD_TYPE=Coverage -DPYTEST_OPTS="--redis-executable=redis/src/redis-server -v"
         make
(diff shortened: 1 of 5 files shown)


yossigo (commented):
I think we should create a matrix and run both TLS and non-TLS.

sjpotter:
so, we shouldn't a specific TLS test (like now) but run all tests as TLS or not TLS?

yossigo:
@sjpotter Yes, I think that makes sense. We may have some tests that test specific TLS functionality, and these should be optional and only be selected if we build with TLS support.

tezc:
@sjpotter @yossigo Do we really need to run all test suites also with TLS? Isn't that too much? (Just to be fair and limit our github actions usage if we don't need). 

Let's run all tests( CI and daily jobs) with TLS enabled and add one `sanitizer-non-tls` job on linux? Wdyt?

yossigo:
@tezc For CI I think that can be enough, but I'd still run full non-TLS in the daily as well. It's not such a big test and it represents a big chunk of what real world scenarios would look like.

tezc:
@sjpotter @yossigo

 Changed PR a bit according to recent discussions.
- Added one test suite to daily to run tests with TLS
- pytest will requite "--tls" option to use TLS in tests

yossigo (approved):

PR #264, opened 2022-02-28
add a slot-config of "" (empty string) to indicate no slots are owned

sjpotter:
Allow one to specify `slot-config ""` on the command line to indicate that no slots are owned.

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1237,6 +1237,10 @@ RedisModuleString *generateSlots(RedisModuleCtx *ctx, ShardGroup *sg)
 {
     RedisModuleString *ret;
 
+    if (sg->slot_ranges_num == 0) {
+        return RedisModule_CreateString(ctx, "", 0);
+    }
+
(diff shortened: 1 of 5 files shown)


yossigo (commented):
@sjpotter Is there a reason why you don't simply use an empty string as an indicator of no hash slots?

sjpotter:
I don't know how to pass an empty string to it....

yossigo:
@sjpotter Do you mean, pass an empty string as a module argument?

sjpotter:
correct.  slot-config "" sharding yes I'd assume would just be the same as slot-config sharding yes if done on the command line.  I can imagine doing it via exec with a constructed argv, but dont know how to do it on command line (at least cleanly).

sjpotter:
changed to magic value of empty string instead of -1

jhershberg (approved):

yossigo (approved):

PR #268, opened 2022-02-28
Avoid deserializing commands if request exists

tezc:
Reuse already deserialized command array from the request rather than deserializing it from the entry

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -184,43 +184,27 @@ static void executeLogEntry(RedisRaftCtx *rr, raft_entry_t *entry, raft_index_t 
 {
     assert(entry->type == RAFT_LOGTYPE_NORMAL);
 
-    RaftRedisCommandArray *array;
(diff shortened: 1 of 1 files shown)


src/raft.c
```diff
@@ -210,7 +214,7 @@ static void executeLogEntry(RedisRaftCtx *rr, raft_entry_t *entry, raft_index_t
     rr->snapshot_info.last_applied_term = entry->term;
     rr->snapshot_info.last_applied_idx = entry_idx;
 
-    RaftRedisCommandArrayFree(&entry_cmds);
```
sjpotter, line 213: do we want to free array here if its from the req?

tezc: It was okay to call free even it's from the request, so we don't introduce another if check. 

Anyway, changed code a bit and did branching once, probably it's more clear right now.

sjpotter (approved):

PR #269, opened 2022-02-28
Delete invalid multi test

tezc:
This test relies on handling multi/exec on append time. This behavior does not exist right now and we won't implement it as it breaks consistency.


sjpotter:
i'm a bit confused by how this is a watch test?  it seems to be a drop test.

tezc:
Right, not a watch test. Changed the title.

sjpotter (approved):

PR #272, opened 2022-03-01
Remove leftover check for multi-exec handling

tezc:
Remove leftover check for multi-exec handling.

This check assumes executing multi-exec commands on the leader node only. This approach breaks consistency and will not be implemented.


tezc:
Test failure is unrelated. Created an issue for it: https://github.com/RedisLabs/redisraft/issues/273

yossigo (approved):

PR #275, opened 2022-03-01
enable acl support

sjpotter:
check ACLs using redis module api before we execute redis commmand
allow acl setuser to be serialized to all nodes via raft log


sjpotter:
todo: 
* acl test
* serialize out acls in snapshot
* read in acls during snapshot load

sjpotter:
gone in a different direction

PR #280, opened 2022-03-06
Move leader transfer command handling into the command callback

tezc:
As part of refactoring, moved leader transfer command handling into the command callback.

### DIFF
--- a/src/redisraft.c
+++ b/src/redisraft.c
@@ -139,7 +139,6 @@ static int cmdRaftNode(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
  */
 static int cmdRaftTransferLeader(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
 {
-    char buf[128];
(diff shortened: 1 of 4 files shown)


src/redisraft.c
```diff
@@ -156,29 +162,54 @@ static int cmdRaftTransferLeader(RedisModuleCtx *ctx, RedisModuleString **argv,
         return REDISMODULE_OK;
     }
 
-    RaftReq *req = RaftReqInit(ctx, RR_TRANSFER_LEADER);
-    req->r.node_to_transfer_leader = target_node_id;
+    int err = raft_transfer_leader(rr->raft, target_node_id, 0);
+    if (err != 0) {
+        switch (err) {
+            case RAFT_ERR_NOT_LEADER:
```
yossigo, line 168: Probably a good idea to find a way to consolidate this with `replyRaftError()`.

tezc: Okay, moved these error cases into `replyRaftError()`

yossigo (approved):

PR #282, opened 2022-03-06
Refactor config command handling

tezc:
Refactored and moved config command handling into the command callback.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -324,33 +324,31 @@ invalid_value:
     return RR_ERROR;
 }
 
-void handleConfigSet(RedisRaftCtx *rr, RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
+void ConfigSet(RedisRaftCtx *rr, RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
 {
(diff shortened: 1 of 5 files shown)


sjpotter:
my only thought here is why we need to allocate 2 new strings then free them, why not simply pass in the const char * and len to processConfigParam() and use strncmp and the like.  with that said, its probably makes a much bigger change to processConfigParam() and as 1) not common 2) very easy scoping of the memory lifetime makes it simpler to go this way.

sjpotter:
merge conflict needs a fix, but others I'm ok.

tezc:
@sjpotter It's always easier to use null terminated strings as you said. Generally, I'm not ignoring this option, we can pass pointer and len together if performance is critical. As config changes don't happen a lot, string allocation wouldn't cause a performance issue here. 

It was previously allocating stack space to add null character at the end but I'm not sure if this is a good practice if you don't know the size of the allocation. It can cause stackoverflow. Even if you know the max size, I wouldn't allocate more than a few KBs on stack.

Rebased the PR btw.

sjpotter (approved):

PR #283, opened 2022-03-06
Refactor snapshot command handling

tezc:
Refactored and moved snapshot command handling into the command callback.


sjpotter (approved):
seems fine, it just removes the req as we collapse the handle function into cmdRaftSnapshotg

PR #285, opened 2022-03-08
Refactor shardgroup command handling

tezc:
Refactored and moved shardgroup command handling into the command callback.


src/raft.c
```diff
@@ -25,12 +25,13 @@ const char *RaftReqTypeStr[] = {
...
     "RR_REDISCOMMAND",
     "RR_INFO",
+    "RR_DEBUG",
     "RR_COMPACT",
     "RR_CLIENT_DISCONNECT",
     "RR_SHARDGROUP_ADD",
     "RR_SHARDGROUP_UPDATE",
-    "RR_SHARDGROUP_GET",
     "RR_SHARDGROUP_LINK",
+    "RR_NODE_SHUTDOWN",
```
sjpotter, line 34: this seems weird.  also whwere did RR_SHARDGROUP_REPLACE go?

src/redisraft.c
```diff
@@ -713,62 +713,154 @@ static int cmdRaftCluster(RedisModuleCtx *ctx, RedisModuleString **argv, int arg
...
         return REDISMODULE_OK;
     }
 
     size_t cmd_len;
     const char *cmd = RedisModule_StringPtrLen(argv[1], &cmd_len);
 
     if (!strncasecmp(cmd, "GET", cmd_len)) {
-        req = RaftReqInit(ctx, RR_SHARDGROUP_GET);
-        handleShardGroupGet(&redis_raft, req);
+        ShardGroup *sg = getShardGroupById(rr, rr->log->dbid);
```
sjpotter, line 739: i feel including the whole former "handle" function inline here, makes this function way too big....  why not just call the handle function directly with just a slight change in prototype?

tezc: The reason is there is no abstraction between raft.c and redisraft.c anymore. Having separate functions in raft.c doesn't make much sense. Maybe we should leave raft related functions in raft.c eventually. Also, these generic handle...() family functions pollutes namespace and against coding convention I believe (Public "connection" functions start with Conn, Shard family functions start with Shard... etc.). So, I'm looking to get rid of these handle functions as much as possible. 

Anyway, moved Shardgroup operations into cluster.c. It makes sense to keep sharding related functions together and makes the function shorter/more readable as you suggested I believe.

sjpotter (approved):

PR #286, opened 2022-03-09
Address TLS usability issues

### LINKED ISSUE #257: Improve TLS Support
1) Support monitoring other redis tls variables

https://github.com/RedisLabs/redisraft/pull/253#discussion_r811005405

```
tls-client-cert-file
tls-client-key-file
tls-key-file-pass
tls-client-key-file-pass
```

2) better TLS Debugging

https://github.com/RedisLabs/redisraft/pull/253#discussion_r811006641

3) improve tls documentation on what redis config options we use

https://github.com/RedisLabs/redisraft/pull/253#discussion_r811001869

4) possibly (if can figure out how) only generate tls certs before a tls test.

https://github.com/RedisLabs/redisraft/pull/253#discussion_r811009080

sjpotter:
* update RedisRaft to latest modules for config event subscriptions
* sbscribe to config events to follow tls config changes
* use client variables if they are defined
* ignore redis tls configs if defined via module config
* add key pass config/callback
* add documentation for configuring TLS via redis-server and via module configuration

+ abstracted out redis config updating and hiredis ssl context creating as used in multiple places

todo:
- [x] improve documentation
- [x] hookup key pass configuration with hiredis
- [x] improve debugging

fixes #257

### DIFF
--- a/docs/TLS.md
+++ b/docs/TLS.md
@@ -16,7 +16,7 @@ In addition, one has to pass the `tls-enabled yes` option to the redisraft modul
 
 ## Configuring RedisRaft TLS support
 
-RedisRaft can be configured by reusing the Redis server's TLS configuration, or by configuring the RedisRaft module directly. In general, we expect people will configure the Redis server. 
+RedisRaft is configured by reusing the Redis server's TLS configuration. 
 
 ### Configuring RedisRaft via Redis.
 
@@ -29,7 +29,3 @@ RedisRaft can be configured by reusing the Redis server's TLS configuration, or 
 | tls-client-cert-file     | File Containing the PEM encoded public signed CERT (overrides tls-cert-file)                              |                                                                                                          |
 | tls-key-file-pass        | String containing password to decrypt the private key, if encrypted                                       |
 | tls-client-key-file-pass | String containing password to decrypt the private key if using client key field                           | 
-
-### Configuring RedisRaft
-
-RedisRaft can also be configured via module configuration using the same Redis configuration key names if one wants to keep it separate from Redis.  One thing to note, if one configures RedisRaft via the module configuration, one should ensure that one configures all required fields via the module configuration options.
--- a/docs/TLS.md
+++ b/docs/TLS.md
@@ -16,17 +16,17 @@ In addition, one has to pass the `tls-enabled yes` option to the redisraft modul
 
 ## Configuring RedisRaft TLS support
 
-RedisRaft can be configured by reusing the Redis server's TLS configuration, or by configuring the RedisRaft module directly.  In general, we expect people will configure the Redis server. 
+RedisRaft can be configured by reusing the Redis server's TLS configuration, or by configuring the RedisRaft module directly. In general, we expect people will configure the Redis server. 
 
 ### Configuring RedisRaft via Redis.
 
 | Config Key Name          | Meaning                                                                                                   |
 |--------------------------|-----------------------------------------------------------------------------------------------------------|
-| tls-ca-cert-file         | File Containing the CA Sertificate                                                                        |
+| tls-ca-cert-file         | File Containing the CA Certificate                                                                        |
 | tls-key-file             | File Containing the PEM encoded private key file                                                          |
-| tls-client-key-file      | File Containing the PEM encoded private key file for client connecitons (overrides value of tls-key-file) |
-| tls-cert-file            | File Containg the PEM encoded public signed CERT                                                          |
-| tla-client-cert-file     | File Containg the PEM encoded public signed CERT (overrides tla-cert-file)                                |                                                                                                          |
+| tls-client-key-file      | File Containing the PEM encoded private key file for client connections (overrides value of tls-key-file) |
+| tls-cert-file            | File Containing the PEM encoded public signed CERT                                                        |
+| tls-client-cert-file     | File Containing the PEM encoded public signed CERT (overrides tls-cert-file)                              |                                                                                                          |
 | tls-key-file-pass        | String containing password to decrypt the private key, if encrypted                                       |
 | tls-client-key-file-pass | String containing password to decrypt the private key if using client key field                           | 
 
@@ -34,9 +34,9 @@ RedisRaft can be configured by reusing the Redis server's TLS configuration, or 
 
 | Redis Module Config Key Name | Maps to Redis Configuration |
 |------------------------------|-----------------------------|
-| tls-ca-cert                  | tla-ca-cert-file            |
+| tls-ca-cert                  | tls-ca-cert-file            |
 | tls-key                      | tls-key-file                |
-| tls-cert                     | tla-cert-file               |
-| tla-key-pass                 | tile-key-file-pass          |
+| tls-cert                     | tls-cert-file               |
+| tls-key-pass                 | tls-key-file-pass           |
 
 One thing to note, if one configures RedisRaft via the module configuration, one should ensure that one configures all required fields via the module configuration options.
--- a/docs/TLS.md
+++ b/docs/TLS.md
@@ -1,5 +1,4 @@
-RedisRaft TLS Support
-=====================
+# RedisRaft TLS Support
 
 RedisRaft's TLS support mirrors Redis's native TLS support.  
 
@@ -15,10 +14,29 @@ In addition, one has to pass the `tls-enabled yes` option to the redisraft modul
 --loadmodule ./redisraft.so <other options> tls-enabled yes
 ```
 
-RedisRaft is smart enough to figure out which files to use for the local cert/key/ca by querying the underlying Redis server.  However, if a user wants to use different files for the client aspects vs server aspects, one can specify RedisRaft module options as well
+## Configuring RedisRaft TLS support
 
-```asm
-tls-ca-cert
-tls-cert
-tls-key
-```
+RedisRaft can be configured by reusing the Redis server's TLS configuration, or by configuring the RedisRaft module directly.  In general, we expect people will configure the Redis server. 
+
+### Configuring RedisRaft via Redis.
+
+| Config Key Name          | Meaning                                                                                                   |
+|--------------------------|-----------------------------------------------------------------------------------------------------------|
+| tls-ca-cert-file         | File Containing the CA Sertificate                                                                        |
+| tls-key-file             | File Containing the PEM encoded private key file                                                          |
+| tls-client-key-file      | File Containing the PEM encoded private key file for client connecitons (overrides value of tls-key-file) |
+| tls-cert-file            | File Containg the PEM encoded public signed CERT                                                          |
+| tla-client-cert-file     | File Containg the PEM encoded public signed CERT (overrides tla-cert-file)                                |                                                                                                          |
+| tls-key-file-pass        | String containing password to decrypt the private key, if encrypted                                       |
+| tls-client-key-file-pass | String containing password to decrypt the private key if using client key field                           | 
+
+### Configuring RedisRaft
+
+| Redis Module Config Key Name | Maps to Redis Configuration |
+|------------------------------|-----------------------------|
+| tls-ca-cert                  | tla-ca-cert-file            |
+| tls-key                      | tls-key-file                |
+| tls-cert                     | tla-cert-file               |
+| tla-key-pass                 | tile-key-file-pass          |
+
+One thing to note, if one configures RedisRaft via the module configuration, one should ensure that one configures all required fields via the module configuration options.
--- a/docs/TLS.md
+++ b/docs/TLS.md
@@ -32,11 +32,4 @@ RedisRaft can be configured by reusing the Redis server's TLS configuration, or 
 
 ### Configuring RedisRaft
 
-| Redis Module Config Key Name | Maps to Redis Configuration |
-|------------------------------|-----------------------------|
-| tls-ca-cert                  | tls-ca-cert-file            |
-| tls-key                      | tls-key-file                |
-| tls-cert                     | tls-cert-file               |
-| tls-key-pass                 | tls-key-file-pass           |
-
-One thing to note, if one configures RedisRaft via the module configuration, one should ensure that one configures all required fields via the module configuration options.
+RedisRaft can also be configured via module configuration using the same Redis configuration key names if one wants to keep it separate from Redis.  One thing to note, if one configures RedisRaft via the module configuration, one should ensure that one configures all required fields via the module configuration options.
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1540,7 +1540,12 @@ static void linkHandleResponse(redisAsyncContext *c, void *r, void *privdata)
     JoinLinkState *state = ConnGetPrivateData(conn);
 
     if (!reply) {
+#ifndef HAVE_TLS
         LOG_WARNING("RAFT.SHARDGROUP GET failed: connection dropped.");
+#else
+        unsigned long e = ERR_peek_last_error();
+        LOG_WARNING("RAFT.SHARDGROUP GET failed: connection dropped: %s", ERR_reason_error_string(e));
+#endif
     } else if (reply->type == REDIS_REPLY_ERROR) {
         /* -MOVED? */
         if (strlen(reply->str) > 6 && !strncmp(reply->str, "MOVED ", 6)) {
--- a/src/config.c
+++ b/src/config.c
@@ -51,7 +51,7 @@ static const char *CONF_TLS_ENABLED = "tls-enabled";
 static const char *CONF_TLS_CA_CERT = "tls-ca-cert";
 static const char *CONF_TLS_CERT = "tls-cert";
 static const char *CONF_TLS_KEY = "tls-key";
-static const char *CONF_TLS_KEY_FILE_PASS = "tls-key-file-pass";
+static const char *CONF_TLS_KEY_PASS = "tls-key-pass";
 static const char *CONF_CLUSTER_USER = "cluster-user";
 static const char *CONF_CLUSTER_PASSWORD = "cluster-password";
 
@@ -428,12 +428,12 @@ static RRStatus processConfigParam(const char *keyword, const char *value, Redis
             RedisModule_Free(target->tls_key);
         }
         target->tls_key = RedisModule_Strdup(value);
-    } else if (!strcmp(keyword, CONF_TLS_KEY_FILE_PASS)) {
+    } else if (!strcmp(keyword, CONF_TLS_KEY_PASS)) {
         target->tls_manual = true;
-        if (target->tls_key_file_pass) {
-            RedisModule_Free(target->tls_key_file_pass);
+        if (target->tls_key_pass) {
+            RedisModule_Free(target->tls_key_pass);
         }
-        target->tls_key_file_pass = RedisModule_Strdup(value);
+        target->tls_key_pass = RedisModule_Strdup(value);
     } else if (!strcmp(keyword, CONF_CLUSTER_PASSWORD)) {
         if (target->cluster_password) {
             RedisModule_Free(target->cluster_password);
@@ -663,11 +663,11 @@ void updateTLSConfig(RedisModuleCtx *ctx, RedisRaftConfig *config) {
         RedisModule_Free(config->tls_key);
         config->tls_key = key;
     }
-    config->tls_key_file_pass = getRedisConfig(ctx, "tls-key-file-pass");
+    config->tls_key_pass = getRedisConfig(ctx, "tls-key-file-pass");
     char *pass = getRedisConfig(ctx, "tls-client-key-file-pass");
     if (pass) {
-        RedisModule_Free(config->tls_key_file_pass);
-        config->tls_key_file_pass = pass;
+        RedisModule_Free(config->tls_key_pass);
+        config->tls_key_pass = pass;
     }
     config->tls_cert = getRedisConfig(ctx, "tls-cert-file");
     char *cert = getRedisConfig(ctx, "tls-client-cert-file");
--- a/src/config.c
+++ b/src/config.c
@@ -48,6 +48,7 @@ static const char *CONF_IGNORED_COMMANDS = "ignored-commands";
 static const char *CONF_EXTERNAL_SHARDING = "external-sharding";
 static const char *CONF_MAX_APPEND_REQ_IN_FLIGHT = "max-append-req-in-flight";
 static const char *CONF_TLS_ENABLED = "tls-enabled";
+static const char *CONF_TLS_TRACE = "tls-trace";
 static const char *CONF_TLS_CA_CERT = "tls-ca-cert";
 static const char *CONF_TLS_CERT = "tls-cert";
 static const char *CONF_TLS_KEY = "tls-key";
@@ -410,6 +411,11 @@ static RRStatus processConfigParam(const char *keyword, const char *value, Redis
         if (parseBool(value, &val) != RR_OK)
             goto invalid_value;
         target->tls_enabled = val;
+    } else if (!strcmp(keyword, CONF_TLS_TRACE)) {
+        bool val;
+        if (parseBool(value, &val) != RR_OK)
+            goto invalid_value;
+        target->tls_trace = val;
     } else if (!strcmp(keyword, CONF_TLS_CA_CERT)) {
         target->tls_manual = true;
         if (target->tls_ca_cert) {
@@ -660,20 +666,32 @@ void updateTLSConfig(RedisModuleCtx *ctx, RedisRaftConfig *config) {
     config->tls_key = getRedisConfig(ctx, "tls-key-file");
     char *key = getRedisConfig(ctx, "tls-client-key-file");
     if (key) {
-        RedisModule_Free(config->tls_key);
-        config->tls_key = key;
+        if (strcmp("", key)) {
+            RedisModule_Free(config->tls_key);
+            config->tls_key = key;
+        } else {
+            RedisModule_Free(key);
+        }
     }
     config->tls_key_pass = getRedisConfig(ctx, "tls-key-file-pass");
     char *pass = getRedisConfig(ctx, "tls-client-key-file-pass");
     if (pass) {
-        RedisModule_Free(config->tls_key_pass);
-        config->tls_key_pass = pass;
+        if (strcmp("", pass)) {
+            RedisModule_Free(config->tls_key_pass);
+            config->tls_key_pass = pass;
+        } else {
+            RedisModule_Free(pass);
+        }
     }
     config->tls_cert = getRedisConfig(ctx, "tls-cert-file");
     char *cert = getRedisConfig(ctx, "tls-client-cert-file");
     if (cert) {
-        RedisModule_Free(config->tls_cert);
-        config->tls_cert = cert;
+        if (strcmp("", cert)) {
+            RedisModule_Free(config->tls_cert);
+            config->tls_cert = cert;
+        } else {
+            RedisModule_Free(cert);
+        }
     }
 
 }
--- a/src/config.c
+++ b/src/config.c
@@ -48,7 +48,6 @@ static const char *CONF_IGNORED_COMMANDS = "ignored-commands";
 static const char *CONF_EXTERNAL_SHARDING = "external-sharding";
 static const char *CONF_MAX_APPEND_REQ_IN_FLIGHT = "max-append-req-in-flight";
 static const char *CONF_TLS_ENABLED = "tls-enabled";
-static const char *CONF_TLS_TRACE = "tls-trace";
 static const char *CONF_TLS_CA_CERT = "tls-ca-cert";
 static const char *CONF_TLS_CERT = "tls-cert";
 static const char *CONF_TLS_KEY = "tls-key";
@@ -411,11 +410,6 @@ static RRStatus processConfigParam(const char *keyword, const char *value, Redis
         if (parseBool(value, &val) != RR_OK)
             goto invalid_value;
         target->tls_enabled = val;
-    } else if (!strcmp(keyword, CONF_TLS_TRACE)) {
-        bool val;
-        if (parseBool(value, &val) != RR_OK)
-            goto invalid_value;
-        target->tls_trace = val;
     } else if (!strcmp(keyword, CONF_TLS_CA_CERT)) {
         target->tls_manual = true;
         if (target->tls_ca_cert) {
@@ -661,8 +655,15 @@ void ConfigGet(RedisRaftCtx *rr, RedisModuleCtx *ctx, RedisModuleString **argv, 
     RedisModule_ReplySetArrayLength(ctx, len * 2);
 }
 
-void updateTLSConfig(RedisModuleCtx *ctx, RedisRaftConfig *config) {
+void updateTLSConfig(RedisModuleCtx *ctx, RedisRaftConfig *config)
+{
+    if (config->tls_ca_cert) {
+        RedisModule_Free(config->tls_ca_cert);
+    }
     config->tls_ca_cert = getRedisConfig(ctx, "tls-ca-cert-file");
+    if (config->tls_key) {
+        RedisModule_Free(config->tls_key);
+    }
     config->tls_key = getRedisConfig(ctx, "tls-key-file");
     char *key = getRedisConfig(ctx, "tls-client-key-file");
     if (key) {
@@ -672,6 +673,9 @@ void updateTLSConfig(RedisModuleCtx *ctx, RedisRaftConfig *config) {
         } else {
             RedisModule_Free(key);
         }
+    }
+    if (config->tls_key_pass) {
+        RedisModule_Free(config->tls_key_pass);
     }
     config->tls_key_pass = getRedisConfig(ctx, "tls-key-file-pass");
     char *pass = getRedisConfig(ctx, "tls-client-key-file-pass");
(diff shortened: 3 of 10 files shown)


sjpotter:
so this is what errors look like now with mismatched CA

on the client (as before)

```
(error) ERR failed to join cluster, please check logs
```

in the logs of the server

```
<redisraft> RAFT.NODE ADD failed: connection dropped: certificate verify failed
```

on the leader (that we attempted to join) we see

```
Error accepting a client connection: error:1409441B:SSL routines:ssl3_read_bytes:tlsv1 alert decrypt error  (this error is generated by redis)
```

docs/TLS.md
```diff
@@ -15,10 +14,29 @@ In addition, one has to pass the `tls-enabled yes` option to the redisraft modul
...
+RedisRaft can be configured by reusing the Redis server's TLS configuration, or by configuring the RedisRaft module directly.  In general, we expect people will configure the Redis server. 
+
+### Configuring RedisRaft via Redis.
+
+| Config Key Name          | Meaning                                                                                                   |
+|--------------------------|-----------------------------------------------------------------------------------------------------------|
+| tls-ca-cert-file         | File Containing the CA Sertificate                                                                        |
+| tls-key-file             | File Containing the PEM encoded private key file                                                          |
+| tls-client-key-file      | File Containing the PEM encoded private key file for client connecitons (overrides value of tls-key-file) |
+| tls-cert-file            | File Containg the PEM encoded public signed CERT                                                          |
```
tezc, line 28:
```suggestion
| tls-cert-file            | File Containing the PEM encoded public signed CERT                                                          |
```

docs/TLS.md
```diff
@@ -15,10 +14,29 @@ In addition, one has to pass the `tls-enabled yes` option to the redisraft modul
...
+| tla-client-cert-file     | File Containg the PEM encoded public signed CERT (overrides tla-cert-file)                                |                                                                                                          |
```
tezc, line 29:
```suggestion
| tls-client-cert-file     | File Containing the PEM encoded public signed CERT (overrides tla-cert-file)                                |                                                                                                          |
```

docs/TLS.md
```diff
@@ -15,10 +14,29 @@ In addition, one has to pass the `tls-enabled yes` option to the redisraft modul
...
-RedisRaft is smart enough to figure out which files to use for the local cert/key/ca by querying the underlying Redis server.  However, if a user wants to use different files for the client aspects vs server aspects, one can specify RedisRaft module options as well
+## Configuring RedisRaft TLS support
 
-```asm
-tls-ca-cert
-tls-cert
-tls-key
-```
\ No newline at end of file
+RedisRaft can be configured by reusing the Redis server's TLS configuration, or by configuring the RedisRaft module directly.  In general, we expect people will configure the Redis server.
```
tezc, line 19:
```suggestion
RedisRaft can be configured by reusing the Redis server's TLS configuration, or by configuring the RedisRaft module directly. In general, we expect people will configure the Redis server. 
```

docs/TLS.md
```diff
@@ -15,10 +14,29 @@ In addition, one has to pass the `tls-enabled yes` option to the redisraft modul
...
+| tls-cert-file            | File Containg the PEM encoded public signed CERT                                                          |
+| tla-client-cert-file     | File Containg the PEM encoded public signed CERT (overrides tla-cert-file)                                |                                                                                                          |
+| tls-key-file-pass        | String containing password to decrypt the private key, if encrypted                                       |
+| tls-client-key-file-pass | String containing password to decrypt the private key if using client key field                           | 
+
+### Configuring RedisRaft
+
+| Redis Module Config Key Name | Maps to Redis Configuration |
+|------------------------------|-----------------------------|
+| tls-ca-cert                  | tla-ca-cert-file            |
```
tezc, line 37:
```suggestion
| tls-ca-cert                  | tls-ca-cert-file            |
```

docs/TLS.md
```diff
@@ -15,10 +14,29 @@ In addition, one has to pass the `tls-enabled yes` option to the redisraft modul
...
+| tls-key                      | tls-key-file                |
+| tls-cert                     | tla-cert-file               |
+| tla-key-pass                 | tile-key-file-pass          |
```
tezc, line 40:
```suggestion
| tls-key-pass                 | tile-key-file-pass          |
```

docs/TLS.md
```diff
@@ -15,10 +14,29 @@ In addition, one has to pass the `tls-enabled yes` option to the redisraft modul
...
+| tls-key-file-pass        | String containing password to decrypt the private key, if encrypted                                       |
+| tls-client-key-file-pass | String containing password to decrypt the private key if using client key field                           | 
+
+### Configuring RedisRaft
+
+| Redis Module Config Key Name | Maps to Redis Configuration |
+|------------------------------|-----------------------------|
+| tls-ca-cert                  | tla-ca-cert-file            |
+| tls-key                      | tls-key-file                |
+| tls-cert                     | tla-cert-file               |
```
tezc, line 39:
```suggestion
| tls-cert                     | tls-cert-file               |
```

docs/TLS.md
```diff
@@ -15,10 +14,29 @@ In addition, one has to pass the `tls-enabled yes` option to the redisraft modul
...
\ No newline at end of file
+RedisRaft can be configured by reusing the Redis server's TLS configuration, or by configuring the RedisRaft module directly.  In general, we expect people will configure the Redis server. 
+
+### Configuring RedisRaft via Redis.
+
+| Config Key Name          | Meaning                                                                                                   |
+|--------------------------|-----------------------------------------------------------------------------------------------------------|
+| tls-ca-cert-file         | File Containing the CA Sertificate                                                                        |
+| tls-key-file             | File Containing the PEM encoded private key file                                                          |
+| tls-client-key-file      | File Containing the PEM encoded private key file for client connecitons (overrides value of tls-key-file) |
```
tezc, line 27:
```suggestion
| tls-client-key-file      | File Containing the PEM encoded private key file for client connections (overrides value of tls-key-file) |
```

docs/TLS.md
```diff
@@ -15,10 +14,29 @@ In addition, one has to pass the `tls-enabled yes` option to the redisraft modul
...
-tls-key
-```
\ No newline at end of file
+RedisRaft can be configured by reusing the Redis server's TLS configuration, or by configuring the RedisRaft module directly.  In general, we expect people will configure the Redis server. 
+
+### Configuring RedisRaft via Redis.
+
+| Config Key Name          | Meaning                                                                                                   |
+|--------------------------|-----------------------------------------------------------------------------------------------------------|
+| tls-ca-cert-file         | File Containing the CA Sertificate                                                                        |
```
tezc, line 25:
```suggestion
| tls-ca-cert-file         | File Containing the CA Certificate                                                                        |
```

docs/TLS.md
```diff
@@ -15,10 +14,29 @@ In addition, one has to pass the `tls-enabled yes` option to the redisraft modul
...
 
-```asm
-tls-ca-cert
-tls-cert
-tls-key
-```
\ No newline at end of file
+RedisRaft can be configured by reusing the Redis server's TLS configuration, or by configuring the RedisRaft module directly.  In general, we expect people will configure the Redis server. 
+
+### Configuring RedisRaft via Redis.
```
tezc, line 21: Why don't we use Redis config names?

src/config.c
```diff
@@ -281,21 +283,35 @@ static RRStatus processConfigParam(const char *keyword, const char *value, Redis
         if (parseBool(value, &val) != RR_OK)
             goto invalid_value;
         target->tls_enabled = val;
+    } else if (!strcmp(keyword, CONF_TLS_TRACE)) {
```
tezc, line 286: Is this leftover? I don't see where we use it.

sjpotter: yes.

src/config.c
```diff
@@ -584,6 +600,41 @@ static char *getRedisConfig(RedisModuleCtx *ctx, const char *name)
     return buf;
 }
 
+void updateTLSConfig(RedisModuleCtx *ctx, RedisRaftConfig *config) {
```
tezc, line 603:
```suggestion
void updateTLSConfig(RedisModuleCtx *ctx, RedisRaftConfig *config) 
{
```

src/config.c
```diff
@@ -584,6 +600,41 @@ static char *getRedisConfig(RedisModuleCtx *ctx, const char *name)
...
+    config->tls_ca_cert = getRedisConfig(ctx, "tls-ca-cert-file");
+    config->tls_key = getRedisConfig(ctx, "tls-key-file");
+    char *key = getRedisConfig(ctx, "tls-client-key-file");
+    if (key) {
+        if (strcmp("", key)) {
+            RedisModule_Free(config->tls_key);
```
tezc, line 609: We've updated on line 605, freeing here. Similar to problem above.

src/redisraft.c
```diff
@@ -1008,6 +1008,104 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
...
+    if (pass_len > (size_t) size) return -1;
+    memcpy(buf, pass, pass_len);
+
+    return (int) pass_len;
+}
+
+SSL_CTX *generateSSLContext(RedisModuleCtx *ctx, RedisRaftCtx *rr) {
+    SSL_CTX *ssl_ctx = SSL_CTX_new(SSLv23_client_method());
+    if (!ssl_ctx) {
+        RedisModule_Log(ctx, REDIS_WARNING, "REDIS_SSL_CTX_CREATE_FAILED");
```
tezc, line 1030: Let's use LOG_() family macros for logging.  LOG_ERROR() ... etc.

src/redisraft.c
```diff
@@ -1008,6 +1008,104 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
...
+    SSL_CTX_set_verify(ssl_ctx, SSL_VERIFY_PEER, NULL);
+
+    if ((rr->config->tls_cert != NULL && rr->config->tls_key == NULL) ||
+            (rr->config->tls_key != NULL && rr->config->tls_cert == NULL)) {
+        RedisModule_Log(ctx, REDIS_WARNING, "REDIS_SSL_CTX_CERT_KEY_REQUIRED");
+//        if (error) *error = REDIS_SSL_CTX_CERT_KEY_REQUIRED;
+        goto error;
+    }
+
+    if (!SSL_CTX_load_verify_locations(ssl_ctx, rr->config->tls_ca_cert, NULL)) {
```
tezc, line 1045: Don't we want to print TLS errors here and in similar checks down below?

src/redisraft.c
```diff
@@ -1008,6 +1008,104 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
...
+//        if (error) *error = REDIS_SSL_CTX_CLIENT_CERT_LOAD_FAILED;
+        goto error;
+    }
+    if (!SSL_CTX_use_PrivateKey_file(ssl_ctx, rr->config->tls_key, SSL_FILETYPE_PEM)) {
+        RedisModule_Log(ctx, REDIS_WARNING, "REDIS_SSL_CTX_PRIVATE_KEY_LOAD_FAILED");
+//        if (error) *error = REDIS_SSL_CTX_PRIVATE_KEY_LOAD_FAILED;
+        goto error;
+    }
+
+    if (*rr->config->tls_key_pass != 0) {
```
tezc, line 1062:
```suggestion
    if (*rr->config->tls_key_pass != '\0') {
```

Let's make it obvious and compare with char.  Btw, `tls_key_pass` can be NULL right? If so, this line would crash.

src/redisraft.c
```diff
@@ -1008,6 +1008,104 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
...
+        SSL_CTX_set_default_passwd_cb(ssl_ctx, tlsPasswordCallback);
+        SSL_CTX_set_default_passwd_cb_userdata(ssl_ctx, (void *) rr->config->tls_key_pass);
+    }
+
+    RedisModule_Log(ctx, REDIS_WARNING, "returning %p", ssl_ctx);
+    return ssl_ctx;
+
+error:
+    SSL_CTX_free(ssl_ctx);
+    RedisModule_Log(ctx, REDIS_WARNING, "returning %p", NULL);
```
tezc, line 1072: Letftover?

src/redisraft.c
```diff
@@ -1008,6 +1008,104 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
...
+//        if (error) *error = REDIS_SSL_CTX_PRIVATE_KEY_LOAD_FAILED;
+        goto error;
+    }
+
+    if (*rr->config->tls_key_pass != 0) {
+        SSL_CTX_set_default_passwd_cb(ssl_ctx, tlsPasswordCallback);
+        SSL_CTX_set_default_passwd_cb_userdata(ssl_ctx, (void *) rr->config->tls_key_pass);
+    }
+
+    RedisModule_Log(ctx, REDIS_WARNING, "returning %p", ssl_ctx);
```
tezc, line 1067: Leftover? If not, let's use debug level and LOG_DEBUG().

src/redisraft.c
```diff
@@ -1265,23 +1363,20 @@ __attribute__((__unused__)) int RedisModule_OnLoad(RedisModuleCtx *ctx, RedisMod
...
-                                         &ssl_error);
-         if (!rr->ssl) {
-             RedisModule_Log(ctx, REDIS_WARNING, "Failed to create ssl context.");
-             LOG_ERROR("Error: %s", redisSSLContextGetError(ssl_error));
-             return REDISMODULE_ERR;
-         }
+        RedisModule_Log(ctx, REDIS_WARNING, "Generating SSL Context");
+        rr->ssl = generateSSLContext(ctx, rr);
     }
+    RedisModule_Log(ctx, REDIS_WARNING, "ssl context = %p", rr->ssl);
```
tezc, line 1379: Leftover? If not, let's use debug level and LOG_DEBUG()

docs/TLS.md
sjpotter: we can, just seemed we could be a little simpler.  not biggy.

src/config.c
```diff
@@ -584,6 +600,41 @@ static char *getRedisConfig(RedisModuleCtx *ctx, const char *name)
     return buf;
 }
 
+void updateTLSConfig(RedisModuleCtx *ctx, RedisRaftConfig *config) {
+    config->tls_ca_cert = getRedisConfig(ctx, "tls-ca-cert-file");
```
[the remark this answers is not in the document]
sjpotter: yes, fixed, free added before.

src/redisraft.c
sjpotter: don't understand the Q? like an openssl error function to get an error string?  I don't see those that do it.  otherwise we have our own logging (now converted to `LOG_ERROR()`)

src/redisraft.c
sjpotter: fixed both issues.

src/redisraft.c
sjpotter: yes, already got to actually.

src/redisraft.c
sjpotter: leftover, was from debugging.

docs/TLS.md
tezc: You mean with shorter names? I love short names but in this case, probably it's better to follow Redis config name. So, we don't maintain conversion table here and people don't have to look at it.

src/redisraft.c
tezc: When `SSL_CTX_load_verify_locations()` fails, we won't see what is the error. Why don't we print the error? Something like 
```
unsigned long e = ERR_peek_last_error();
LOG_ERROR("SSL_CTX_load_verify_locations(): %s", ERR_reason_error_string(e))
```

sjpotter: ok, done.

src/connection.c
```diff
@@ -87,6 +87,12 @@ static void ConnFree(Connection *conn)
 
     LIST_REMOVE(conn, entries);
 
+#ifdef SSL
```
tezc, line 90:
```suggestion
#ifdef HAVE_TLS
```

I don't know if `SSL` was defined somewhere. Otherwise, it'll be a leak probably. Just fyi, when you need, you can manually trigger daily CI for your branch here: https://github.com/RedisLabs/redisraft/actions/workflows/daily.yml

src/redisraft.h
```diff
@@ -412,9 +418,12 @@ typedef struct RedisRaftConfig {
     char *ignored_commands;             /* Comma delimited list of commands that should not be intercepted */
     int external_sharding;              /* use external sharding orchestrator only */
     bool tls_enabled;                   /* use TLS for all inter cluster communication */
+    bool tls_trace;                     /* trace TLS connections for debugging purposes */
```
tezc, line 421: leftover?

src/config.c
```diff
@@ -545,7 +555,7 @@ static RRStatus setRedisConfig(RedisModuleCtx *ctx, const char *param, const cha
     return ret;
 }
 
-static char *getRedisConfig(RedisModuleCtx *ctx, const char *name)
+char *getRedisConfig(RedisModuleCtx *ctx, const char *name)
```
tezc, line 558: If I'm not mistaken, we don't need to export this function

src/redisraft.c
```diff
@@ -1008,6 +1008,100 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
...
+
+    if (!pass) return -1;
+    pass_len = strlen(pass);
+    if (pass_len > (size_t) size) return -1;
+    memcpy(buf, pass, pass_len);
+
+    return (int) pass_len;
+}
+
+SSL_CTX *generateSSLContext(RedisModuleCtx *ctx, RedisRaftCtx *rr) {
```
tezc, line 1027:
```suggestion
SSL_CTX *generateSSLContext(RedisModuleCtx *ctx, RedisRaftCtx *rr) 
{
```

src/redisraft.c
```diff
@@ -1008,6 +1008,100 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
...
+    }
+
+#ifdef HAVE_TLS
+    if (!redis_raft.config->tls_enabled || redis_raft.config->tls_manual) {
+        return;
+    }
+
+    RedisModuleConfigChangeV1 *ei = data;
+
+    for(unsigned int i = 0; i < ei->num_changes; i++) {
```
tezc, line 1087:
```suggestion
    for (unsigned int i = 0; i < ei->num_changes; i++) {
```

src/redisraft.c
```diff
@@ -1265,23 +1359,20 @@ __attribute__((__unused__)) int RedisModule_OnLoad(RedisModuleCtx *ctx, RedisMod
...
-                                         &ssl_error);
-         if (!rr->ssl) {
-             RedisModule_Log(ctx, REDIS_WARNING, "Failed to create ssl context.");
-             LOG_ERROR("Error: %s", redisSSLContextGetError(ssl_error));
-             return REDISMODULE_ERR;
-         }
+        RedisModule_Log(ctx, REDIS_WARNING, "Generating SSL Context");
+        rr->ssl = generateSSLContext(ctx, rr);
     }
+    RedisModule_Log(ctx, REDIS_WARNING, "ssl context = %p", rr->ssl);
```
tezc, line 1375: leftover?

src/redisraft.c
```diff
@@ -1008,6 +1008,100 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
     return REDISMODULE_OK;
 }
 
+#ifdef HAVE_TLS
+/* Callback for passing a keyfile password stored as a char * to OpenSSL,  copied from redis */
+static int tlsPasswordCallback(char *buf, int size, int rwflag, void *u) {
```
tezc, line 1013: just for consistency, I'd write this function with redisraft style.

tezc:
LGTM, my last comments are minor & nitpicks. Just realized, we don't have a test for reconfiguration, it might be added (maybe something like just reloading same certs). Otherwise, lgtm.

src/config.c
sjpotter: its now called by the callback for config changes, hence why its exported.  I guess I could put the callback in config.c but then its in a different place from the other callbacks.

src/redisraft.c
sjpotter: ok, think I did that now.

sjpotter:
looking into writing a test for the "reconfiguration" as noted

src/config.c
tezc: If I'm not mistaken, this is called from `updateTLSConfig()` only, this function is also inside config.c. So, `getRedisConfig()` might still be static. Just ignore my comment if this is not the case :)

sjpotter:
I'm not quite happy with the SSL_free issue, not understanding why its happening.  As we never saw it before handling reconfiguration, my only thought is that its a bug in openssl re us freeing the SSL_CTX which causes problem when we then do an SSL_free() on the per connection ssl object.  but I have no good idea otherwise.

src/redisraft.c
```diff
@@ -1265,22 +1364,18 @@ __attribute__((__unused__)) int RedisModule_OnLoad(RedisModuleCtx *ctx, RedisMod
...
-                                         rr->config->tls_cert,
-                                         rr->config->tls_key,
-                                         NULL,
-                                         &ssl_error);
-         if (!rr->ssl) {
-             RedisModule_Log(ctx, REDIS_WARNING, "Failed to create ssl context.");
-             LOG_ERROR("Error: %s", redisSSLContextGetError(ssl_error));
-             return REDISMODULE_ERR;
-         }
+        RedisModule_Log(ctx, REDIS_WARNING, "Generating SSL Context");
```
yossigo, line 1377: Leftover?

src/cluster.c
```diff
@@ -1527,7 +1527,12 @@ static void linkHandleResponse(redisAsyncContext *c, void *r, void *privdata)
     JoinLinkState *state = ConnGetPrivateData(conn);
 
     if (!reply) {
+#ifndef HAVE_TLS
         LOG_ERROR("RAFT.SHARDGROUP GET failed: connection dropped.");
+#else
+        unsigned long e = ERR_peek_last_error();
+        LOG_ERROR("RAFT.SHARDGROUP GET failed: connection dropped: %s", ERR_reason_error_string(e));
```
yossigo, line 1534: Consider keeping these ifdefs local and have a single function to handle getting the extended error information?

Besides that, are we sure we even need to report an SSL error always? For example, if we experience a TCP connect error - wouldn't we report a stale and misleading OpenSSL error here?

docs/TLS.md
```diff
@@ -15,10 +14,22 @@ In addition, one has to pass the `tls-enabled yes` option to the redisraft modul
...
-RedisRaft is smart enough to figure out which files to use for the local cert/key/ca by querying the underlying Redis server.  However, if a user wants to use different files for the client aspects vs server aspects, one can specify RedisRaft module options as well
+## Configuring RedisRaft TLS support
 
-```asm
-tls-ca-cert
-tls-cert
-tls-key
-```
\ No newline at end of file
+RedisRaft can be configured by reusing the Redis server's TLS configuration, or by configuring the RedisRaft module directly. In general, we expect people will configure the Redis server.
```
yossigo, line 19: Proposing to reconsider the double configuration option. What use case does it support?
I don't see a scenario where a RedisRaft redis-server will ever use the `tls-client-*` settings for anything, because it's never a cluster node / replica. That being the case, perhaps we can just assume we own these parameters and avoid the need to maintain this extra complexity?

yossigo (commented):
@sjpotter What is the concern regarding `SSL_free()`? I see that hiredis frees it when freeing its own context, am I missing something?

src/cluster.c
sjpotter: re: last point, this seems like it should be hiredis then?

docs/TLS.md
sjpotter: not 100% sure I understand what you mean.

sjpotter:
> @sjpotter What is the concern regarding `SSL_free()`? I see that hiredis frees it when freeing its own context, am I missing something?

hmm, if hiredis is freeing it, that's the issue and in general makes sense.  as there's no fundamental difference than using hiredis where we pass the ssl object in and hiredis creating it itself.  ok.  now that i understand it, can remove all the calls.

src/redisraft.c
sjpotter: yes, removed.

src/cluster.c
sjpotter: so I now understand the first part.  you dislike the #ifdef/else/#endif block in this location, and would rather a connectionFailedReasong() that produces a string and would contain it.

with that said, my comment on the last point still stands, as all connection handling is abstracted away from hiredis users, hiredis itself would have to tell us why the connection failed to do it in any way reliably, and I just don't see how to do that at the moment.

docs/TLS.md
sjpotter: do you mean, that we should just not have module config options and just always read it out of redis?

src/config.c
sjpotter: you were right I forgot I consolidated them into that function when I had them duplicated before and needed it exported.

docs/TLS.md
sjpotter: removed the module configs and simplified doc

src/cluster.c
sjpotter: adding 2 tickets for redisraft and hiredis https://github.com/RedisLabs/redisraft/issues/296

src/connection.c
```diff
@@ -87,6 +87,13 @@ static void ConnFree(Connection *conn)
 
     LIST_REMOVE(conn, entries);
 
+#ifdef HAVE_TLS
+    if (conn->ssl) {
+//        SSL_free(conn->ssl);
+        conn->ssl = NULL;
```
tezc, line 93: I think we don't even need `conn->ssl`. We don't use it anymore?

src/connection.c
```diff
@@ -318,12 +325,26 @@ static void handleResolved(void *arg)
...
+        }
+
+        SSL *ssl = SSL_new(conn->rr->ssl);
+        if (!ssl) {
+            RedisModule_Log(conn->rr->ctx, REDIS_WARNING, "Couldn't create SSL object");
+            goto fail;
+        }
+        int result = redisInitiateSSL(&conn->rc->c, ssl);
+        if (result != REDIS_OK) {
+//            SSL_free(ssl);
```
tezc, line 343: Taking a look at `redisInitiateSSL()`, it calls `SSL_free()` for some error cases but not all :( 
- OOM: It does not call `SSL_free()` on OOM. We can ignore this though.
- If the connection is already associated with a `redisSSL`, it doesn't call `SSL_free()`. 

It's not clear to me what we should do at this point. I believe this is a bug in hiredis API. Ownership is not clear. I also see other error paths inside hiredis this would be leaked. Maybe I'm missing something.

tezc: https://github.com/RedisLabs/redisraft/blob/5e90f4437ade2a3cfb0f35f4fdcf3e2770bc162f/deps/hiredis/ssl.c#L315

Hmm, we added this line. Probably, we should also add `SSL_free()` to other error cases up above.

docs/TLS.md
sjpotter: my impression was that there was no need to have module configs, as if we would want to distinguish the keys redisraft uses from the keys redis uses on the server side we would use tls-client-* as redis wont be using them (as no replicas/clustering)

tezc: Oh okay that's even better. 

>removed the module configs and simplified doc

Did you forget to push the commit for this? :) I still see module configs in the PR.

sjpotter: yes, that also goes with the other comment, now pushed

tezc: Ok so, we'll only use `tls-client-*` configs? If one wants to config redisraft, they set `tls-client-*` configs. I like this, makes code simpler. 

It needs another commit as far as I can see right? With this decision, we don't even need to touch `tls-key-file` , `tls-key-file-pass`, `tls-ca-cert-file` or `tls-cert-file`. There are places still using these configs. Or did I misunderstand again? :)

sjpotter: we need to set tls-key... as this is the server's files (i.e. redis-server)  we can choose to set tls-client-* if we want redisraft to us different files than the server.  if we don't, we don't have to set them.

tezc: Why don't we just ask users to set `tls-client-*` options? If they want to configure RedisRaft, they will set `tls-client-*`. If they want to use same certs with "server mode"(incoming side), they'll copy paste them as `tls-client-*`. So, RedisRaft will use them in the client mode (outgoing). Is there a downside for this? If we can do this, we just say RedisRaft TLS config is `tls-client-*`, that's it. Wdyt?

tezc: Just to recap:

Your suggestion: No module param configs. Use Redis' `tls-client-*` if exists, otherwise use `tls-` files in RedisRaft.
Mine: No module param configs. Use Redis' 'tls-client-*` only in RedisRaft.

Maybe mine is unnecessary, differs a bit from Redis anyway.

src/config.c
```diff
@@ -608,9 +625,10 @@ void ConfigInit(RedisModuleCtx *ctx, RedisRaftConfig *config)
     config->slot_config = "0:16383";
     config->shardgroup_update_interval = REDIS_RAFT_DEFAULT_SHARDGROUP_UPDATE_INTERVAL;
     config->max_appendentries_inflight = REDIS_RAFT_DEFAULT_MAX_APPENDENTRIES;
-    config->tls_ca_cert = getRedisConfig(ctx, "tls-ca-cert-file");
-    config->tls_key = getRedisConfig(ctx, "tls-key-file");
-    config->tls_cert = getRedisConfig(ctx, "tls-cert-file");
+#ifdef HAVE_TLS
+    updateTLSConfig(ctx, config);
+    config->tls_manual = false;
```
tezc, line 630: We don't need tls_manual now right?

sjpotter: correct.  will remove

src/redisraft.c
```diff
@@ -1008,6 +1008,105 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
...
+error:
+    SSL_CTX_free(ssl_ctx);
+    return NULL;
+}
+#endif
+
+
+void handleConfigChangeEvent(RedisModuleCtx *ctx, RedisModuleEvent eid, uint64_t subevent, void *data)
+{
+    LOG_ERROR("handleConfigChangeEvent: enter");
```
tezc, line 1080: leftover?

src/connection.c
sjpotter: yes, we should probably revert our change and fix hiredis / pull in updates hiredis.

src/redisraft.c
sjpotter: yes, will remove.

src/redisraft.c
```diff
@@ -1008,6 +1008,105 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
...
+    if (!redis_raft.config->tls_enabled || redis_raft.config->tls_manual) {
+        return;
+    }
+
+    RedisModuleConfigChangeV1 *ei = data;
+
+    for (unsigned int i = 0; i < ei->num_changes; i++) {
+        if (strcmp("tls-ca-cert-file", ei->config_names[i]) == 0 ||
+                strcmp("tls-key-file", ei->config_names[i]) == 0 ||
+                strcmp("tls-cert-file", ei->config_names[i]) == 0) {
```
tezc, line 1095: I think we'll need to detect `tls-client-*` family changes as well.

docs/TLS.md
sjpotter: the point is tls-client-* can be redundant.  we can't run redis-server without the normal tls-* as it handles incoming connections itself.  only if redisraft wants to use different files, will we need to set tls-client-*, otherwise it just be duplicating entries on the command line.

src/redisraft.c
sjpotter: correct.

src/connection.c
tezc: Damn, https://github.com/redis/hiredis/pull/939. 

They took this change back later, we didn't see it and merged that line into our hiredis version. So, you're right. I think we need to revert the change and call `SSL_free()` if this function fails.

src/redisraft.c
sjpotter: adde

src/connection.c
tezc: After rebasing to https://github.com/RedisLabs/redisraft/commit/dc1752ea467fe0328fd4dc0f88d9cf4b07ff1336
We can uncomment this line I believe.

src/connection.c
```diff
@@ -311,10 +311,18 @@ static void handleResolved(void *arg)
...
     }
 
+
 #ifdef HAVE_TLS
     if (conn->rr->config->tls_enabled) {
-        if (redisInitiateSSLWithContext(&conn->rc->c, conn->rr->ssl) != REDIS_OK) {
-            CONN_LOG_WARNING(conn, "SSL Error!");
+        SSL *ssl = SSL_new(conn->rr->ssl);
+        if (!ssl) {
+            CONN_LOG_WARNING(conn, "Couldn't create SSL object");
```
yossigo, line 319: Consider printing the error message from the OpenSSL stack here, and on any other OpenSSL failure, to help debugging.

yossigo (approved):
LGTM, minor comment about logging messages.

src/connection.c
sjpotter: ok, done

sjpotter:
@yossigo added a bit more libssl error printing in the CONN_LOG_WARNING as you seemed to want.  need a reapprove I think.

yossigo (approved):

PR #287, opened 2022-03-13
Follower should reply existing requests with MOVED or CLUSTERDOWN

tezc:
If a leader node becomes follower, it can redirect existing requests from the previous term to the new leader.

@sjpotter suggested this change a while ago IIRC. In case of leader transfer or for any other reason, when there is another leader, we can redirect requests if possible.

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -79,10 +79,17 @@ RaftReq *entryDetachRaftReq(RedisRaftCtx *rr, raft_entry_t *entry)
  */
 void entryFreeAttachedRaftReq(raft_entry_t *ety)
 {
+    RedisRaftCtx *rr = &redis_raft;
     RaftReq *req = entryDetachRaftReq(&redis_raft, ety);
 
     if (req) {
-        RedisModule_ReplyWithError(req->ctx, "TIMEOUT not committed yet");
+        /* If not the leader, checkLeader() will generate a reply. */
+        if (checkLeader(rr, req->ctx, NULL) == RR_OK) {
+            /* We should not reach here as entry rollback only happens when
+             * another node becomes leader. Just being defensive and leaving
+             * this line here. */
+            RedisModule_ReplyWithError(req->ctx, "TIMEOUT not committed yet");
+        }
         RaftReqFree(req);
     }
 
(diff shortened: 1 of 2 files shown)


sjpotter:
in the process of reviewing this, I was digging into `pop_read_queue()` usage and when can_read is 0 or 1, and I'm wondering if `raft_process_read_queue()` could be much simpler.

my first thought is that "timeout" is just incorrect.  this implies (to me) a configurable parameter that was exceeded.  This isn't true.  it means that leadership transfered.  Perhaps, that should be made clear.

my next thought goes to `raft_process_read_queue()`.  I'm wondering if it can simplified / improved.

i.e. we first do

```
    /* As a follower we drop all queued read requests */
    if (raft_is_follower(me_)) {
        while (me->read_queue_head) {
            pop_read_queue(me, 0);
        }
        return;
    }
```

then we do

```
    if (!raft_is_leader(me_))
        return;
```

that seems redundant, I'd think?

I'm wondering if in general, the whole thing can be simplified to the check at the end

```    while (me->read_queue_head &&
            me->read_queue_head->msg_id <= last_acked_msgid &&
            me->read_queue_head->read_idx <= last_applied_idx) {
        pop_read_queue(me, me->read_queue_head->read_term == me->current_term);
```

i.e. wouldn't this handle leadership transfer as well.

tezc:
I believe we can simplify it as part of another PR ofcourse. There is another hard to understand "if check" in this function which is due to we don't commit NOOP entry if term is 1. No need for this, we can commit NOOP entry even it's the first entry. These changes need another PR, we can take a look later I guess.

yossigo (commented):
I wonder how this affects guarantees around replies, though. I think the essence of `-TIMEOUT` is we don't know if the operation was committed or not. With `-MOVED` we generally guarantee the operation is not performed and should be repeated after redirection, so I think this might create some kind of ambiguity no?

sjpotter:
is it possible to force close the connection without a response?  that seems to be the analogue to what would happen in redis cluster?

sjpotter:
is RedisModule_DeauthenticateAndCloseClient() what I'm proposing above?

tezc:
Don't know how Redis handle it but I feel like any response is better than no response.

sjpotter:
its not really no response, it be a proactively broken connection, but I can understand why that might seem distasteful.  don't have huge strong feelings beyond what I already said.

tezc:
Client will not understand our intention. So, let's keep `-TIMEOUT`.

PR #292, opened 2022-03-14
Catch crashes and failures together in Jepsen results

tezc:
We don't catch crashes in Jepsen test results, so, tests become green in CI even all tests are crashed. Changed regex to require "0 crashed" and  "0 failures" together, in any order.

Sample output to try:
```
0 successes
0 unknown
20 crashed
0 failures
Cleaning up orphan processes
```


sjpotter:
did you mean to close this?

tezc:
Yeah, realized we treat Jepsen crashes as part of normal workflow, this fix wouldn't work. Need to fix those crashes first.

PR #293, opened 2022-03-16
ASK/ASKING support - POC

sjpotter:


src/raft.c
```diff
@@ -1667,6 +1848,9 @@ void handleRedisCommand(RedisRaftCtx *rr,RaftReq *req)
      * commands we've received this as a RAFT.ENTRY input and bundling, probably through a
      * proxy, and bundling was done before.
      */
+
+    handleAsking(req);
```
tezc, line 1852: Why do you do this here and not inside `executeRaftRedisCommandArray()` ? 

Also, why do you serialize "ASKING" differently? Why not just leave it as it is and check inside `executeRaftRedisCommandArray()` like "MULTI" here: 

https://github.com/RedisLabs/redisraft/blob/e573dd291b76591e736d74d3eea1be059aac8bad/src/raft.c#L148

tezc, line 1852: Why do you do this here and not inside `executeRaftRedisCommandArray()` ? 

Also, why do you serialize "ASKING" differently? Why not just leave it as it is and check inside `executeRaftRedisCommandArray()` like "MULTI" here: 

https://github.com/RedisLabs/redisraft/blob/e573dd291b76591e736d74d3eea1be059aac8bad/src/raft.c#L148

sjpotter: 1) we have to tag asking for multis.  i.e. it will be

asking multi
get x
get y
exec

sjpotter: 2. the MULTI isn't visible in executeRaftRedisCommandAarray, that's in handleMulti() I'd say we are handling asking similiar to multi impractice in the manner I implemented.

tezc: Hmm, okay I see. 

Need to take another look but if multi handling also checks for "ASKING MULTI" along with "MULTI", will it solve the problem? We can serialize "ASKING" without special serializing. Inside `executeRaftRedisCommandArray()`, as the first thing `validateRaftRedisCommandArray()` runs and takes "ASKING" out from the array. Rest of the code stays same. Something like that. 

Just saying this because it might solve the problem with less changes.

sjpotter: in multi we don't serialize the multi/exec portions of it, we just serialize the array of commands in between them.  so, an asking multi would need a flag.  this seems like the cleanest way, i.e. asking is basically setting a flag on the request, and that's how we treat it.  In practice, the changes weren't that extensive (imo).

src/raft.c
```diff
@@ -183,6 +183,126 @@ static void executeRaftRedisCommandArray(RaftRedisCommandArray *array,
...
 }
 
+typedef enum KeysStatus{
+    NoKeys,
+    AllExist,
+    SomeExist,
+    NoneExist,
+} KeysStatus;
+
+KeysStatus validateKeyExistence(RedisRaftCtx *rr, RaftRedisCommandArray *cmds) {
```
tezc, line 193:
```c
    int total_keys = 0;
    int found = 0;

    for (int i = 0; i < cmds->len; i++) {
        RaftRedisCommand *cmd = cmds->commands[i];

        /* Iterate command keys */
        int num_keys;
        int *keyindex = RedisModule_GetCommandKeys(rr->ctx, cmd->argv, cmd->argc, &num_keys);

        total_keys += num_keys;
        for (int j = 0; j < num_keys; j++) {
            found += RedisModule_KeyExists(rr->ctx, cmd->argv[keyindex[j]]);
        }
        RedisModule_Free(keyindex);
    }

    if (found != total_keys) {
        return (found == 0) ? NoneExist : SomeExist;
    }

    return (total_keys == 0) ? NoKeys : AllExist;
```
I didn't try but something like above, would that be equivalent? Counting keys found and compare at the end. You can move this logic  into place you call function, then maybe no need for enum definitions and switch case after function call. Just consider if it makes sense. (And ofcourse if it is correct).

src/raft.c
```diff
@@ -1384,7 +1519,11 @@ static void handleReadOnlyCommand(void *arg, int can_read)
         goto exit;
     }
 
-    executeRaftRedisCommandArray(&req->r.redis.cmds, req->ctx, req->ctx);
+    RaftRedisCommandArray * cmds = &req->r.redis.cmds;
+
+    if (!redis_raft.config->sharding || validateRaftRedisCommandArray(&redis_raft, cmds, req->ctx) == RR_OK) {
```
tezc, line 1524: `redis_raft.config->sharding` is this check for shortcut circuit, like for performance? Is there way to avoid this? Because state machine becomes dependent on config, we take logs and rdb, restart in another machien without `sharding` config. We'll have hard time to understand what is going on :)

sjpotter: yea, but if sharding is disabled, we might have other issues too (i.e. allow multi slot operations through)

tezc: That applies to new operations. I'm worried about current data set. I think log replay should not rely on things that are not part of the state machine. Just easy to make mistakes.  It should be deterministic, replay should always give the same result. Just consider if we really need this check.

src/raft.c
tezc: Iterating all keys should not be a problem as we are going to do that for almost all operations. (For most operations all keys will be found). But yeah, you can add a shortcut too.

src/raft.c
sjpotter: ok, that makes sense.  I could do the same optimization as the slot is stored, so we can default the slot to -1 for non sharding cases and use that to short circuit.

sjpotter: ok, made the change I mentioned, no longer based on configuration, based on slot == -1 which it will be by default unless handleSharding() changes it.

src/raft.c
tezc: https://github.com/RedisLabs/redisraft/blob/e573dd291b76591e736d74d3eea1be059aac8bad/src/raft.c#L148

Maybe I'm missing something but array of commands have "multi" in it, we treat it differently up above. 

I think this is not directly related with what I suggest here. So, my suggestion is keeping "asking" at the head whether it is multi or not and handling it just before we apply the command.

Here some pseudo-code, just consider if it makes sense. If it works and doesn't cause any other problems (not sure), I feel like you may avoid some of the changes, special handling (serialization, checks around executeRaftRedisCommandArray() etc). 

```c
validateRaftRedisCommandArray(commandArray) 
{
     if (commandArray->commands[0]->argv == "ASKING") {
         asking = true;
         popAskingFromArray(commandArray);
     }
     
     ...
}

executeRaftRedisCommandArray(commandArray) 
{
      if (validateRaftRedisCommandArray(commandArray) != OK) {
          return;
      }
      
      ...
      applyCommands();
      
}
```

src/cluster.c
```diff
@@ -1170,7 +1170,9 @@ RRStatus computeHashSlotOrReplyError(RedisRaftCtx *rr,
             } else {
                 if (*slot != thisslot) {
                     RedisModule_Free(keyindex);
-                    RedisModule_ReplyWithError(ctx, "CROSSSLOT Keys in request don't hash to the same slot");
+                    if (ctx) {
```
tezc, line 1173: Probably, it's better to change this function to `computeHashSlot()`. Caller can decide to reply if ctx exists or not.

src/raft.c
```diff
@@ -195,21 +388,39 @@ static void executeLogEntry(RedisRaftCtx *rr, raft_entry_t *entry, raft_index_t
...
+        reply_ctx = NULL;
+    }
+
+    HandleAsking(cmds);
+
+    /* When we're in cluster mode, go through handleSharding. This will perform
+     * hash slot validation and return an error / redirection if necessary. We do this
+     * before checkLeader() to avoid multiple redirect hops.
+     */
+    if (rr->config->sharding && handleSharding(rr, req, cmds) != RR_OK) {
```
tezc, line 418: I think we shouldn't rely on config at this point. I've forgotten completely, why do we have this config anyway? :)

src/raft.c
```diff
@@ -1298,14 +1517,25 @@ void handleAppendEntries(RedisRaftCtx *rr, RaftReq *req)
...
-    executeRaftRedisCommandArray(&req->r.redis.cmds, req->ctx, req->ctx);
+    RaftRedisCommandArray * cmds = &req->r.redis.cmds;
+
+    HandleAsking(cmds);
+
+    if (rr->config->sharding && handleSharding(rr, req, cmds) != RR_OK) {
+        goto exit;
+    }
+
+    if (cmds->slot == -1 || validateRaftRedisCommandArray(&redis_raft, cmds, req->ctx) == RR_OK) {
```
tezc, line 1536: I think if you move this part info `executeRaftRedisCommandArray()`, you can avoid changes in this function, `handleReadOnlyCommand()` and in `executeLogEntry()`. Both needs to handle "asking, sharding and slot validation". If `executeRaftRedisCommandArray()` does not, we'll doing these in a single place. Just consider if it makes sense.

src/raft.c
sjpotter: sharding enabled prevents cross slot, while sharding disabled allows cross slots in a multi/multi key command.

so if running a single redis raft cluster, might want sharding disabled for more flexibility (i.e. normal single redis server behavior).

the only way to handle this is to put the behavior I previously had back in, i.e. expanding the log format a little to calculate the slot value before append, so each node will behave the same and not be based on configuration.

tezc: Hmm, okay, I see.

There might be a few ways to solve this. Just thinking out loud: 

1 - One way might be having that config in the state machine. As we currently have "sharding info" in the state machine, can we use that somehow? Like can we decide if sharding is enabled or not from that?. Personally, I need sometime just to go over code, can't think right now :) 

2- Your suggestion is another option. Just writing down:
- Leader computes the slot, if sharding is enabled rejects cross slot violation before appending
- If sharding is disabled, it sets `slot` as `-1` to tell state machine skip validating `slot` against slot mapping.
- State machine will get "slot" of command array as `-1` or `some real value`. If slot is `-1`, we skip validation slot mapping as these are cross slot commands and sharding is disabled. If it's a "real value", we'll check sharding mapping.

According to this, leader computes `slot`, and then sets `slot` in RaftReq and writes it to the log. On state machine, if RaftReq exists, we use slot from it, otherwise deserialize entry. Just a minor addition to this, what if we skip writing `slot` to the log but still set `slot` in RaftReq. If RaftReq exists, state machine will use from it. If it doesn't exist, `slot` value is computed after deserializing. We may detect cross slot violation at this point but this is only because sharding was disabled and this log entry was allowed. In this case, same as before, we can skip validating slot. 

Just thinking if there is a simple way we can avoid writing `slot` to the log. (For performance reasons).

src/raft.c
sjpotter: ok, tried this.

src/raft.c
sjpotter: so this is my conception.  

if we have a single shard and that shard is the full slot range, then we don't have to test sharding enabled config.

yes, this could conceptually be reconfigured later, but at this point in time its the equivalent of just a single shard cluster so might as well treat it as such.

thoughts?  @yossi @ozan?

sjpotter: basically something alone the lines of

```
bool isSharding(RedisRaftCtx *rr) {
    if (si->shard_groups_num != 1) {
        return false;
    }
    
    RedisModuleDictIter *iter = RedisModule_DictIteratorStartC(si->shard_group_map, "^", NULL, 0);

    char *key;
    size_t key_len;
    ShardGroup *sg;

    key = RedisModule_DictNextC(iter, &key_len, (void **) &sg)) != NULL);
    RedisModule_DictIteratorStop(iter);
    
    if (!sg->local || sg->slot_ranges_num != 1 || 
        sg->slot_ranges[0].start_slot != 0 || sg->slot_ranges[0].end_slot != 16383 ||
        sg->slot_ranges[0].type != SLOTRANGE_TYPE_STABLE) {
        return false
    }
    
    return true;
}
```

src/cluster.c
sjpotter: ok done

src/raft.c
sjpotter: moved the handleAsking into RaftExecuteCommandArray() which I think addresses the issue?

sjpotter:
one thing I realized here is that our command processing needs to be smarter to skip ASKING in its determination of what the command is, so needs to be extended for that.

PR #294, opened 2022-03-17
Allow info command even cluster is not initialized

tezc:
We intercept "info" command and check cluster state before replying. If cluster is uninitialized, we return an error. Changed it to work even without cluster is initialized.


sjpotter:
probably useful to add a test for info working before initialization. otherwise LGTM

sjpotter (dismissed):

tezc:
@sjpotter Added a test. Need approval again.

sjpotter (approved):

PR #295, opened 2022-03-21
Move nodeshutdown handling to command callback

tezc:
Move nodeshutdown handling to command callback.

Also, fixes minor leak if nodeshutdown id is wrong.

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -17,21 +17,21 @@
 #include "redisraft.h"
 
 const char *RaftReqTypeStr[] = {
-    "<undef>",
-    "RR_CLUSTER_INIT",
-    "RR_CLUSTER_JOIN",
-    "RR_CFGCHANGE_ADDNODE",
-    "RR_CFGCHANGE_REMOVENODE",
-    "RR_APPENDENTRIES",
-    "RR_REDISCOMMAND",
-    "RR_INFO",
-    "RR_DEBUG",
-    "RR_CLIENT_DISCONNECT",
-    "RR_SHARDGROUP_ADD",
-    "RR_SHARDGROUPS_REPLACE",
-    "RR_SHARDGROUP_GET",
-    "RR_SHARDGROUP_LINK",
-    "RR_TRANSFER_LEADER"
+    [0]                       = "<undef>",
+    [RR_CLUSTER_INIT]         = "RR_CLUSTER_INIT",
+    [RR_CLUSTER_JOIN]         = "RR_CLUSTER_JOIN",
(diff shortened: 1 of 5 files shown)


src/raft.c
```diff
@@ -25,10 +25,10 @@ const char *RaftReqTypeStr[] = {
     "RR_APPENDENTRIES",
     "RR_REDISCOMMAND",
     "RR_INFO",
-    "RR_COMPACT",
+    "RR_DEBUG",
     "RR_CLIENT_DISCONNECT",
     "RR_SHARDGROUP_ADD",
-    "RR_SHARDGROUP_UPDATE",
+    "RR_SHARDGROUPS_REPLACE",
```
sjpotter, line 31: my only nit is this.  yea, hard to keep in sync.  perhaps there's some gcc magic we can use to convert defines into strings so don't need this list?

sjpotter (dismissed):

src/raft.c
tezc: I don't know any simple magic for this and complex tricks isn't worth the effort most of the time. :) 

We can even delete this array just print integer enum values instead. This is only logging anyway. 

Alternatively, I added a commit. Manually, assigning enums to their strings and a test for if we have a gap in the array. Probably, this might help. Let me know which one you prefer.

src/raft.c
```diff
@@ -17,21 +17,21 @@
...
+    [RR_APPENDENTRIES]        = "RR_APPENDENTRIES",
+    [RR_REDISCOMMAND]         = "RR_REDISCOMMAND",
+    [RR_INFO]                 = "RR_INFO",
+    [RR_DEBUG]                = "RR_DEBUG",
+    [RR_CLIENT_DISCONNECT]    = "RR_CLIENT_DISCONNECT",
+    [RR_SHARDGROUP_ADD]       = "RR_SHARDGROUP_ADD",
+    [RR_SHARDGROUPS_REPLACE]  = "RR_SHARDGROUPS_REPLACE",
+    [RR_SHARDGROUP_GET]       = "RR_SHARDGROUP_GET",
+    [RR_SHARDGROUP_LINK]      = "RR_SHARDGROUP_LINK",
+    [RR_TRANSFER_LEADER]      = "RR_TRANSFER_LEADER"
```
sjpotter, line 34: I assume this is to make it obvious at compile time if we remove a req type, so build will break?

sjpotter (approved):

src/raft.c
tezc: https://github.com/RedisLabs/redisraft/pull/295#discussion_r830889466 

I asked you here if you like this one better :)

PR #298, opened 2022-03-22
try to help debug weird error by better testing

sjpotter:
reorder and add some tests to help diagnose error in future

we had a weird error here https://github.com/sjpotter/redisraft/runs/5648599700?check_suite_focus=true

that i've only seen once (and we've run this test many times), but as it should never fail, a bit concerned, so tried to improve the test a bit to help diagnose it.

### DIFF
--- a/tests/integration/test_random.py
+++ b/tests/integration/test_random.py
@@ -313,12 +313,14 @@ def test_raft_sort_sets(cluster):
     sinter = cluster.execute("raft._sort_reply", "sinter", "test", "test1")
     sunion = cluster.execute("raft._sort_reply", "sunion", "test", "test1")
     sdiff = cluster.execute("raft._sort_reply", "sdiff", "test", "test1")
-    smembers = cluster.execute("raft._sort_reply", "smembers", "test")
+    smembers1 = cluster.execute("raft._sort_reply", "smembers", "test")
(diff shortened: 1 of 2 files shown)


tezc (dismissed):

tezc:
@sjpotter  I saw a failure on my local, opened this issue: https://github.com/RedisLabs/redisraft/issues/302

sjpotter:
as your case was different, I did a more intensive change and to add debug logs and turn on debugging in their pytests

tests/integration/test_random.py
```diff
@@ -20,6 +20,8 @@ def test_hash_deterministic_order(cluster):
     """
 
     cluster.create(3)
+    for i in [1, 2, 3]:
+        cluster.node(i).raft_config_set('loglevel', 'debug')
```
tezc, line 24: Don't we already use "debug" level in tests?

tezc (approved):

tests/integration/test_random.py
sjpotter: wasn't sure, wanted to be sure, as this isn't common, but you might be right.  doesn't hurt as is, just redundant at worst.

tezc: okay

Issue #288, opened 2022-03-13
RAFT.SHARDGROUP LINK failure will mark state as REDIS_RAFT_UNINITIALIZED


tezc:
If wrong address is given, RAFT.SHARDGROUP LINK failure will set state flag as REDIS_RAFT_UNINITIALIZED. Shardgroup link uses same functions with join. 

The problematic line: 
https://github.com/RedisLabs/redisraft/blob/8e7446e094f6279aca72f2028773b3bc07234da8/src/common.c#L232

From now on, state is actually up but the flag is REDIS_RAFT_UNINITIALIZED. We can expect a crash depending on the next operations.

PR #299, opened 2022-03-23
Refactor cluster commands

tezc:
- Moved cluster command handling into command callback
- Refactored and moved raft library initialization into a single function

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -143,6 +143,15 @@ RRStatus checkRaftNotLoading(RedisRaftCtx *rr, RedisModuleCtx *ctx)
     return RR_OK;
 }
 
+RRStatus checkRaftUninitialized(RedisRaftCtx *rr, RedisModuleCtx *ctx)
+{
+    if (rr->state != REDIS_RAFT_UNINITIALIZED) {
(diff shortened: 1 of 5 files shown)


sjpotter (approved):
its fine, one thing I want to fix is the joinLinkIdleCallback (probably another PR)

Currently, it is the element itself the moves the addr (and with the ugly started bool I added).

I'm pondering if it should just take the addr head as is, and its up for the callbacks to iterate the addr on failure.  that way, its easier to implement different strategies (do you loop back to the front or not, how many times...).  The negative is, if it turns out that we only have one strategy, it would have to be implemented in multiple places.

tezc:
Regarding `started` bool, I think we can set `iter = first->addr` when we create `join` object, so, `idleCallback()` will take `iter` to a local variable and and advance `iter = iter->next` immediately. Later if we find `iter == NULL` means we tried all addresses. It might simplify a bit? 

I think we need to refactor sometime. Maybe when we are fixing this issue: https://github.com/RedisLabs/redisraft/issues/288 . I also find `handleNodeAddResponse()` difficult to read. Maybe we can simplify it as well.

PR #304, opened 2022-03-30
Refactor multi handling

tezc:
Refactored multi handling.

- Moved multi command handling into multi.c
  - Changed multi handling to work without RaftReq
  - Deleted leftover checks like  `REDISMODULE_CTX_FLAGS_MULTI_DIRTY`. 
  - Changed client disconnect handling not to use RaftReq

### DIFF
--- a/src/multi.c
+++ b/src/multi.c
@@ -1,8 +1,7 @@
 /* Handle MULTI/EXEC transactions here.
  *
  * If this logic was applied, the return value is true, indicating no further
- * processing is required. Otherwise, the main handleRedisCommand() flow is
- * applied.
+ * processing is required.
  *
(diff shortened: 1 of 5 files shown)


src/multi.c
```diff
@@ -0,0 +1,146 @@
+/* Handle MULTI/EXEC transactions here.
```
yossigo, line 1: @tezc Are there changes here? It's very hard to review if that's the case. I suggest to keep one commit for changes and another for transfer of code.

tezc: Yes, updated the PR. Please take a look at the first PR for the changes.

tezc: Yes, updated the PR. Please take a look at the first PR for the changes.

yossigo (approved):

PR #307, opened 2022-03-31
refactor "intercepted" (i.e. reinterpreted) commands

sjpotter:
instead of treating them like normal redis commands in raft, we intercept them early before the RaftReq for normal redis commands is created.  Previously, it was after and handled as part of handling the RaftReq.

this is motivated by migrate really needing its own RaftReq type


sjpotter:
ok, not blocking on this.  just thought it might be useful, but seems already happening elsewhere

yossigo:
Closing as obsolete.

PR #317, opened 2022-04-04
only reset redisraft state on connection failure for join

### LINKED ISSUE #288: RAFT.SHARDGROUP LINK failure will mark state as REDIS_RAFT_UNINITIALIZED

sjpotter:
we were resetting to unitialized on any connection failure, which was breaking clusters with shardgroup link

fixes #288

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -239,8 +239,8 @@ void joinLinkIdleCallback(Connection *conn)
 exit_fail:
     ConnAsyncTerminate(conn);
     /* only reset state to UNINITIALIZED on a join failure */
-    if (!strcmp(state->type, "join")) {
-        rr->state = REDIS_RAFT_UNINITIALIZED;
+    if (state->fail_callback) {
+        state->fail_callback(conn);
     }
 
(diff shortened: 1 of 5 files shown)


tezc:
@sjpotter I think we can call "complete callback" even on failure. 

https://github.com/RedisLabs/redisraft/blob/14416ab2dab5ac6a468baf28a0cf740be28c88ac/src/join.c#L106

We can extend it, `void (*complete_callback)(RaftReq *req, bool success))`. So, we always call the callback if provided, either with `success` true or false. So, caller can handle it in the same place and join.c stays agnostic to raft state. Wdyt?

sjpotter:
sure.

tezc (approved):

PR #319, opened 2022-04-05
Skip OOM test in redis suite

tezc:
Skip OOM test in redis suite as `maxmemory` config is not supported in RedisRaft.


sjpotter:
why?   i assume its fine, but description should probably say why.

tezc:
@sjpotter sure, added.

sjpotter (approved):

PR #322, opened 2022-04-06
Fix config get crash

bugwz:
# Describe

When I use `config get *`, I find that the server crashes directly, I found that the of variable `ignored_commands` was not initialized. In addition, I found some problems with `stringmatchlen`:
+ https://github.com/redis/redis/commit/c710d4afdccc0c797745bc3264f3f32a4cdd85da
+ https://github.com/redis/redis/commit/e17f9311c8ee18996e4b932a1284e4cfe29c05b4
+ https://github.com/redis/redis/commit/f43eb5adcfa3be6368f116656348a29f22f580eb

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ utils/create-cluster/*.db
 utils/create-cluster/*.db.idx
 venv
 buildinfo.h
+build/*
--- a/src/config.c
+++ b/src/config.c
@@ -598,7 +598,7 @@ void ConfigGet(RedisRaftCtx *rr, RedisModuleCtx *ctx, RedisModuleString **argv, 
     }
     if (stringmatch(pattern, CONF_IGNORED_COMMANDS, 1)) {
         len++;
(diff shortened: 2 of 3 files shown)


tezc:
@bugwz LGTM. Thanks for the fix. I didn't know before but looks like `*` means all files in the current directory in bash :O. Maybe the crash you're seeing related with bash somehow if you are using cli? Are you sure server crashes with that command? If so, can you send the stacktrace? Maybe we'll need to investigate it separately. 

For the scope of this PR, you're right, we need to check `ignored_commands` whether it is null or not, otherwise `raft.config get ignored-commands` will crash RedisRaft and this PR fixes it. Thank you.

tezc (approved):

bugwz:
@tezc Thanks for your reply. As you said, when we use `*`, it means to obtain all configuration items, so it will trigger the problem of `ignored-commands` configuration.

eg.
```bash
bugwz@local ~ $ redis-cli -p 6379
127.0.0.1:6379> raft.config get *
Error: Server closed the connection
```

Issue #316, opened 2022-04-04
Improve non-quorum read safety


yossigo:
Add a safety mechanism to prevent processing non-quorum reads on a new leader before it has successfully committed a new entry.

PR #326, opened 2022-04-10
Improve non-quorum read safety

tezc:
Before serving non-quorum read requests, wait until the new leader applies an entry from the current term.

Otherwise, we might process a request before replaying logs. The state machine will be in an older state. Reading from it might look like going backward in time.

Fixes https://github.com/RedisLabs/redisraft/issues/316


yossigo (approved):
@tezc Can you think of a simple test for this?

yossigo (approved):

PR #329, opened 2022-04-24
tests that will have to pass for asking to work correctly

sjpotter:
simulates requests in the middle of a migration by creating keys with a stable slot, then converting the slot to a migrating or importing slot

### DIFF
--- a/tests/integration/test_sharding.py
+++ b/tests/integration/test_sharding.py
@@ -403,7 +403,6 @@ def test_shard_group_refresh(cluster_factory):
     assert_after(check_slots, 10)
 
 
-
 def test_shard_group_no_slots(cluster):
     cluster.create(3, raft_args={
         'sharding': 'yes',
@@ -426,6 +425,7 @@ def test_shard_group_reshard_to_migrate(cluster):
 
     cluster.execute("set", "key", "value");
 
+    assert cluster.execute(
         'RAFT.SHARDGROUP', 'REPLACE',
         '2',
         '12345678901234567890123456789013',
@@ -493,4 +493,4 @@ def test_shard_group_reshard_to_import(cluster):
     assert cluster.execute("asking", "del", "key") == 1
 
     with raises(ResponseError, match="TRYAGAIN"):
-        cluster.execute("asking", "get", "key")
+        cluster.execute("asking", "get", "key1")
--- a/tests/integration/test_sharding.py
+++ b/tests/integration/test_sharding.py
@@ -419,7 +419,6 @@ def test_shard_group_reshard_to_migrate(cluster):
 
     cluster.execute("set", "key", "value");
 
-    assert cluster.execute(
         'RAFT.SHARDGROUP', 'REPLACE',
         '2',
         '12345678901234567890123456789013',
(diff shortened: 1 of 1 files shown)


sjpotter:
this adds tests that currently fail, as no ask/asking support, but should be reviewed for accuracy and completeness.

tests/integration/test_sharding.py
```diff
@@ -52,13 +52,6 @@ def test_shard_group_sanity(cluster):
     with raises(ResponseError, match='MOVED [0-9]+ 1.1.1.1:111'):
         c.set('key', 'value')
 
-    # Follower redirect straight to remote shardgroup to
```
yossigo, line 55: Why is this test no longer valid and removed?

tests/integration/test_sharding.py
```diff
@@ -397,3 +390,81 @@ def test_shard_group_no_slots(cluster):
...
+        cluster.leader_node().raft_info()['dbid'],
+        '1', '1',
+        '0', '16383', '2',
+        '1234567890123456789012345678901234567890', '2.2.2.2:2222',
+    ) == b'OK'
+
+    with raises(ResponseError, match="MOVED 12539 3.3.3.3:3333"):
+        cluster.leader_node().client.get("key")
+
+    assert cluster.leader_node().client.execute_command("asking", "get", "key") == b'value'
```
tezc, line 462:
```suggestion
    assert cluster.execute("asking", "get", "key") == b'value'
```

Btw, I think you can just call execute() for the same effect here and for other usages too.

tezc, line 462:
```suggestion
    assert cluster.execute("asking", "get", "key") == b'value'
```

Btw, I think you can just call execute() for the same effect here and for other usages too.

tests/integration/test_sharding.py
sjpotter: from slack

[Shaya Potter](https://app.slack.com/team/UM9BFGE2V)  [4:34 PM](https://redis.slack.com/archives/C027R1TTHJ6/p1648474470182639)
so this breaks a test (not the code, but a test)
[4:34](https://redis.slack.com/archives/C027R1TTHJ6/p1648474486067229)
```# Follower redirect straight to remote shardgroup to
        # avoid another redirect hop.
        cluster.wait_for_unanimity()
        cluster.node(2).wait_for_log_applied()  # to get shardgroup
        with raises(ResponseError, match='MOVED [0-9]+ 1.1.1.1:111'):
>           cluster.node(2).client.set('key', 'value')
E           AssertionError: Pattern 'MOVED [0-9]+ 1.1.1.1:111' not found in 'MOVED 0 localhost:5001'
```
[4:35](https://redis.slack.com/archives/C027R1TTHJ6/p1648474519443559)
need to explore that
[Shaya Potter](https://app.slack.com/team/UM9BFGE2V)  [4:50 PM](https://redis.slack.com/archives/C027R1TTHJ6/p1648475435841149)
previously it would be caught by handleSharding earlier and redirect to remote cluster then and never hit checkLeader() now it does.  I'm unsure if we should try to optimize it, and just remove the test?
[Ozan Tezcan](https://app.slack.com/team/U0270RYB881)  [4:55 PM](https://redis.slack.com/archives/C027R1TTHJ6/p1648475737366189)
yeah I agree, follower can just redirect to leader, leader will redirect to remote cluster anyway

sjpotter: though going through the code again, might not be necessary to remove, as the handleSharding call there was put back.

sjpotter: tried it with current code and it works, so putting back.

PR #331, opened 2022-04-25
change to computeHashSlotOrReplyError to not reply error

sjpotter:
move error returning to caller

based on feedback from @tezc

### DIFF
--- a/src/redisraft.c
+++ b/src/redisraft.c
@@ -419,10 +419,7 @@ static RRStatus handleSharding(RedisRaftCtx *rr,
     int slot;
 
     if (computeHashSlot(rr, cmds, &slot) != RR_OK) {
-        if (ctx) {
-            RedisModule_ReplyWithError(ctx, "CROSSSLOT Keys in request don't hash to the same slot");
-        }
-
+        RedisModule_ReplyWithError(ctx, "CROSSSLOT Keys in request don't hash to the same slot");
         return RR_ERROR;
     }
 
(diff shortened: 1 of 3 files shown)


src/redisraft.c
```diff
@@ -418,7 +418,11 @@ static RRStatus handleSharding(RedisRaftCtx *rr,
 {
     int slot;
 
-    if (computeHashSlotOrReplyError(rr, ctx, cmds, &slot) != RR_OK) {
+    if (computeHashSlot(rr, cmds, &slot) != RR_OK) {
+        if (ctx) {
```
tezc, line 422: At this point, we always have a `ctx`, don't we?

tezc:
New commit is a bit different than the previous commit and PR title :) Is it intentional or by mistake? Also, this pr will merge into master, just fyi.

tezc:
@sjpotter Did you see the comment above?

tezc:
@sjpotter Did you see the comment above?

src/redisraft.c
sjpotter: I guess I was thinking that it may be called by a follower, though perhaps thats impossible

sjpotter:
> @sjpotter Did you see the comment above?

removed extra check

tezc:
> New commit is a bit different than the previous commit and PR title :) Is it intentional or by mistake? Also, this pr will merge into master, just fyi.

I meant this one.

sjpotter:
I meant this to go to both.  figured this was simple enough to go straight into master.

sjpotter:
oh, and yes, that was a mistake.  that shouldn't have been here.

sjpotter:
was a rebase mistake. I rebased against wrong tree and then pushed to the pr against master

tezc (approved):

PR #332, opened 2022-04-25
parse out asking from commands sent to redis

sjpotter:
parses out asking in all places where we parse redis commands and in execute mark the command array as asking so it can be treated differently

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -216,7 +216,22 @@ unsigned int CommandSpecGetAggregateFlags(RaftRedisCommandArray *array, unsigned
 {
     unsigned int flags = 0;
     for (int i = 0; i < array->len; i++) {
-        const CommandSpec *cs = CommandSpecGet(array->commands[i]->argv[0]);
+        /* we get the flag for the command that is asking, not the asking */
+        RedisModuleString * cmd = array->commands[i]->argv[0];
+        size_t cmd_len;
+        const char * cmd_str = RedisModule_StringPtrLen(cmd, &cmd_len);
+
+        if (!strncasecmp("asking", cmd_str, cmd_len)) {
+            /* TODO: when asking <magic> support is added, need to adjust to argv[2] */
+            if (array->commands[i]->argc > 1) {
+                cmd = array->commands[i]->argv[1];
+            } else {
+                /* TODO: Is this protection necessary? */
+                continue;
+            }
+        }
+
+        const CommandSpec *cs = CommandSpecGet(cmd);
         if (cs) {
             flags |= cs->flags;
         } else {
--- a/src/commands.c
+++ b/src/commands.c
@@ -226,7 +226,7 @@ unsigned int CommandSpecGetAggregateFlags(RaftRedisCommandArray *array, unsigned
             if (array->commands[i]->argc > 1) {
                 cmd = array->commands[i]->argv[1];
             } else {
-                /* TODO: Is this protection necessary? */
+                /* necessary to protect/ignore a bare "asking" command */
                 continue;
             }
         }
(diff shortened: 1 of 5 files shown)


src/multi.c
```diff
@@ -85,8 +85,14 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
...
     const char *cmd_str = RedisModule_StringPtrLen(cmd->argv[0], &cmd_len);
+    size_t cmd1_len;
+    const char *cmd1_str = "";
+    if (cmd->argc > 1) { /* TODO: will have to be > 2 in future with magic value in asking for argv[2] */
+        cmd1_str = RedisModule_StringPtrLen(cmd->argv[1], &cmd1_len);
+    }
 
-    if (cmd_len == 5 && !strncasecmp(cmd_str, "MULTI", 5)) {
+    if ((cmd_len == 5 && !strncasecmp(cmd_str, "MULTI", 5)) ||
+       (cmd_len == 6 && !strncasecmp(cmd_str, "ASKING", 6) && cmd1_len == 5 && !strncasecmp(cmd1_str, "MULTI", 5))) {
```
tezc, line 95:
```c
    if (cmd_len == 6 && !strncasecmp(cmd_str, "ASKING", 6)) {
        if (cmd->argc > 1) {
            cmd_str = RedisModule_StringPtrLen(cmd->argv[1], &cmd_len);
        }
    }
```

Can't make a suggestion because of the deleted line but would that code above be equivalent to these changes? Checking "asking" just before "multi".

sjpotter: somewhat.  it make an "ASKING exec" command valid as well.  I guess thats ok.  the current way is a bit of an artifact of when I was storing an asking flag in the multi struct, which no longer doing, so this simplification is probably ok.  will do that.

src/raft.c
```diff
@@ -122,12 +122,14 @@ void entryAttachRaftReq(RedisRaftCtx *rr, raft_entry_t *entry, RaftReq *req)
...
 void RaftExecuteCommandArray(RedisModuleCtx *ctx,
                              RedisModuleCtx *reply_ctx,
-                             RaftRedisCommandArray *array)
+                             RaftRedisCommandArray *cmds)
 {
     int i;
 
-    for (i = 0; i < array->len; i++) {
-        RaftRedisCommand *c = array->commands[i];
+    HandleAsking(cmds);
```
tezc, line 129: If there are no other usages, I wouldn't create a function for this. Advantages: No need to declare it, no need to add a field to RaftRedisCommandArray (you can have a local bool variable here).

src/commands.c
```diff
@@ -216,7 +216,22 @@ unsigned int CommandSpecGetAggregateFlags(RaftRedisCommandArray *array, unsigned
...
+        RedisModuleString * cmd = array->commands[i]->argv[0];
+        size_t cmd_len;
+        const char * cmd_str = RedisModule_StringPtrLen(cmd, &cmd_len);
+
+        if (!strncasecmp("asking", cmd_str, cmd_len)) {
+            /* TODO: when asking <magic> support is added, need to adjust to argv[2] */
+            if (array->commands[i]->argc > 1) {
+                cmd = array->commands[i]->argv[1];
+            } else {
+                /* TODO: Is this protection necessary? */
```
tezc, line 229: I think necessary, just sending "asking" without a command from cli,  it comes till here.

src/multi.c
tezc: Yeah, we'll be ignoring any "asking" here with this change. Probably, that's okay.

src/raft.c
sjpotter: as of now, we don't do anything with the asking flag, in the future we will.  we will also have the slot variable that the commands are associated with.

i.e. like https://github.com/RedisLabs/redisraft/pull/293/files#diff-10528055f382929eb163463ff1402e7d59eb4fbe95ec1d005b52aaf2b148d2b9R479-R480

tezc: okay

sjpotter:
was a rebase mistake.  I rebased against wrong tree and then pushed to the pr against master

PR #333, opened 2022-04-26
handle migrating and importing slot/shardgroup

sjpotter:
resolves valid shardgroup to resolve with asking or not.

If the shardgroup is local and migrating

1) if all keys exist, handle normally
2) if some keys exist, return a TRYAGAIN response
3) if no keys exist, return an ASK response

if that request is asking and the local shardgroup is importing

1) if all keys eist, handle normally
2) otherwise, return TRYAGAIn


src/raft.c
```diff
@@ -115,19 +115,187 @@ void entryAttachRaftReq(RedisRaftCtx *rr, raft_entry_t *entry, RaftReq *req)
...
+    if (si->shard_groups_num != 1) {
+        return true;
+    }
+
+    size_t key_len;
+    ShardGroup *sg;
+
+    RedisModuleDictIter *iter = RedisModule_DictIteratorStartC(si->shard_group_map, "^", NULL, 0);
+    RedisModule_DictNextC(iter, &key_len, (void **) &sg);
+    RedisModule_DictIteratorStop(iter);
```
tezc, line 130: This function is in the critical path (performance-wise)  and these lines make bunch of malloc/free calls unfortunately. Probably, we can avoid this. Let's try to find something else sometime. Maybe we keep a `bool sharding` variable in ShardingInfo and update it when we apply shardgroup changes to it? Or is there any other way to do it?

sjpotter: ok.  that makes sense.  looking into it.

sjpotter: moved test into ShardingInfoAddShardGroup() so don't have to do test each time around.  With that said, in practice I'm unsure how big an issue this is.  This really only mattered when we had a single shard group, but change should take care of the issue.

src/raft.c
```diff
@@ -115,19 +115,187 @@ void entryAttachRaftReq(RedisRaftCtx *rr, raft_entry_t *entry, RaftReq *req)
...
+                }
+                return RR_ERROR;
+            case AllExist:
+            case NoKeys:
+                return RR_OK;
+        }
+    } else if (slot_type == SLOTRANGE_TYPE_IMPORTING) {
+        switch (validateKeyExistence(rr, cmds)) {
+            case SomeExist:
+            case NoneExist:
```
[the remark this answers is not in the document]
sjpotter: no, that has to be fixed.  thanks

src/raft.c
```diff
@@ -115,19 +115,170 @@ void entryAttachRaftReq(RedisRaftCtx *rr, raft_entry_t *entry, RaftReq *req)
...
+    ShardGroup *sg = GetCommandShardGroup(rr, cmds);
+    if (!sg) {
+        if (reply_ctx) {
+            RedisModule_ReplyWithError(reply_ctx, "CLUSTERDOWN Hash slot is not served");
+        }
+        return RR_ERROR;
+    }
+
+    /* this is slightly questionable.
+     * I'm keen to argue that all slots belongingto a single shardgroup have to have the same slot type
```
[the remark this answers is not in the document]
sjpotter: hmm, I was thinking (now incorrectly) that a single redisraft cluster could be part of more than one shardgroup (i.e. can be shardgroup a and shardgroup b with node x in both cases).  Doesn't work because the shardgroup id is matched against dbid.  so need to rework this.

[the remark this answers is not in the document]
sjpotter: modified code

```
    SlotRangeType slot_type = SLOTRANGE_TYPE_UNDEF;
    
    for (size_t i = 0; i < sg->slot_ranges_num; i++) {
        if (sg->slot_ranges[i].start_slot <= cmds->slot && cmds->slot <= sg->slot_ranges[i].end_slot) {
            slot_type = sg->slot_ranges[i].type;
            break;
        } 
    }
    
    if (slot_type == SLOTRANGE_TYPE_UNDEF) {
        if (reply_ctx) {
            RedisModule_ReplyWithError(reply_ctx, "ERR internal error, couldn't associate a shardgroup slot to this request");
        }
        return RR_ERROR;
    }
```

src/common.c
```diff
@@ -70,6 +70,20 @@ void replyRedirect(RedisModuleCtx *ctx, int slot, NodeAddr *addr)
...
+void replyAsk(RedisRaftCtx *rr, RedisModuleCtx *ctx, int slot)
+{
+    ShardGroup *sg = rr->sharding_info->importing_slots_map[slot];
+    if (!sg) {
+        RedisModule_ReplyWithError(ctx, "ERR no importing shard group to ask");
+        return;
+    }
+
+    char buf[sizeof(sg->nodes[0].addr.host) + 256];
+    snprintf(buf, sizeof(buf), "ASK %d %s:%u", slot, sg->nodes[0].addr.host, sg->nodes[0].addr.port);
```
sjpotter, line 83: do we want to iterate through nodes?

src/raft.c
```diff
@@ -115,19 +115,185 @@ void entryAttachRaftReq(RedisRaftCtx *rr, raft_entry_t *entry, RaftReq *req)
...
+        if (found != 0 && found != total_keys) {
+            return SomeExist;
+        }
+    }
+
+    if (found != total_keys) {
+        return (found == 0) ? NoneExist : SomeExist;
+    }
+
+    return (total_keys == 0) ? NoKeys : AllExist;
```
sjpotter, line 156: consider if NoKeys/AllExist is needed, as similiar state.

src/raft.c
```diff
@@ -115,19 +115,185 @@ void entryAttachRaftReq(RedisRaftCtx *rr, raft_entry_t *entry, RaftReq *req)
...
+
+        for (int j = 0; j < num_keys; j++) {
+            found += RedisModule_KeyExists(rr->ctx, cmd->argv[keyindex[j]]);
+        }
+        RedisModule_Free(keyindex);
+
+        /* shortcut as we know the result is now SomeExist */
+        if (found != 0 && found != total_keys) {
+            return SomeExist;
+        }
```
sjpotter, line 149: this short cut probably isn't necessary.

src/raft.c
```diff
@@ -115,19 +115,185 @@ void entryAttachRaftReq(RedisRaftCtx *rr, raft_entry_t *entry, RaftReq *req)
...
+    if (sg && sg->local) {
+        return sg;
+    }
+
+    if (cmds->asking) {
+        ShardGroup *isg = rr->sharding_info->importing_slots_map[cmds->slot];
+        if (isg && isg->local) {
+            return isg;
+        }
+    }
```
sjpotter, line 187: if the command is asking, but not meant for the local cluster, should we redirect to the "migrating" cluster or to the "importing" cluster?

src/raft.c
```diff
@@ -115,19 +115,185 @@ void entryAttachRaftReq(RedisRaftCtx *rr, raft_entry_t *entry, RaftReq *req)
...
+        return RR_ERROR;
+    }
+
+    if (!sg->local) {
+        if (reply_ctx) {
+            if (sg->next_redir >= sg->nodes_num) {
+                sg->next_redir = 0;
+            }
+
+            replyRedirect(reply_ctx, cmds->slot, &sg->nodes[0].addr);
```
sjpotter, line 231: either remove the next_redir code above or replace 0 in sg->nodes[0].addr with the next_redir value

src/raft.c
```diff
@@ -115,19 +115,185 @@ void entryAttachRaftReq(RedisRaftCtx *rr, raft_entry_t *entry, RaftReq *req)
...
+    /* Make sure hash slot is mapped and handled locally. */
+    ShardGroup *sg = GetCommandShardGroup(rr, cmds);
+    if (!sg) {
+        if (reply_ctx) {
+            RedisModule_ReplyWithError(reply_ctx, "CLUSTERDOWN Hash slot is not served");
+        }
+        return RR_ERROR;
+    }
+
+     if (cmds->slot < 0) {
```
sjpotter, line 202: figure out a better solution, perhaps RM_Assert if this is < 0 at this point?

src/raft.c
```diff
@@ -115,19 +115,185 @@ void entryAttachRaftReq(RedisRaftCtx *rr, raft_entry_t *entry, RaftReq *req)
...
+            case NoneExist:
+                if (reply_ctx) {
+                    replyAsk(rr, reply_ctx, cmds->slot);
+                }
+                return RR_ERROR;
+            case AllExist:
+            case NoKeys:
+                return RR_OK;
+        }
+    } else if (slot_type == SLOTRANGE_TYPE_IMPORTING) {
```
fadidahanna, line 254: else can be removed here, for styling consistency

src/redisraft.c
```diff
@@ -423,13 +422,15 @@ static RRStatus handleSharding(RedisRaftCtx *rr,
         return RR_ERROR;
```
sjpotter, line 422: ReplyWithError() has to be surrounded by an `if (ctx) {}`

sjpotter: and below

src/redisraft.c
```diff
@@ -413,8 +413,7 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
  * 3. If the hash slot is associated with a foreign ShardGroup, perform a redirect.
  * 4. If the hash slot is not mapped, produce a CLUSTERDOWN error.
  */
-static RRStatus handleSharding(RedisRaftCtx *rr,
-                               RedisModuleCtx *ctx, RaftRedisCommandArray *cmds)
+RRStatus handleSharding(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRedisCommandArray *cmds)
```
sjpotter, line 416: right now its more complicated than it needs to be, as its done at both append time and apply time, we should perhaps consider getting rid of the code at append time, which would simplify the function, as wouldn't need to validate shard group (sg) here) as would be validated in the `validateRaftRedisCommandArray()` function call

src/raft.c
```diff
@@ -115,19 +115,185 @@ void entryAttachRaftReq(RedisRaftCtx *rr, raft_entry_t *entry, RaftReq *req)
...
+     * hash slot validation and return an error / redirection if necessary. We do this
+     * before checkLeader() to avoid multiple redirect hops.
+     */
+    if (isSharding(rr) && handleSharding(rr, reply_ctx, cmds) != RR_OK) {
+        return;
+    }
+
+    if (cmds->slot != -1 && validateRaftRedisCommandArray(rr, reply_ctx, cmds) != RR_OK) {
+        return;
+    }
```
sjpotter, line 295: if we are removing handleSharding from append path, can combine the `validateRaftRedisCommandArray()` into `handleSharding()`

src/redisraft.h
```diff
@@ -476,6 +476,7 @@ typedef struct {
 
 typedef struct {
     bool asking;        /* if this command array is an asking */
+    int slot;           /* key slot for cluster mode, so doesn't have to be recalculated */
```
sjpotter, line 479: might not be neccessary to add to struct depending on the consolidation we do noted above

tests/integration/test_sharding.py
```diff
@@ -403,3 +396,81 @@ def test_shard_group_no_slots(cluster):
...
+
+    assert cluster.leader_node().client.execute_command("asking", "get", "key") == b'value'
+
+    with raises(ResponseError, match="TRYAGAIN"):
+        cluster.leader_node().client.execute_command("asking", "get", "key1")
+
+    assert cluster.leader_node().client.execute_command("asking", "del", "key") == 1
+
+    with raises(ResponseError, match="TRYAGAIN"):
+        cluster.leader_node().client.execute_command("asking", "get", "key1")
```
sjpotter, line 476: needs to be key, not key1 as testing a non existent (deleted above) key

tests/integration/test_sharding.py
```diff
@@ -403,3 +396,81 @@ def test_shard_group_no_slots(cluster):
     splits = results[0].split(b" ")
     assert len(splits) == 9
     assert splits[8] == b""
+
+
+def test_shard_group_reshard_to_migrate(cluster):
+    cluster.create(3, raft_args={
+        'sharding': 'yes',
```
sjpotter, line 403: all this code will have to handle asking be a separate command

src/common.c
sjpotter: @yossigo thoughts?  inside a cluster we had a mechanism for iterating through the nodes of the cluster.  I'm wondering if its as useful here.  a cluster client should have a picture of the topology anyways....

src/raft.c
sjpotter: @yossigo my guess is that it makes sense to always redirect to the migrating cluster, as that's the "primary" in the eyes of the client, but I'm open to being wrong about that gut feeling

src/raft.c
sjpotter: @yossigo thoughts about this?

src/raft.c
sjpotter: @fadidahanna don't understand?

src/raft.c
yossigo: @sjpotter I commented about this topic in a different location.

src/raft.c
yossigo: @sjpotter I think redirecting to the migrating cluster might be a bit safer.

src/common.c
yossigo: This issue was raised in different comments below.

src/raft.c
sjpotter: ok, that's what would happen with current design.

PR #335, opened 2022-04-27
Use local function inside Lua scripts

tezc:
Fixed test failure: https://github.com/RedisLabs/redisraft/runs/6191624506?check_suite_focus=true#step:9:913

```
        if isinstance(response, ResponseError):
>           raise response
E           redis.exceptions.ResponseError: user_script:15: Attempt to modify a readonly table script: f1f84d8634b7f351dc36e8132cba23b9751e47e9, on @user_script:15.
```

After https://github.com/redis/redis/pull/10651, redis does not allow non-local function definitions. Changed the function definition to local in the script.

Also, I had to change function name, for some reason lua compiler didn't like it.

### DIFF
--- a/tests/integration/test_random.py
+++ b/tests/integration/test_random.py
@@ -46,16 +46,16 @@ for i = 65,  90 do table.insert(charset, string.char(i)) end
 for i = 97, 122 do table.insert(charset, string.char(i)) end
 
 -- 2. recursive function to generate random string of length
-function string.random(length)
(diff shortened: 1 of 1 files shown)


tezc:
@yossigo Don't know anything about Lua but this looks like a breaking change, I guess it's highly likely that people have scripts with non-local function definitions. Just fyi.

yossigo (approved):

PR #337, opened 2022-04-27
implement raft.import

sjpotter:
includes a sanity test


src/migrate.c
```diff
@@ -0,0 +1,156 @@
...
+    RedisModuleString * zero = RedisModule_CreateString(rr->ctx, "0", 1);
+
+    for (size_t i = 0; i < import_keys.num_keys; i++) {
+        RedisModuleString * temp[3];
+        temp[0] = import_keys.key_names[i];
+        temp[1] = zero;
+        temp[2] = import_keys.key_serialized[i];
+
+        enterRedisModuleCall();
+        RedisModuleCallReply *reply = RedisModule_Call(rr->ctx, "restore", "v", temp, 3);
```
tezc, line 88: I think we should check error cases here, just in case.

src/redisraft.h
```diff
@@ -558,8 +561,18 @@ typedef struct ShardingInfo {
     ShardGroup *stable_slots_map[REDIS_RAFT_HASH_SLOTS];
     ShardGroup *importing_slots_map[REDIS_RAFT_HASH_SLOTS];
     ShardGroup *migrating_slots_map[REDIS_RAFT_HASH_SLOTS];
+
+    raft_term_t max_importing_term[REDIS_RAFT_HASH_SLOTS];
 } ShardingInfo;
 
+typedef struct {
+    raft_term_t term;
+    int magic;
```
tezc, line 570: I don't remember what will we use for `magic` number but we should be careful not to cause overflow (undefined behavior). I suggest using unsigned numbers for this purpose or even fixed length integers.

src/serialization.c
```diff
@@ -279,4 +331,93 @@ RRStatus RaftRedisCommandArrayDeserialize(RaftRedisCommandArray *target, const v
...
+        return RR_ERROR;
+    }
+    p += n; buf_size -= n;
+    target->num_keys = num_keys;
+
+    target->key_names = RedisModule_Calloc(num_keys, sizeof(RedisModuleString*));
+    target->key_serialized = RedisModule_Calloc(num_keys, sizeof(RedisModuleString*));
+
+    for (size_t i = 0; i < num_keys; i++) {
+        if ((n = decodeString(p, buf_size, &target->key_names[i])) < 0 || !target->key_names[i]) {
```
tezc, line 369: Just curious, does this ever execute second part of the if check: `!target->key_names[i]` ? If `decodeString()` returns negative, it won't be executed. If `decodeString()` is positive, that part is never null? Just consider if it makes sense.

src/migrate.c
sjpotter: added an assert that it didn't fail and wraped in an existence check to not restore when key already exists.

src/redisraft.h
sjpotter: magic was just a shared value for the importing/migrating shardgroup to let importing know that its getting a request from a migrating one that is on the same "round" as it.  (not from something stale for some reason).

tezc: Alternatively, we can declare `magic` as `0 <= magic <= 2^31` , still need to do check before casting:

```c
    if (magic < 0 || magic > INT32_MAX) {
        ReplyError();
    }
```

sjpotter: I just made magic take a long long to fit with redis module api.

src/serialization.c
sjpotter: looking at it again.  conceptually, one could pass in a string length of 0, which would then return an empty string which would e an error.  so the check is probably useful.

src/migrate.c
```diff
@@ -0,0 +1,166 @@
...
+    req->r.import_keys.term = (raft_term_t) term;
+    req->r.import_keys.magic = magic;
+
+    int num_keys = (argc -3) / 2;
+    req->r.import_keys.num_keys = num_keys;
+    req->r.import_keys.key_names = RedisModule_Calloc(num_keys, sizeof(RedisModuleString *));
+    req->r.import_keys.key_serialized = RedisModule_Calloc(num_keys, sizeof(RedisModuleString *));
+
+    for (int i = 0; i < num_keys; i++) {
+        req->r.import_keys.key_names[i] = RedisModule_CreateStringFromString(rr->ctx, argv[3 + (i*2)]);
```
yossigo, line 141: Consider to avoid duplicating the argv strings and just hold a refcount.

src/migrate.c
```diff
@@ -0,0 +1,166 @@
+#include <string.h>
+
+#include "redisraft.h"
+
+int validSlotMagic(RedisRaftCtx *rr, unsigned int slot, int magic)
```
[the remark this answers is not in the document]
sjpotter: ok

src/migrate.c
```diff
@@ -0,0 +1,166 @@
...
+        RaftReqFree(req);
+    }
+    FreeImportKeys(&import_keys);
+}
+
+int cmdRaftImport(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
+{
+    RedisRaftCtx *rr = &redis_raft;
+
+    if (argc < 3 || ((argc -3)) % 2 != 0) {
```
sjpotter, line 114: argc < 5 not 3.  add comment that 5 implies the 3 strings raft.import, magic, term and the next 2 are the first key/serialized combo

src/migrate.c
```diff
@@ -0,0 +1,166 @@
...
+    if (!validSlotTerm(rr, slot, import_keys.term)) {
+        // this is acceptable (suspended old leader
+        LOG_WARNING("ignoring import keys as old term");
+        if (req) {
+            RedisModule_ReplyWithError(req->ctx, "ERR invalid term");
+        }
+        goto exit;
+    }
+
+    if (!validSlotMagic(rr, slot, import_keys.magic)) {
```
sjpotter, line 72: magic verification should be before term validation

src/migrate.c
```diff
@@ -0,0 +1,166 @@
...
+        }
+        goto exit;
+    }
+
+    RedisModuleString * zero = RedisModule_CreateString(rr->ctx, "0", 1);
+
+    for (size_t i = 0; i < import_keys.num_keys; i++) {
+        RedisModuleString * temp[3];
+        temp[0] = import_keys.key_names[i];
+        temp[1] = zero;
```
sjpotter, line 84: comment that this zero is the "ttl" part of the restore operation

src/migrate.c
```diff
@@ -0,0 +1,166 @@
...
+        temp[2] = import_keys.key_serialized[i];
+
+        if (!RedisModule_KeyExists(rr->ctx, temp[0])) {
```
sjpotter, line 87: take out the KeyExists if and use "REPLACE" in the restore rm_call.

src/migrate.c
```diff
@@ -0,0 +1,166 @@
...
+        }
+    }
+
+    RedisModule_FreeString(rr->ctx, zero);
+
+    if (req) {
+        RedisModule_ReplyWithSimpleString(req->ctx, "OK");
+    }
+
+    exit:
```
sjpotter, line 103: whitespace

src/serialization.c
```diff
@@ -279,4 +331,93 @@ RRStatus RaftRedisCommandArrayDeserialize(RaftRedisCommandArray *target, const v
...
+        n = encodeString(p, sz, import_keys->key_names[i]);
+        assert (n != -1);
+        p += n; sz -= n;
+        n = encodeString(p, sz, import_keys->key_serialized[i]);
+        assert (n != -1);
+        p += n; sz -= n;
+    }
+
+    return ety;
+}
```
sjpotter, line 423: new line

src/commands.c
```diff
@@ -215,7 +215,22 @@ unsigned int CommandSpecGetAggregateFlags(RaftRedisCommandArray *array, unsigned
...
+        size_t cmd_len;
+        const char * cmd_str = RedisModule_StringPtrLen(cmd, &cmd_len);
+
+        if (!strncasecmp("asking", cmd_str, cmd_len)) {
+            /* TODO: when asking <magic> support is added, need to adjust to argv[2] */
+            if (array->commands[i]->argc > 1) {
+                cmd = array->commands[i]->argv[1];
+            } else {
+                /* necessary to protect/ignore a bare "asking" command */
+                continue;
```
yossigo, line 229: @sjpotter So fixing `ASKING` to be compatible with Redis Cluster Spec is in another PR?

src/migrate.c
```diff
@@ -0,0 +1,171 @@
...
+
+    if (!validSlotMagic(rr, slot, import_keys.magic)) {
+        if (req) {
+            RedisModule_ReplyWithError(req->ctx, "ERR invalid magic");
+        }
+        goto exit;
+    }
+
+    if (!validSlotTerm(rr, slot, import_keys.term)) {
+        // this is acceptable (suspended old leader
```
yossigo, line 71:
```suggestion
        // this is acceptable (suspended old leader)
```

src/migrate.c
```diff
@@ -0,0 +1,171 @@
...
+        LOG_WARNING("ignoring import keys as old term");
```
yossigo, line 72: I think this kind of logging is problematic as it can blow up logs very quickly:
* Consider making this a debug message.
* Consider having a counter for this, i.e. count the number of import keys ignored for different reasons. Note that we could try to leverage the fact that Redis is tracking different errors (`INFO errorstats`).

src/migrate.c
```diff
@@ -0,0 +1,171 @@
...
+        if (req) {
+            RedisModule_ReplyWithError(req->ctx, "ERR invalid term");
+        }
+        goto exit;
+    }
+
+    /* 2  'static' strings need for restore command */
+    RedisModuleString *zero = RedisModule_CreateString(rr->ctx, "0", 1); /* ttl of zero */
```
yossigo, line 80: How about a shared pool of static strings, like Redis does?

src/migrate.c
```diff
@@ -0,0 +1,171 @@
...
+    for (size_t i = 0; i < import_keys.num_keys; i++) {
+        RedisModuleString * temp[4];
+        temp[0] = import_keys.key_names[i];
+        temp[1] = zero;
+        temp[2] = import_keys.key_serialized[i];
+        temp[3] = replace;
+
+        enterRedisModuleCall();
+        RedisModuleCallReply *reply;
+        RedisModule_Assert((reply = RedisModule_Call(rr->ctx, "restore", "v", temp, 3)) != NULL);
```
yossigo, line 92: I think it's good practice to separate actual statements from assert, since theoretically asserts can be filtered out (even if not currently supported by `RM_Assert`). Saw this in several places.

src/serialization.c
```diff
@@ -115,6 +122,40 @@ static size_t calcSerializedSize(RaftRedisCommand *cmd)
     return sz;
 }
 
+static int decodeInteger(const char *ptr, size_t sz, char expect_prefix, size_t *val)
```
yossigo, line 125: It's not clear why we need this function and how exactly it behaves, I suggest to document its interface (same applies for other functions in this file, maybe it's time to LIBTYFI).

src/migrate.c
sjpotter: this goes to a different Q, what error checking / validation should we do around RM_Calls().

i.e. RM_Call should never return a null reply if we give it a valid command, right?  we would always have the command itself error out on us.  And how much checking of the output command do we have to do.  in soe cases I'm validating all parts of the output when we get what is essentially a data structure back (array of arrays...)

src/migrate.c
sjpotter: documented as https://github.com/RedisLabs/redisraft/issues/357

src/migrate.c
sjpotter: documented as https://github.com/RedisLabs/redisraft/issues/358

src/migrate.c
sjpotter: made an issue for this

PR #338, opened 2022-04-28
migrate command

sjpotter:
1) parses migrate command and creates an entry committed to raft log
2) adds raft log entries to lock keys and unlock/delete keys

the next PR does the actual migration with the raft.import command to the remote cluster.

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -23,8 +23,11 @@
  * However if the key contains the {...} pattern, only the part between
  * { and } is hashed. This may be useful in the future to force certain
  * keys to be in the same node (assuming no resharding is in progress). */
-unsigned int keyHashSlot(const char *key, int keylen) {
-    int s, e; /* start-end indexes of { and } */
+unsigned int keyHashSlot(RedisModuleString * str) {
+    size_t keylen;
+    const char * key = RedisModule_StringPtrLen(str, &keylen);
+
+    size_t s, e; /* start-end indexes of { and } */
 
     for (s = 0; s < keylen; s++)
         if (key[s] == '{') break;
@@ -1200,9 +1203,9 @@ RRStatus computeHashSlot(RedisRaftCtx *rr,
         int num_keys = 0;
         int *keyindex = RedisModule_GetCommandKeys(rr->ctx, cmd->argv, cmd->argc, &num_keys);
         for (int j = 0; j < num_keys; j++) {
-            size_t key_len;
-            const char *key = RedisModule_StringPtrLen(cmd->argv[keyindex[j]], &key_len);
-            int thisslot = (int) keyHashSlot(key, (int) key_len);
+            RedisModuleString *key = cmd->argv[keyindex[j]];
+
+            int thisslot = (int) keyHashSlot(key);
 
             if (*slot == -1) {
                 /* First key */
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -23,9 +23,10 @@
  * However if the key contains the {...} pattern, only the part between
  * { and } is hashed. This may be useful in the future to force certain
  * keys to be in the same node (assuming no resharding is in progress). */
-unsigned int keyHashSlot(RedisModuleString * str) {
+unsigned int keyHashSlot(RedisModuleString *str)
+{
     size_t keylen;
-    const char * key = RedisModule_StringPtrLen(str, &keylen);
+    const char *key = RedisModule_StringPtrLen(str, &keylen);
 
     size_t s, e; /* start-end indexes of { and } */
 
@@ -784,7 +785,7 @@ ShardGroup *getShardGroupById(RedisRaftCtx *rr, const char *id)
 {
     ShardingInfo *si = rr->sharding_info;
 
-    return RedisModule_DictGetC(si->shard_group_map, (char *) id, strlen(id), NULL);
+    return RedisModule_DictGetC(si->shard_group_map, (void *) id, strlen(id), NULL);
 }
 
 /* Update an existing ShardGroup in the active ShardingInfo.
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -47,14 +47,6 @@ unsigned int keyHashSlot(RedisModuleString * str) {
     return crc16_ccitt(key+s+1,e-s-1) & 0x3FFF;
 }
 
-unsigned int KeyHashSlotRedisString(RedisModuleString *s)
-{
-    size_t key_len;
-    const char *key_str = RedisModule_StringPtrLen(s, &key_len);
-    return keyHashSlot(key_str, (int) key_len);
-}
-
-
 /* -----------------------------------------------------------------------------
  * ShardGroup Handling
  * -------------------------------------------------------------------------- */
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -785,11 +785,11 @@ RRStatus ShardingInfoValidateShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
     return RR_OK;
 }
 
-static ShardGroup *getShardGroupById(RedisRaftCtx *rr, char *id)
+ShardGroup *getShardGroupById(RedisRaftCtx *rr, const char *id)
 {
     ShardingInfo *si = rr->sharding_info;
 
-    return RedisModule_DictGetC(si->shard_group_map, id, strlen(id), NULL);
+    return RedisModule_DictGetC(si->shard_group_map, (char *) id, strlen(id), NULL);
 }
 
 /* Update an existing ShardGroup in the active ShardingInfo.
--- a/src/migrate.c
+++ b/src/migrate.c
@@ -267,7 +267,7 @@ static void transferKeys(Connection *conn)
     argv_len[1] = n;
     argv[2] = RedisModule_Alloc(32);
     // FIXME needs to be taken from sg (in raft.import pr))
-    n = snprintf(argv[1], 32, "%u", 0);
+    n = snprintf(argv[1], 32, "%llu", (long long unsigned) 0);
     argv_len[2] = n;
 
     for (size_t i = 0; i < req->r.migrate_keys.num_keys; i++) {
--- a/src/migrate.c
+++ b/src/migrate.c
@@ -176,7 +176,7 @@ fail:
     return REDISMODULE_OK;
 }
 
-void raftAppendRaftDeleteEntry(RedisRaftCtx *rr, RaftReq *req)
+static void raftAppendRaftUnlockDeleteEntry(RedisRaftCtx *rr, RaftReq *req)
 {
     raft_entry_resp_t response;
 
@@ -187,10 +187,11 @@ void raftAppendRaftDeleteEntry(RedisRaftCtx *rr, RaftReq *req)
 
     int e = raft_recv_entry(rr->raft, entry, &response);
     if (e != 0) {
+        RedisModule_ReplyWithError(req->ctx, "ERR Unable to unlock/delete migrated keys, try again");
         replyRaftError(req->ctx, e);
         entryDetachRaftReq(rr, entry);
         raft_entry_release(entry);
-        goto exit;
+        goto error;
     }
 
     raft_entry_release(entry);
@@ -200,13 +201,12 @@ void raftAppendRaftDeleteEntry(RedisRaftCtx *rr, RaftReq *req)
      */
     return;
 
-    exit:
+error:
     RaftReqFree(req);
 }
 
 static void transferKeysResponse(redisAsyncContext *c, void *r, void *privdata)
 {
-    LOG_WARNING("calling transferKeysResponse");
     Connection *conn = privdata;
     JoinLinkState *state = ConnGetPrivateData(conn);
     RedisRaftCtx *rr = ConnGetRedisRaftCtx(conn);
@@ -215,24 +215,22 @@ static void transferKeysResponse(redisAsyncContext *c, void *r, void *privdata)
     redisReply *reply = r;
 
     if (!reply) {
-        LOG_WARNING("RAFT.IMPORT failed: connection dropped.");
         ConnMarkDisconnected(conn);
+        RedisModule_ReplyWithError(req->ctx, "ERR: connection dropped impporting keys into remote cluster, try again");
+        RaftReqFree(req);
     } else if (reply->type == REDIS_REPLY_ERROR) {
         ConnAsyncTerminate(conn);
-
-        LOG_WARNING("RAFT.IMPORT failed: %.*s", (int) reply->len, reply->str);
-        RedisModule_ReplyWithError(req->ctx, "ERR: Migrate failed importing keys into remote cluster, try again");
+        replyWithFormatErrorString(req->ctx, "RAFT.IMPORT failed: %.*s", (int) reply->len, reply->str);
         RaftReqFree(req);
     } else if (reply->type != REDIS_REPLY_STATUS || reply->len != 2 || strncmp(reply->str, "OK", 2)) {
         ConnAsyncTerminate(conn);
-
         /* FIXME: above should be changed to string eventually? */
-        LOG_WARNING("RAFT.IMPORT unexpected response: type = %d (wanted %d), len = %ld, response = %.*s", reply->type, REDIS_REPLY_STATUS, reply->len, (int) reply->len, reply->str);
-        RedisModule_ReplyWithError(req->ctx, "ERR: received unexpected response from remote cluster, see logs");
+        replyWithFormatErrorString(req->ctx, "ERR: received unexpected response from remote cluster, type = %d (wanted %d), len = %ld, response = %.*s", reply->type, REDIS_REPLY_STATUS, reply->len, (int) reply->len, reply->str);
         RaftReqFree(req);
     } else {
+        /* SUCCESS */
         ConnAsyncTerminate(conn);
-        raftAppendRaftDeleteEntry(rr, req);
+        raftAppendRaftUnlockDeleteEntry(rr, req);
     }
 
     redisAsyncDisconnect(c);
@@ -240,7 +238,6 @@ static void transferKeysResponse(redisAsyncContext *c, void *r, void *privdata)
 
 static void transferKeys(Connection *conn)
 {
-    LOG_WARNING("calling transferKeys");
     RedisRaftCtx *rr = ConnGetRedisRaftCtx(conn);
     JoinLinkState *state = ConnGetPrivateData(conn);
     RaftReq *req = state->req;
@@ -288,8 +285,10 @@ static void transferKeys(Connection *conn)
     }
 
     if (redisAsyncCommandArgv(ConnGetRedisCtx(conn), transferKeysResponse, conn, argc, (const char **) argv, argv_len) != REDIS_OK) {
+        RedisModule_ReplyWithError(req->ctx, "ERR failed to submit RAFT.IMPORT command, try again");
         redisAsyncDisconnect(ConnGetRedisCtx(conn));
         ConnMarkDisconnected(conn);
+        RaftReqFree(req);
     }
 
     for(int i = 0; i < argc; i++) {
@@ -304,7 +303,7 @@ void MigrateKeys(RedisRaftCtx *rr, RaftReq *req)
     JoinLinkState *state = RedisModule_Calloc(1, sizeof(*state));
     state->type = "migrate";
     state->connect_callback = transferKeys;
-    time(&(state->start));
+    state->start = time(NULL);
     ShardGroup * sg = getShardGroupById(rr, req->r.migrate_keys.shardGroupId);
     if (sg == NULL) {
         RedisModule_ReplyWithError(req->ctx, "ERR couldn't resolve shardgroup id");
--- a/src/migrate.c
+++ b/src/migrate.c
@@ -256,46 +256,32 @@ static void transferKeys(Connection *conn)
     }
 
     // raft.import term magic <key1_name> <key1_serialized> ... <keyn_name> <keyn_serialized>
-    int argc = 3 + (req->r.migrate_keys.num_keys * 2);
+    int argc = 3 + (req->r.migrate_keys.num_serialized_keys * 2);
     char **argv = RedisModule_Calloc(argc, sizeof(char *));
     size_t *argv_len = RedisModule_Calloc(argc, sizeof(size_t));
 
     argv[0] = RedisModule_Strdup("RAFT.IMPORT");
     argv_len[0] = strlen("RAFT.IMPORT");
     argv[1] = RedisModule_Alloc(32);
-    int n = snprintf(argv[1], 64, "%ld", raft_get_current_term(rr->raft));
+    int n = snprintf(argv[1], 64, "%ld", req->r.migrate_keys.migrate_term);
     argv_len[1] = n;
     argv[2] = RedisModule_Alloc(32);
 
     for (size_t i = 0; i < req->r.migrate_keys.num_keys; i++) {
+        if (req->r.migrate_keys.keys_serialized[i] == NULL) {
+            continue;
+        }
+
         size_t key_len;
         const char * key = RedisModule_StringPtrLen(req->r.migrate_keys.keys[i], &key_len);
         argv[3 + (i*2)] = RedisModule_Strdup(key);
         argv_len[3+ (i*2)] = key_len;
 
-        enterRedisModuleCall();
-        RedisModuleCallReply * reply = RedisModule_Call(rr->ctx, "DUMP", "c", key);
-        exitRedisModuleCall();
-
-        if (reply && RedisModule_CallReplyType(reply) == REDISMODULE_REPLY_STRING) {
-            size_t str_len;
-            const char * str = RedisModule_CallReplyStringPtr(reply, &str_len);
-            argv[3 + (i*2) + 1] = RedisModule_Alloc(str_len);
-            memcpy(argv[3 + (i*2) + 1], str, str_len);
-            argv_len[3 + (i*2) + 1] = str_len;
-            RedisModule_FreeCallReply(reply);
-        } else {
-            if (reply) {
-                LOG_WARNING("unexpected response type = %d", RedisModule_CallReplyType(reply));
-                RedisModule_FreeCallReply(reply);
-            } else {
-                LOG_WARNING("didn't get a reply!");
-            }
-            ConnAsyncTerminate(conn);
-            RedisModule_ReplyWithError(req->ctx, "ERR see logs");
-            RaftReqFree(req);
-            goto exit;
-        }
+        size_t str_len;
+        const char * str = RedisModule_StringPtrLen(req->r.migrate_keys.keys_serialized[i], &str_len);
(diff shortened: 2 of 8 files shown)


src/migrate.c
```diff
@@ -0,0 +1,161 @@
...
+        return;
+    }
+
+    for (unsigned int i = 0; i < sg->nodes_num; i++) {
+        LOG_WARNING("MigrateKeys: adding %s:%d", sg->nodes[i].addr.host, sg->nodes[i].addr.port);
+        NodeAddrListAddElement(&state->addr, &sg->nodes[i].addr);
+    }
+    state->req = req;
+
+    state->conn = ConnCreate(rr, state, joinLinkIdleCallback, joinLinkFreeCallback);
```
tezc, line 160: No need to do right now but probably it is a good idea to keep a connection open until migration is completed. Currently, a timer kicks in every 100 millisecond and establish the connection (this will probably take 1-2 milliseconds on cloud). If we are going to process thousands of migrate commands, then it'll prolong migration time significantly.

src/raft.c
```diff
@@ -177,6 +179,62 @@ void RaftExecuteCommandArray(RedisModuleCtx *ctx,
...
+
+    RaftReq *req = entry->user_data;
+
+    size_t num_keys;
+    RedisModuleString ** keys;
+
+    keys = RaftRedisLockKeysDeserialize(entry->data, entry->data_len, &num_keys);
+
+    enterRedisModuleCall();
+    RedisModuleCallReply *reply = RedisModule_Call(redis_raft.ctx, "del", "v", keys, num_keys);
```
tezc, line 222: Maybe we can assert() here, just to verify command has been executed.

src/redisraft.c
```diff
@@ -370,6 +370,161 @@ static void handleInfoCommand(RedisRaftCtx *rr,
...
+    RaftReq *req = NULL;
+
+    // 0. assert minimum size here
+    if (cmd->argc < 8) {
+        RedisModule_WrongArity(ctx);
+        return NULL;
+    }
+
+    // 1. Extract migration info and store it in a dict rr->migrate_info
+    MigrationInfo * migrationInfo = extractMigrationInfo(rr, cmd);
```
tezc, line 455: migrationInfo will be leaked in error cases. Btw, as far as I can see, we don't need MigrationInfo at all. Probably you can just parse here and avoid handling another object. Something like:

```c
static RaftReq *cmdToMigrate(RedisRaftCtx *rr, RedisModuleCtx * ctx, RaftRedisCommand *cmd)
{
    RaftReq *req = NULL;
    char username[256] = {0};
    char password[256] = {0};
    char shard_group_id[RAFT_DBID_LEN + 1] = {0};

    // 0. assert minimum size here
    if (cmd->argc < 8) {
        RedisModule_WrongArity(ctx);
        return NULL;
    }

    size_t dbid_len;
    const char *dbid = RedisModule_StringPtrLen(cmd->argv[1], &dbid_len);
    if (dbid_len > RAFT_DBID_LEN) {
        RedisModule_ReplyWithError(ctx, "ERR shardgroup id is too long");
        return NULL;
    }
    strncpy(shard_group_id, dbid, dbid_len);

    for (int idx = 6; idx < cmd->argc; idx++) {
        size_t auth_len;
        const char *auth = RedisModule_StringPtrLen(cmd->argv[idx], &auth_len);
        if (auth_len == 5 && !strncasecmp("AUTH2", auth, 5)) {
            if (idx + 2 > cmd->argc) {
                RedisModule_ReplyWithError(ctx, "ERR couldn't parse username");
                return NULL;
            }

            size_t len;
            const char *ptr;

            ptr = RedisModule_StringPtrLen(cmd->argv[idx + 1], &len);
            if (len > 255) {
                RedisModule_ReplyWithError(ctx, "ERR username is too long");
                return NULL;
            }
            memcpy(username, ptr, len);

            ptr = RedisModule_StringPtrLen(cmd->argv[idx + 2], &len);
            if (len > 255) {
                RedisModule_ReplyWithError(ctx, "ERR password is too long");
                return NULL;
            }
            memcpy(password, ptr, len);
            break;
        } else if (auth_len == 4 && !strncasecmp("AUTH", auth, 4)) {
            if (idx + 1 > cmd->argc) {
                RedisModule_ReplyWithError(ctx, "ERR couldn't parse username");
                return NULL;
            }

            size_t len;
            const char *ptr = RedisModule_StringPtrLen(cmd->argv[idx + 1], &len);
            if (len > 255) {
                RedisModule_ReplyWithError(ctx, "ERR username is too long");
                return NULL;
            }
            memcpy(username, ptr, len);
            break;
        }
    }

    // 2. assert single key is empty
    size_t key_len;
    RedisModule_StringPtrLen(cmd->argv[3], &key_len);
    if (key_len != 0) {
        RedisModule_ReplyWithError(ctx, "ERR RedisRaft doesn't support old style single key migration");
        return NULL;
    }

    /* find "keys" for parsing out keys */
    int idx = 6;
    for(; idx < cmd->argc; idx++) {
        size_t str_len;
        const char * str = RedisModule_StringPtrLen(cmd->argv[idx], &str_len);
        if (str_len == 4 && !strcasecmp("keys", str)) {
            break;
        }
    }

    /* move idx past "keys" argv */
    idx++;

    if (idx >= cmd->argc) {
        RedisModule_ReplyWithError(ctx, "ERR Didn't provide any keys to migrate");
        return NULL;
    }

    RedisModuleString **keys = RedisModule_Alloc(sizeof(RedisModuleString *) * cmd->argc - idx);
    int num_keys = cmd->argc - idx;
    for (int i = 0; i < num_keys; i++) {
        keys[i] = RedisModule_CreateStringFromString(ctx, cmd->argv[idx + i]);
    }

    req = RaftReqInit(ctx, RR_MIGRATE_KEYS);

    /* Overwrite with new data */
    req->r.migrate_keys.num_keys = num_keys;
    req->r.migrate_keys.keys = keys;
    memcpy(req->r.migrate_keys.shardGroupId, shard_group_id, RAFT_DBID_LEN);
    memcpy(req->r.migrate_keys.auth_username, username, 256);
    memcpy(req->r.migrate_keys.auth_password, password, 256);

    return req;
}

```

src/migrate.c
sjpotter: ok, would have to figure out how to persist it.

src/raft.c
sjpotter: yes, that's probably a good idea.  it shouldn't be able to error, but it can fail to execute and get a null response.

src/redisraft.c
sjpotter: and re migrationInfo leak, yeah, the return NULL was supposed to be a goto exit in my head and just never did that.  done now.

src/redisraft.c
```diff
@@ -370,6 +370,161 @@ static void handleInfoCommand(RedisRaftCtx *rr,
...
+static MigrationInfo * extractMigrationInfo(RedisRaftCtx *rr, RaftRedisCommand *cmd)
+{
+    MigrationInfo * ret = RedisModule_Calloc(1, sizeof(MigrationInfo));
+    size_t str_len;
+    const char * str = RedisModule_StringPtrLen(cmd->argv[1], &str_len);
+    if (str_len > RAFT_DBID_LEN) {
+        LOG_WARNING("shardgroup id is too long");
+        goto fail;
+    }
+    if (getShardGroupById(rr, str) == NULL) {
```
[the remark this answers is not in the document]
sjpotter: don't understand the Q.  here we are validating the parameters.  so what has happened.

we've made ourselves migrating, and another shardgroup importing.  we give the migrate command the import shardgroup id.  we validate that at this point in time in state machine we know about this id.  if we don't, we fail early.  unsure there's anything really bad about that.

src/migrate.c
```diff
@@ -0,0 +1,161 @@
...
+    argv[2] = RedisModule_Alloc(32);
+
+    for (size_t i = 0; i < req->r.migrate_keys.num_keys; i++) {
+        size_t key_len;
+        const char * key = RedisModule_StringPtrLen(req->r.migrate_keys.keys[i], &key_len);
+        argv[3 + (i*2)] = RedisModule_Strdup(key);
+        argv_len[3+ (i*2)] = key_len;
+
+        enterRedisModuleCall();
+        RedisModuleCallReply * reply = RedisModule_Call(rr->ctx, "DUMP", "c", key);
```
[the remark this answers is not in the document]
sjpotter: ok, that's reasonable.  right now, its called by the connection callback, we will set it up before the connection async stuff is setup and store it in the req.

src/redisraft.c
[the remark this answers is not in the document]
tezc: Why do we have to lock? When we are about to apply, we can check this condition there and if something is wrong, we don't lock the keys. Again, here we are reading state machine out of raft logic, this check is not reliable or deterministic anyway.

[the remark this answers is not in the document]
sjpotter: I dont understand the Q.  are you saying that just because the id exits here, it might not exist later?

yes.  that's true.  It can possibly have a false negative failure here.  I'm not sure why that would be a problem (as long as we also ensure its valid later and not rely on it being valid at this point in time).

If you really find this to be a code smell, I can refactor it.

or am I misunderstanding the Q?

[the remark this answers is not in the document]
tezc: 1- I'm just against reading state machine map directly, as always :) . `getShardGroupById()` does that internally, accesses state machine without going through log. So, I just wonder if there is a good reason to do it here. If this is an important check, we can also serialize `shard group id` and do that check in the `lockKeys()` as the first thing. That was my suggestion. 

This check here can return false negative and maybe false positive in rare conditions? A stale request can pass this check if the state machine is not advanced enough? So, is this going to be a problem then? If the answer is yes, we have to handle this inside state machine anyway. No? I just think that these kind of checks add complexity and they are harder to reason about. It'll also make debugging/testing harder as it relies on timing. So, just want to be sure we are doing the correct thing. 

2- Slightly related, I realized that we don't have any protection against stale migrate commands. Assume that client sends migrate command, RedisRaft completes the migration but client gets disconnected before getting the ack. Client sends the same command again and we lock the keys without checking them. 
- If the command locks non-existing keys, these keys will be leaked as they will be in the map forever.
- If the command locks the keys that we are not migrating, that's another problem I guess. 

So, do we need some sort of validation here? Slot validation and we need to check key existence? My assumption was it is always okay to lock the keys even a command fails to complete the operation because the whole operation will be retried and these keys will be moved eventually. So, I guess we need to make sure we lock the keys that we will actually move.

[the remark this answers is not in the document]
tezc: I thought I posted twice yesterday, deleted one comment. Now I see, I deleted both probably or github is having hard time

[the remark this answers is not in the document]
tezc: @sjpotter Yes, personally I like 3 better. Just to make everything deterministic again. Also, I realized something else, just take a look at question 2. Probably, we need some sort of validation there? Wdyt?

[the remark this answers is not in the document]
sjpotter: yes, this code wasn't as designed.

```
    for (size_t i = 0; i < num_keys; i++) {
        size_t str_len;
        const char * str = RedisModule_StringPtrLen(keys[i], &str_len);
        LOG_WARNING("locking %.*s", (int) str_len, str);
        RedisModule_DictSet(rr->locked_keys, keys[i], NULL);
        RedisModule_FreeString(rr->ctx, keys[i]);
    }
```

changing to

```
    for (size_t i = 0; i < num_keys; i++) {
        RedisModuleString * key = RedisModule_StringPtrLen(keys[i];
        if (RedisModule_KeyExists(rr->ctx, key)) {
            size_t str_len;
            const char *str = RedisModule_StringPtrLen(key, &str_len);
            LOG_WARNING("locking %.*s", (int) str_len, str);
            RedisModule_DictSet(rr->locked_keys, keys[i], NULL);
        }
        RedisModule_FreeString(rr->ctx, key);
    }
```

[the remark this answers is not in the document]
sjpotter: I have some more sanity checks I have to put into the state machine operation.  going over it now.

src/redisraft.c
```diff
@@ -370,6 +370,152 @@ static void handleInfoCommand(RedisRaftCtx *rr,
...
+
+    raft_entry_resp_t response;
+    raft_entry_t *entry = RaftRedisLockKeysSerialize(req->r.migrate_keys.keys, req->r.migrate_keys.num_keys);
+    entry->id = rand();
+    entry->type = RAFT_LOGTYPE_LOCK_KEYS;
+    entryAttachRaftReq(rr, entry, req);
+    int e = raft_recv_entry(rr->raft, entry, &response);
+    if (e != 0) {
+        replyRaftError(req->ctx, e);
+        entryDetachRaftReq(rr, entry);
```
tezc, line 513:
```suggestion
        entryDetachRaftReq(rr, entry);
        raft_entry_release(entry);
        RaftReqFree(req);
        return;
```

Probably, we need to free RaftReq here.

sjpotter: yes, that seems sort of right, but wondering if the raft_entry_releaes/return needs to be there, or just use common code.  an asthetic taste and am leaning to what you did, but thinking about it.

tezc: Yes, you're right, looks like it's not needed here. RaftReqFree() is enough.

CMakeLists.txt
```diff
@@ -167,6 +167,7 @@ add_library(redisraft MODULE
         src/fsync.c
         src/join.c
         src/log.c
+	src/migrate.c
```
sjpotter, line 170: whitespace

CMakeLists.txt
```diff
@@ -219,6 +220,7 @@ add_executable(main
...
```
sjpotter, line 223: whitespace

src/cluster.c
```diff
@@ -23,8 +23,11 @@
  * However if the key contains the {...} pattern, only the part between
  * { and } is hashed. This may be useful in the future to force certain
  * keys to be in the same node (assuming no resharding is in progress). */
-unsigned int keyHashSlot(const char *key, int keylen) {
-    int s, e; /* start-end indexes of { and } */
+unsigned int keyHashSlot(RedisModuleString * str) {
```
sjpotter, line 26: `*str` curley

src/cluster.c
```diff
@@ -23,8 +23,11 @@
...
+    size_t keylen;
+    const char * key = RedisModule_StringPtrLen(str, &keylen);
```
sjpotter, line 28: `*key`

src/cluster.c
```diff
@@ -777,11 +780,11 @@ RRStatus ShardingInfoValidateShardGroup(RedisRaftCtx *rr, ShardGroup *new_sg)
     return RR_OK;
 }
 
-static ShardGroup *getShardGroupById(RedisRaftCtx *rr, char *id)
+ShardGroup *getShardGroupById(RedisRaftCtx *rr, const char *id)
 {
     ShardingInfo *si = rr->sharding_info;
 
-    return RedisModule_DictGetC(si->shard_group_map, id, strlen(id), NULL);
+    return RedisModule_DictGetC(si->shard_group_map, (char *) id, strlen(id), NULL);
```
sjpotter, line 787: perhaps change the cast to void *

src/raft.c
```diff
@@ -177,6 +179,118 @@ void RaftExecuteCommandArray(RedisModuleCtx *ctx,
     }
 }
 
+unsigned int keyHashSlot(RedisModuleString * str);
```
fadidahanna, line 182: please fix all the unnecessary whitespaces over the PR. 🙏

fadidahanna, line 182:
```suggestion
unsigned int keyHashSlot(RedisModuleString *str);
```

src/redisraft.c
```diff
@@ -370,6 +370,153 @@ static void handleInfoCommand(RedisRaftCtx *rr,
...
+
+            size_t password_len;
+            const char * password = RedisModule_StringPtrLen(cmd->argv[idx+1], &password_len);
+            if (password_len > 255) {
+                LOG_WARNING("username is too long: limited to 255 characters");
+                goto fail;
+            }
+            strncpy(ret->password, password, password_len);
+            break;
+        }
```
sjpotter, line 425: do we need to support "auth" or is supporting just auth2 ok?

src/redisraft.c
```diff
@@ -370,6 +370,153 @@ static void handleInfoCommand(RedisRaftCtx *rr,
...
+
+    if (idx >= cmd->argc) {
+        RedisModule_ReplyWithError(ctx, "ERR Didn't provide any keys to migrate");
+        goto exit;
+    }
+
+    RedisModuleString **keys = RedisModule_Alloc(sizeof(RedisModuleString *) * (cmd->argc - idx));
+    int num_keys = cmd->argc - idx;
+    for (int i = 0; i < num_keys; i++) {
+        keys[i] = RedisModule_CreateStringFromString(ctx, cmd->argv[idx + i]);
```
sjpotter, line 481: use HoldString instead

src/redisraft.c
```diff
@@ -370,6 +370,153 @@ static void handleInfoCommand(RedisRaftCtx *rr,
...
+    // 2. assert single key is empty
+    size_t key_len;
+    RedisModule_StringPtrLen(cmd->argv[3], &key_len);
+    if (key_len != 0) {
+        RedisModule_ReplyWithError(ctx, "ERR RedisRaft doesn't support old style single key migration");
+        goto exit;
+    }
+
+    /* find "keys" for parsing out keys */
+    int idx = 6;
```
sjpotter, line 461: document the reason for the 6 or just start at 0 and do extra string comparisons

src/raft.c
```diff
@@ -177,6 +179,118 @@ void RaftExecuteCommandArray(RedisModuleCtx *ctx,
...
+
+static void lockKeys(RedisRaftCtx *rr, raft_entry_t *entry)
+{
+    assert(entry->type == RAFT_LOGTYPE_LOCK_KEYS);
```
sjpotter, line 186: RM_Assert

src/raft.c
```diff
@@ -177,6 +179,118 @@ void RaftExecuteCommandArray(RedisModuleCtx *ctx,
...
+
+    RaftReq *req = entry->user_data;
+
+    /* FIXME: can optimize this for leader by getting it out of the req, keeping code simple for now */
+    size_t num_keys;
+    RedisModuleString ** keys = RaftRedisLockKeysDeserialize(entry->data, entry->data_len, &num_keys);
```
sjpotter, line 192: add ticket to optimize this to not have to deserialize from leader / node that has req.

src/raft.c
```diff
@@ -177,6 +179,118 @@ void RaftExecuteCommandArray(RedisModuleCtx *ctx,
...
+    for (size_t i = 0; i < num_keys; i++) {
+        RedisModuleString * key = keys[i];
+
+        unsigned int thisslot = keyHashSlot(key);
+        if (slot == -1) {
+            slot = (int) thisslot;
+        } else {
+            if (slot != (int) thisslot) {
+                if (req) {
+                    RedisModule_ReplyWithError(req->ctx, "ERR keys don't all belong to same slot");
```
sjpotter, line 205: CROSSLOT error here instead of my own

src/raft.c
```diff
@@ -177,6 +179,118 @@ void RaftExecuteCommandArray(RedisModuleCtx *ctx,
...
+        if (req) {
+            RedisModule_ReplyWithSimpleString(req->ctx, "OK");
+        }
+        goto error;
+    }
+
+    ShardingInfo *si = rr->sharding_info;
+    if (!si->migrating_slots_map[slot] || !si->migrating_slots_map[slot]->local) {
+        if (req) {
+            RedisModule_ReplyWithError(req->ctx, "ERR not migratable keys");
```
sjpotter, line 222: perhaps a better error response?

src/raft.c
```diff
@@ -177,6 +179,118 @@ void RaftExecuteCommandArray(RedisModuleCtx *ctx,
...
+        }
+    }
+
+    if (req) {
+        memcpy(req->r.migrate_keys.shardGroupId, si->importing_slots_map[slot]->id, RAFT_DBID_LEN);
+        /* currently, we just return, but eventually this is where should kick off migration of the keys */
+        RedisModule_ReplyWithSimpleString(req->ctx, "OK");
+        RaftReqFree(req);
+//        MigrateKeys(rr, req);
+        goto exit;
```
sjpotter, line 250: should be after the `}` before the error label

src/raft.c
```diff
@@ -177,6 +179,118 @@ void RaftExecuteCommandArray(RedisModuleCtx *ctx,
...
+            RedisModule_DictSet(rr->locked_keys, keys[i], NULL);
+        }
+    }
+
+    if (req) {
+        memcpy(req->r.migrate_keys.shardGroupId, si->importing_slots_map[slot]->id, RAFT_DBID_LEN);
+        /* currently, we just return, but eventually this is where should kick off migration of the keys */
+        RedisModule_ReplyWithSimpleString(req->ctx, "OK");
+        RaftReqFree(req);
+//        MigrateKeys(rr, req);
```
sjpotter, line 249: this depends on the raft.import pr #337

src/migrate.c
```diff
@@ -0,0 +1,184 @@
...
+    RedisModule_Free(argv_len);
+}
+
+
+void MigrateKeys(RedisRaftCtx *rr, RaftReq *req)
+{
+    JoinLinkState *state = RedisModule_Calloc(1, sizeof(*state));
+    state->type = "migrate";
+    state->connect_callback = transferKeys;
+    time(&(state->start));
```
sjpotter, line 134: `start->start = time(NULL)`

src/migrate.c
```diff
@@ -0,0 +1,184 @@
...
+        RedisModule_ReplyWithCString(req->ctx, "OK");
+        goto exit;
+    }
+
+     for (unsigned int i = 0; i < sg->nodes_num; i++) {
+        LOG_WARNING("MigrateKeys: adding %s:%d", sg->nodes[i].addr.host, sg->nodes[i].addr.port);
+        NodeAddrListAddElement(&state->addr, &sg->nodes[i].addr);
+    }
+    state->req = req;
+
```
sjpotter, line 178: really should depend on #348 here to enable proper username/password being passed to this connection

src/migrate.c
```diff
@@ -0,0 +1,184 @@
...
+        ConnAsyncTerminate(conn);
+        raftAppendRaftDeleteEntry(rr, req);
+    }
+
+    redisAsyncDisconnect(c);
+}
+
+static void transferKeys(Connection *conn)
+{
+    LOG_WARNING("calling transferKeys");
```
sjpotter, line 69: this was for debugging, can remove

src/migrate.c
```diff
@@ -0,0 +1,184 @@
...
+        size_t str_len;
+        const char * str = RedisModule_StringPtrLen(req->r.migrate_keys.keys_serialized[i], &str_len);
+        argv[3 + (i*2) + 1] = RedisModule_Alloc(str_len);
+        memcpy(argv[3 + (i*2) + 1], str, str_len);
+        argv_len[3 + (i*2) + 1] = str_len;
+    }
+
+    if (redisAsyncCommandArgv(ConnGetRedisCtx(conn), transferKeysResponse, conn, argc, (const char **) argv, argv_len) != REDIS_OK) {
+        redisAsyncDisconnect(ConnGetRedisCtx(conn));
+        ConnMarkDisconnected(conn);
```
sjpotter, line 118: need to respond to user with error here if error?

src/migrate.c
```diff
@@ -0,0 +1,184 @@
...
+     */
+    return;
+
+    exit:
+    RaftReqFree(req);
+}
+
+static void transferKeysResponse(redisAsyncContext *c, void *r, void *privdata)
+{
+    LOG_WARNING("calling transferKeysResponse");
```
sjpotter, line 35: remove this debug info

src/migrate.c
```diff
@@ -0,0 +1,184 @@
...
+    Connection *conn = privdata;
+    JoinLinkState *state = ConnGetPrivateData(conn);
+    RedisRaftCtx *rr = ConnGetRedisRaftCtx(conn);
+    RaftReq *req = state->req;
+
+    redisReply *reply = r;
+
+    if (!reply) {
+        LOG_WARNING("RAFT.IMPORT failed: connection dropped.");
+        ConnMarkDisconnected(conn);
```
sjpotter, line 45: return error to user

src/migrate.c
```diff
@@ -0,0 +1,184 @@
...
+    } else if (reply->type != REDIS_REPLY_STATUS || reply->len != 2 || strncmp(reply->str, "OK", 2)) {
+        ConnAsyncTerminate(conn);
+
+        /* FIXME: above should be changed to string eventually? */
+        LOG_WARNING("RAFT.IMPORT unexpected response: type = %d (wanted %d), len = %ld, response = %.*s", reply->type, REDIS_REPLY_STATUS, reply->len, (int) reply->len, reply->str);
+        RedisModule_ReplyWithError(req->ctx, "ERR: received unexpected response from remote cluster, see logs");
+        RaftReqFree(req);
+    } else {
+        ConnAsyncTerminate(conn);
+        raftAppendRaftDeleteEntry(rr, req);
```
sjpotter, line 61: rename to UnlockDeleteEntry

src/migrate.c
```diff
@@ -0,0 +1,184 @@
...
+    entry->id = rand();
+    entry->type = RAFT_LOGTYPE_DELETE_UNLOCK_KEYS;
+    entryAttachRaftReq(rr, entry, req);
+
+    int e = raft_recv_entry(rr->raft, entry, &response);
+    if (e != 0) {
+        replyRaftError(req->ctx, e);
+        entryDetachRaftReq(rr, entry);
+        raft_entry_release(entry);
+        goto exit;
```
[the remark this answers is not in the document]
sjpotter: we have an error, but perhaps not a standard replyRaftError as we did part of migration, this error has to be to reissue command on new leader.

src/migrate.c
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    raft_entry_release(entry);
+
+    /* Unless applied by raft_apply_all() (and freed by it), the request
+     * is pending so we don't free it or unblock the client.
+     */
+    return;
+
+    exit:
```
sjpotter, line 29: whitespace

src/redisraft.c
sjpotter: removed auth

src/migrate.c
```diff
@@ -0,0 +1,184 @@
...
+            if (reply && RedisModule_CallReplyType(reply) == REDISMODULE_REPLY_STRING) {
+                req->r.migrate_keys.keys_serialized[i] = RedisModule_CreateStringFromCallReply(reply);
+            } else {
+                if (reply) {
+                    LOG_WARNING("unexpected response type = %d", RedisModule_CallReplyType(reply));
+                    RedisModule_FreeCallReply(reply);
+                } else {
+                    LOG_WARNING("didn't get a reply!");
+                }
+                RedisModule_ReplyWithError(req->ctx, "ERR see logs");
```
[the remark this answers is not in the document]
sjpotter: added a void 
`replyWithFormatErrorString(RedisModuleCtx *ctx, const char * fmt, ...)` to make it easier to reply with formatted string content.

PR #340, opened 2022-05-04
refactor intercept cluster command handling for getkeysinslot

sjpotter:
we now treat cluster getkeysinslot as a readonly command so it can be used by clients issues migrate commands


sjpotter:
this approach seems to have failed, due to issues with redis cluster, going to try another approach

PR #343, opened 2022-05-10
implement raft.scan to provide a scan that can filter by slots

sjpotter:
mimics redis's scan i.e. either

`raft.scan slots <slot definition>`

or when called again

`raft.scan <cursor> slots <slot definition>`

slot definition is a comma delimited list of slot ranges.  a slot range is either a single slot (100) or a dash delimited slot range (500-520).  So one can do something like "5,15-80,100,0-2"

it returns a 2 element array.

1st element is the cursor (like scan) for next call
2nd element is an array of 2 element arrays [key,slot] that conain the key and slot it belongs to

ex of return values

```
127.0.0.1:5001> raft.scan 28903 slots 1
1) "22991"
2) (empty array)
```
no keys belonging to slot 1, but when called again
```
127.0.0.1:5001> raft.scan 16731 slots 1
1) "28903"
2) 1) 1) "key11850"
      2) (integer) 1
```
we get one key that belonged

and finally
```
127.0.0.1:5001> raft.scan 22991 slots 1
1) "0"
2) (empty array)
```
cursor 0 means we scanned the entire keyspace.

### DIFF
--- a/src/redisraft.c
+++ b/src/redisraft.c
@@ -1387,6 +1387,92 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
     return REDISMODULE_OK;
 }
 
+static int cmdRaftScan(RedisModuleCtx *ctx,
+                         RedisModuleString **argv, int argc)
+{
+    RedisRaftCtx *rr = &redis_raft;
+
+    if (argc < 3) {
+        RedisModule_WrongArity(ctx);
+        return REDISMODULE_OK;
+    }
+
+    char cursor[15] = {0};
+    char slots[REDIS_RAFT_HASH_SLOTS] = {0};
+    char * slot_str;
+
+    size_t str_len;
+    const char * str = RedisModule_StringPtrLen(argv[1], &str_len);
+    strncpy(cursor, str, str_len);
+
+    str = RedisModule_StringPtrLen(argv[2], &str_len);
+    slot_str = RedisModule_Alloc(str_len+1);
+    strncpy(slot_str, str, str_len);
+
+    int ret = parseHashSlots(slots, slot_str);
+    RedisModule_Free(slot_str);
+    if (ret != RR_OK) {
+        RedisModule_ReplyWithError(ctx, "ERR couldn't parse slots");
+        return REDISMODULE_OK;
+    }
+
+    RedisModuleCallReply *reply;
+    if (!(reply = RedisModule_Call(ctx, "scan", "ccc", cursor, "count", rr->config->scan_size))) {
+        RedisModule_ReplyWithError(ctx, "ERR scan failed");
+        return REDISMODULE_OK;
+    }
+
+    if (RedisModule_CallReplyType(reply) != REDISMODULE_REPLY_ARRAY) {
+        RedisModule_ReplyWithError(ctx, "ERR scan returned unexpected reply type");
+        goto exit;
+    }
+
+    if (RedisModule_CallReplyLength(reply) != 2) {
+        RedisModule_ReplyWithError(ctx, "ERR scan returned unexpected array length");
+        goto exit;
+    }
+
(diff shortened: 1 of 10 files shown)


src/redisraft.c
```diff
@@ -1296,6 +1296,108 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
...
+        if (argc != 3) {
+            RedisModule_WrongArity(ctx);
+            return REDISMODULE_OK;
+        }
+        str = RedisModule_StringPtrLen(argv[2], &str_len);
+        slot_str = strndup(str, str_len);
+    }
+
+    int ret = parseHashSlots(slots, slot_str);
+    free(slot_str);
```
tezc, line 1339: For memory allocation, it's better to use `RM_..` family functions just to track memory usage and get OOM error if we are lucky. 

So, I'd use `RM_Alloc()`, `RM_Free()` and `RM_Strdup()`.

src/util.c
```diff
@@ -398,4 +399,76 @@ void HandleAsking(RaftRedisCommandArray *cmds)
...
+        if (dash == NULL) {
+            char *endptr;
+            unsigned int slot = strtoul(tok, &endptr, 10);
+            if (*endptr != 0 || slot > REDIS_RAFT_HASH_MAX_SLOT) {
+                ret = RR_ERROR;
+                goto exit;
+            }
+            slots[slot] = 1;
+        } else {
+            *dash = 0;
```
tezc, line 451: Minor one, just to make reading easier, it's nice to use same literal type. As this is a string, we can use `'\0'` instead of `0`. Same in other usages, compare with `'\0'` rather than `0` etc.

src/redisraft.c
```diff
@@ -1297,6 +1297,111 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
     return REDISMODULE_OK;
 }
 
+static int cmdRaftScan(RedisModuleCtx *ctx,
```
sjpotter, line 1300: currently the api is

raft.scan <cursor> slots [slot-range]

perhaps it can be

raft.scan [slot-range] <cursor>

src/redisraft.c
```diff
@@ -1297,6 +1297,111 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
...
+                         RedisModuleString **argv, int argc)
+{
+    if (argc < 3) {
+        RedisModule_WrongArity(ctx);
+        return REDISMODULE_OK;
+    }
+
+    char cursor[15] = {'0', 0};
```
tezc, line 1308:
```suggestion
    char cursor[15] = {0};
```

sjpotter: use the '\0' instead of 0 for the string

src/redisraft.c
```diff
@@ -1297,6 +1297,111 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
...
+    }
+
+    int ret = parseHashSlots(slots, slot_str);
+    RedisModule_Free(slot_str);
+    if (ret != RR_OK) {
+        RedisModule_ReplyWithError(ctx, "ERR couldn't parse slots");
+        return REDISMODULE_OK;
+    }
+
+    RedisModuleCallReply *reply;
```
sjpotter, line 1349: whitespace

tests/test_util.c
```diff
@@ -80,8 +80,53 @@ static void test_memory_conversion(void **state)
...
+    }
+}
+
 const struct CMUnitTest util_tests[] = {
     cmocka_unit_test(test_raftreq_str),
     cmocka_unit_test(test_memory_conversion),
+    cmocka_unit_test(test_parse_slots),
     { .test_func = NULL }
-};
+};
```
sjpotter, line 132: newline

src/redisraft.c
sjpotter: not needed anymore as no more default cursor, have to specify 0, so can just zero whole thing out.

PR #345, opened 2022-05-14
Test cluster should wait until configuration entries are applied

tezc:
If configuration entries are not applied yet, the node won't have a connection to others. So, if the node starts an election, the election will not be successful. This issue causes test failures on the daily build. The solution is to wait until log entries are applied.

### DIFF
--- a/tests/integration/sandbox.py
+++ b/tests/integration/sandbox.py
@@ -500,9 +500,10 @@ class Cluster(object):
             else:
                 logging.info("{} joining".format(_id))
                 node.join(['localhost:{}'.format(self.base_port + 1)])
+
         self.leader = 1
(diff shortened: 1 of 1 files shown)


yossigo (approved):

sjpotter (commented):
looks fine, though I wonder if the commit message is accurate re the code itself.  i.e. wait_for_unanimity is about applied across the whole cluster, not just on a single node as wait_for_log_applied() would indicate.

tezc:
Need to wait for all nodes. If we force a follower to start an election before it applies the configuration entry, that node will not know the addresses of the other nodes. In a corner case, this is causing test failures, mostly on slow macos build.

PR #346, opened 2022-05-16
enforce maxmemory limits when we accept the command

sjpotter:
This probably need a test. I'm thinking 

- start a cluster with maxmemory limit set relatively low
- run a loop that adds entries until we get the OOM error (within X iterations we should OOM, if not error).
- verify via debug interface that variable that OOMd is not on any of the nodes.

### DIFF
--- a/deps/raft/tests/requirements.txt
+++ b/deps/raft/tests/requirements.txt
@@ -1,4 +1,4 @@
-cffi==1.14.0
+cffi==1.15.0
 colorama==0.4.3
 coloredlogs==14.0
 docopt==0.6.2
--- a/deps/raft/tests/requirements.txt
+++ b/deps/raft/tests/requirements.txt
@@ -1,4 +1,4 @@
-cffi==1.15.0
+cffi==1.14.0
 colorama==0.4.3
 coloredlogs==14.0
 docopt==0.6.2
--- a/src/common.c
+++ b/src/common.c
@@ -10,6 +10,7 @@
 
 #include <string.h>
 #include <strings.h>
+#include <ctype.h>
 
 int redis_raft_in_rm_call = 0;
 
@@ -304,4 +305,108 @@ void raftNodeToString(char *output, const char *dbid, raft_node_t *raft_node)
 void raftNodeIdToString(char *output, const char *dbid, raft_node_id_t raft_id)
 {
     snprintf(output, RAFT_SHARDGROUP_NODEID_LEN + 1, "%.32s%08x", dbid, raft_id);
+}
+
+/*
+ * Dump and Parse the command array for deny-oom status
+ */
+void updateAllDenyOomStatus(RedisRaftCtx *rr, RedisModuleCtx *ctx)
+{
+    enterRedisModuleCall();
+    RedisModuleCallReply * reply = RedisModule_Call(ctx, "command", "");
+    exitRedisModuleCall();
+
+    RedisModule_Assert(reply);
+    RedisModule_Assert(RedisModule_CallReplyType(reply) == REDISMODULE_REPLY_ARRAY);
+
+    if (rr->command_deny_oom_dict) {
+        RedisModule_FreeDict(ctx, rr->command_deny_oom_dict);
+    }
+
+    rr->command_deny_oom_dict = RedisModule_CreateDict(ctx);
+
+    size_t num_commands = RedisModule_CallReplyLength(reply);
+    for (size_t i = 0; i < num_commands; i++) {
+        RedisModuleCallReply *command = RedisModule_CallReplyArrayElement(reply, i);
+
+        if (RedisModule_CallReplyType(command) != REDISMODULE_REPLY_ARRAY) {
+            LOG_WARNING("updateAllDenyOomStatus: command info returned unexpected element type at %d", (int) i);
+            continue;
+        }
+
+        size_t command_array_len = RedisModule_CallReplyLength(command);
+        if ( command_array_len < 3) {
+            LOG_WARNING("updateAllDenyOomStatus: command returned unexpected element length (%d) at %d",
+                        (int) command_array_len , (int) i);
+            continue;
+        }
+
+        RedisModuleCallReply *cmd_name = RedisModule_CallReplyArrayElement(command, 0);
+        size_t cmd_len;
+        const char *cmd_str = RedisModule_CallReplyStringPtr(cmd_name, &cmd_len);
+
+        RedisModuleCallReply *attributes = RedisModule_CallReplyArrayElement(command, 2);
+        bool set = false;
+        for (size_t i = 0; i < RedisModule_CallReplyLength(attributes); i++) {
+            RedisModuleCallReply *option = RedisModule_CallReplyArrayElement(attributes, i);
+            size_t option_len;
+            const char *option_str = RedisModule_CallReplyStringPtr(option, &option_len);
+            if (strncmp("denyoom", option_str, option_len) == 0) {
+                RedisModule_DictReplaceC(rr->command_deny_oom_dict, (char *) cmd_str, cmd_len, (void *) 1);
+                set = true;
+            }
+        }
+        if (!set) {
+            RedisModule_DictSetC(rr->command_deny_oom_dict, (char *) cmd_str, cmd_len, (void *) 0);
+        }
+    }
+
+    RedisModule_FreeCallReply(reply);
+}
+
+/*
+ * returns the deny_oom status for a particular set of commands
+ */
+bool isDenyOomStatus(RedisRaftCtx *rr, RaftRedisCommandArray *cmds)
+{
+    for (int i = 0; i < cmds->len; i++) {
+        RaftRedisCommand * cmd = cmds->commands[i];
+
+        if (cmd->argc < 1) {
+            continue;
+        }
+
+        size_t cmd_len;
+        const char *cmd_str = RedisModule_StringPtrLen(cmd->argv[0], &cmd_len);
+        char buf[64];
+        char *lcmd = toLowerString(cmd_str, cmd_len, buf);
+
+        int nokey;
+        void *ret = RedisModule_DictGetC(rr->command_deny_oom_dict, lcmd, cmd_len, &nokey);
+        if (lcmd != buf) {
+            RedisModule_Free(lcmd);
+        }
+        if (ret) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+/* need to be able to normalize redis strings for dict lookups */
+char *toLowerString(const char *cmd_str, size_t cmd_len, char *buf) {
+    char *lcmd = buf;
+
+    if (cmd_len >= sizeof(buf)) {
+        lcmd = RedisModule_Alloc(cmd_len + 1);
+    }
+
+    for (size_t i = 0; i < cmd_len; i++) {
+        lcmd[i] = (char) tolower(cmd_str[i]);
+    }
+
+    lcmd[cmd_len] = '\0';
+
+    return lcmd;
 }
--- a/src/common.c
+++ b/src/common.c
@@ -8,9 +8,9 @@
 
 #include "redisraft.h"
 
+#include <ctype.h>
 #include <string.h>
 #include <strings.h>
-#include <ctype.h>
 
 int redis_raft_in_rm_call = 0;
 
@@ -313,7 +313,7 @@ void raftNodeIdToString(char *output, const char *dbid, raft_node_id_t raft_id)
 void updateAllDenyOomStatus(RedisRaftCtx *rr, RedisModuleCtx *ctx)
 {
     enterRedisModuleCall();
-    RedisModuleCallReply * reply = RedisModule_Call(ctx, "command", "");
+    RedisModuleCallReply *reply = RedisModule_Call(ctx, "command", "");
     exitRedisModuleCall();
 
     RedisModule_Assert(reply);
@@ -335,9 +335,9 @@ void updateAllDenyOomStatus(RedisRaftCtx *rr, RedisModuleCtx *ctx)
         }
 
         size_t command_array_len = RedisModule_CallReplyLength(command);
-        if ( command_array_len < 3) {
+        if (command_array_len < 3) {
             LOG_WARNING("updateAllDenyOomStatus: command returned unexpected element length (%d) at %d",
-                        (int) command_array_len , (int) i);
+                        (int) command_array_len, (int) i);
             continue;
         }
 
@@ -370,7 +370,7 @@ void updateAllDenyOomStatus(RedisRaftCtx *rr, RedisModuleCtx *ctx)
 bool isDenyOomStatus(RedisRaftCtx *rr, RaftRedisCommandArray *cmds)
 {
     for (int i = 0; i < cmds->len; i++) {
-        RaftRedisCommand * cmd = cmds->commands[i];
+        RaftRedisCommand *cmd = cmds->commands[i];
 
         if (cmd->argc < 1) {
             continue;
@@ -395,7 +395,8 @@ bool isDenyOomStatus(RedisRaftCtx *rr, RaftRedisCommandArray *cmds)
 }
 
 /* need to be able to normalize redis strings for dict lookups */
-char *toLowerString(const char *cmd_str, size_t cmd_len, char *buf) {
+char *toLowerString(const char *cmd_str, size_t cmd_len, char *buf)
+{
     char *lcmd = buf;
 
     if (cmd_len >= sizeof(buf)) {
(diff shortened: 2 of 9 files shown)


tezc:
Things that comes to my mind:

1- We shouldn't accept commands before applying an entry from the current term. This way, we will always have logs replayed and we'll have the latest memory usage data (more or less). Otherwise, we might think that state machine is empty and we can accept commands even though we'll be over the configured memory limit after log replay.

We already have this check for non-quorum reads (for a different purpose). So, I guess we can check this condition inside `handleRedisCommand()` for all kind of operations.

2- We should allow commands that will free memory. We can take a look at Redis for this, it's doing the same (commands marked with `CMD_DENYOOM`).

3- Let's enable OOM tests in Redis Test Suite and check if these tests are valid: https://github.com/RedisLabs/redisraft/blob/f54615484a2a6d7a5cc5d8a9762d8508378f8756/tests/redis-suite/skip.txt#L45

4- Consider Multi-Exec case and decide whether we want to match Redis' behavior: https://github.com/redis/redis/blob/f62d52a5ad794513ca21168becfd06c532d2b704/src/server.c#L3731

tezc:
Things that comes to my mind:

1- We shouldn't accept commands before applying an entry from the current term. This way, we will always have logs replayed and we'll have the latest memory usage data (more or less). Otherwise, we might think that state machine is empty and we can accept commands even though we'll be over the configured memory limit after log replay.

We already have this check for non-quorum reads (for a different purpose). So, I guess we can check this condition inside `handleRedisCommand()` for all kind of operations.

2- We should allow commands that will free memory. We can take a look at Redis for this, it's doing the same (commands marked with `CMD_DENYOOM`).

3- Let's enable OOM tests in Redis Test Suite and check if these tests are valid: https://github.com/RedisLabs/redisraft/blob/f54615484a2a6d7a5cc5d8a9762d8508378f8756/tests/redis-suite/skip.txt#L45

4- Consider Multi-Exec case and decide whether we want to match Redis' behavior: https://github.com/redis/redis/blob/f62d52a5ad794513ca21168becfd06c532d2b704/src/server.c#L3731

tezc:
I agree on all points, actually we are talking about the same thing but I wasn't clear. 

> the current check compares snapshot's last_applied_term which I'm not sure is as good?

Actually, it's state machine's last applied term. Name of that variable is a bit confusing :) So, we are actually talking about the same thing. Either we do this check: https://github.com/RedisLabs/redisraft/blob/9e3fbd20fd282b6461228bc0e01f31253743fb3b/src/redisraft.c#L524 or we can add `last_applied_term` to the library as you suggested. Actually, as part of this PR, https://github.com/RedisLabs/raft/pull/105, I needed it for some other reason and added there. Just missing a public getter.

sjpotter:
I'm digging into the oom tests we have disabled, and am unsure we can actually support them.  Basically, they seem to be function/script oriented.  we don't have insight into what they do.  we would have to deny them before log append time (i.e. the eval call in case of a script).  We would behave differently than normal redis in these cases and go above the limit.

sjpotter:
and in regards to that, it seems lua scripts can oom on their own, so we'd have to run them all with allow-oom to prevent them from failing randomly.

src/redisraft.c
```diff
@@ -598,6 +599,70 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
...
+        }
+
+        size_t cmd_len;
+        const char * cmd_str = RedisModule_StringPtrLen(cmd->argv[0], &cmd_len);
+
+        int nokey;
+        if (RedisModule_DictGetC(rr->command_deny_oom_dict, (char *) cmd_str, cmd_len, &nokey)) {
+            return true;
+        } else if (nokey) {
+            enterRedisModuleCall();
```
fadidahanna, line 618: I suggest moving the code under `nokey` condition that call to `INFO` and analyse the result into a separate method - `getDenyOomStatus()`, update the dict accordingly in a separate method too, and rename this method to 'isDenyOomCommands()`

src/redisraft.c
```diff
@@ -659,6 +724,15 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
         return;
     }
 
+    bool deny_oom = getDenyOomStatus(rr, ctx, cmds);
+    raft_term_t term = raft_get_current_term(rr->raft);
+
+    /* write command, check maxmemory */
+    if (deny_oom && (RedisModule_GetUsedMemoryRatio() > 1 || term != rr->snapshot_info.last_applied_term)) {
```
fadidahanna, line 731: did we agree for `term != rr->snapshot_info.last_applied_term` state?

src/redisraft.c
```diff
@@ -659,6 +724,15 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
         return;
     }
 
+    bool deny_oom = getDenyOomStatus(rr, ctx, cmds);
```
tezc, line 727: Please take a look at https://github.com/RedisLabs/redisraft/blob/1ef16de3c847d4ece2dccfdcbc43da8254f726c1/src/commands.c#L196

We can define `CMD_SPEC_DENY_OOM` and add it to command flags. Otherwise, we cannot afford fetching "deny-oom" on append time.

fadidahanna: If we are doing it anyway, so yes - it will be easy to maintain such a dictionary, but I wonder what will be the result for module commands - it probably depends on the order we load them, did we have a solution for that? (it's relevant also for `populateCommandSpecFromRedis`) @tezc

fadidahanna:
> 4- Consider Multi-Exec case and decide whether we want to match Redis' behavior: https://github.com/redis/redis/blob/f62d52a5ad794513ca21168becfd06c532d2b704/src/server.c#L3731

As was discussed in our daily today -we do want to have the same behaviour as Redis for the Multi-Exec case

src/redisraft.c
fadidahanna: k, anyway I suggest taking out the `populateCommandSpecFromRedis` call from the config path.

src/redisraft.c
sjpotter: ok, restructured

src/redisraft.c
sjpotter: restuctured to run just "commands" and cache everything at startup and at module load time

deps/raft/tests/requirements.txt
```diff
@@ -1,4 +1,4 @@
-cffi==1.14.0
```
fadidahanna, line 1: This should be done in libraft, please remove

src/common.c
```diff
@@ -305,3 +306,43 @@ void raftNodeIdToString(char *output, const char *dbid, raft_node_id_t raft_id)
...
+            return true;
+        }
+
+    }
+
+    return false;
+}
+
+/* need to be able to normalize redis strings for dict lookups */
+char *toLowerString(const char *cmd_str, size_t cmd_len, char *buf)
```
fadidahanna, line 333: is this method a generic one? if yes, please rename `cmd_str` and `cmd_len`, remove the above comment, and add a comment about the `buf` param.

src/common.c
```diff
@@ -305,3 +306,43 @@ void raftNodeIdToString(char *output, const char *dbid, raft_node_id_t raft_id)
...
+{
+    char *lcmd = buf;
+
+    if (cmd_len >= sizeof(buf)) {
+        lcmd = RedisModule_Alloc(cmd_len + 1);
```
tezc, line 338: There are a few problems here. If `cmd_len` is larger than `sizeof(buf)`, it will allocate memory and we need to free later. Also, `sizeof(buf)` is just `8` as it is `char*`. Previously, it was a local buffer.

Probably, we don't need to make this a function as we don't call it anywhere else.  I suggest reverting it.

src/redisraft.h
```diff
@@ -411,6 +411,8 @@ typedef struct RedisRaftCtx {
     int entered_eval;             /* handling a lua script */
     RedisModuleDict *locked_keys; /* keys thar have been locked for migration */
 
+    RedisModuleDict *command_deny_oom_dict; /* track commands if they are oom or not */
```
tezc, line 414: Is this leftover?

src/common.c
```diff
@@ -305,3 +306,43 @@ void raftNodeIdToString(char *output, const char *dbid, raft_node_id_t raft_id)
 {
     snprintf(output, RAFT_SHARDGROUP_NODEID_LEN + 1, "%.32s%08x", dbid, raft_id);
 }
+
+/*
+ * returns the deny_oom status for a particular set of commands
+ */
+bool isDenyOomStatus(RedisRaftCtx *rr, RaftRedisCommandArray *cmds)
```
tezc, line 313: I don't think we need this function. We already fetch aggregated flags just before we call this function. So, we can just do "if check" there. Also, this way we won't do dictionary lookup twice. (Triggered by `CommandSpecGet()`. )

src/redisraft.c
```diff
@@ -1874,6 +1883,18 @@ static int registerRaftCommands(RedisModuleCtx *ctx)
     return REDISMODULE_OK;
 }
 
+void moduleChangeCallback(RedisModuleCtx *ctx, RedisModuleEvent e, uint64_t sub, void *data)
+{
+    REDISMODULE_NOT_USED(e);
+
+    if (sub != REDISMODULE_SUBEVENT_MODULE_LOADED) {
```
fadidahanna, line 1890: We probably don't need to remove cmds from our dict, but we do want to update them in case a new version of a module was loaded - in the current implementation the dict will not be updated with the new flags.

tezc, line 1890: If we are doing that, I guess we should also do it for `REDISMODULE_SUBEVENT_MODULE_UNLOADED` as well to clear table.

I didn't try these btw, hope it works :)

tezc: We should call this also on `REDISMODULE_SUBEVENT_MODULE_UNLOADED`. 
So, it will refresh the command table. On load, it will add new specs. On unload, it will remove stale specs. Wdyt?

src/common.c
fadidahanna: Agree with reverting

src/redisraft.c
sjpotter: yea, the condition is wrong, unsure what I was thinking.  need to fix.

src/common.c
sjpotter: its not needed now, but its needed for eval (or makes the code cleaner with eval).  We could pass the cs in into it.

src/redisraft.c
fadidahanna: Agree, I guess there is no really need to do any optimization here :)

src/common.c
sjpotter: will try to figure out how to put the isDenyOom into CommandSpecGetAggregateFlags which should solve all issues

tezc:
@sjpotter I've pushed a commit rather than adding many small comments. Hope it's okay. 
Mainly, I added two changes:
- In Multi, I see Redis does not accumulate commands in OOM state. It doesn't check whether the command is `DENY_OOM` or not. It just rejects any command. I've added a change to match that behavior. 
- Moved `term != rr->snapshot_info.last_applied_term` check to the main flow and we should return`-CLUSTERDOWN` in this case. We'll not accept commands before applying the first entry. This is not necessary for regular command handling but no harm comes from this requirement anyway. It will help with non-quorum reads and maxmemory handling for now.

src/multi.c
```diff
@@ -97,15 +97,14 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
...
-        raft_term_t term = raft_get_current_term(rr->raft);
 
         if (cmd_flags & CMD_SPEC_UNSUPPORTED) {
             RedisModule_ReplyWithError(ctx, "ERR not supported by RedisRaft");
             multiState->error = true;
         } else if (cmd_flags & CMD_SPEC_DONT_INTERCEPT) {
             RedisModule_ReplyWithError(ctx, "ERR not supported by RedisRaft inside MULTI/EXEC");
             multiState->error = true;
-        } else if (cmd_flags & CMD_SPEC_DENYOOM && (RedisModule_GetUsedMemoryRatio() > 1.0 || term != rr->snapshot_info.last_applied_term)) {
+        } else if (RedisModule_GetUsedMemoryRatio() > 1.0) {
```
sjpotter, line 107: I don't think this change is correct.  If one has a multi of all GETs, it should work just fime.  i.e. the cmd-flags & CMD_SPEC_DENYOOM is needed.

tests/integration/test_multi.py
```diff
@@ -159,22 +159,33 @@ def test_multi_with_unsupported_commands(cluster_factory):
...
+    node = cluster.leader_node()
+    node.execute('set', 'key1', val)
+    node.execute('config', 'set', 'maxmemory', 100)
+
+    # Test MULTI - COMMAND - EXEC
+    assert node.execute('MULTI') == b'OK'
+    with raises(ResponseError, match='OOM command not allowed'):
+        node.execute('GET', 'key1')
+    with raises(ResponseError, match='Transaction discarded'):
+        node.execute('EXEC')
```
sjpotter, line 179: with my current comment, I don't think should fail, this should be a set transaction.

tests/integration/test_multi.py
```diff
@@ -159,22 +159,33 @@ def test_multi_with_unsupported_commands(cluster_factory):
...
+
+    # Test MULTI - COMMAND - DISCARD
+    assert node.execute('MULTI') == b'OK'
+    with raises(ResponseError, match='OOM command not allowed'):
+        node.execute('GET', 'key1')
+    assert node.execute('DISCARD') == b'OK'
```
sjpotter, line 185: this should be a set

tests/integration/test_multi.py
```diff
@@ -159,22 +159,33 @@ def test_multi_with_unsupported_commands(cluster_factory):
...
+
+    # Clear OOM
+    node.execute('config', 'set', 'maxmemory', 100000000)
+    assert node.execute('MULTI') == b'OK'
+    assert node.execute('GET', 'key1') == b'QUEUED'
+    assert node.execute('EXEC') == [val.encode()]
```
sjpotter, line 191: should also be a set

src/multi.c
tezc: I mentioned in the comment above, this is how Redis works right now. I don't agree much with this behavior but I think we should try to follow Redis. This is why I've made this change. 

Please see: https://github.com/redis/redis/issues/9926. Someone complains about it (rightfully) and it is still unresolved. If Redis solves it, we can follow it. Wdyt?

sjpotter: ok, I see.  I was confused by the special work done for scripts.  as noted in the thread there seems inconsistent to treat eval and multi differently, but <shrug>

sjpotter:
with @tezc changes I think this is ok for a final review before merging

tezc (dismissed):
LGTM. 

One note, I threw the idea of catching MODULE_LOAD event to rebuild cmdspec table but I didn't try it. It'd probably be better to implement it in another PR with a test. (Similar to Redis, we can have simple modules and load them at the same time) . I guess we can do it in a follow-up PR.

fadidahanna (dismissed):

fadidahanna:
+ Agree with @tezc comment regarding MODULE_LOAD/UNLOAD events

sjpotter:
I thought we already did that.  I added a module change callback which recreates the table

```
void moduleChangeCallback(RedisModuleCtx *ctx, RedisModuleEvent e, uint64_t sub, void *data)
{
    REDISMODULE_NOT_USED(e);

    /* recreate the CommandSpec dict on any module change */
    RRStatus ret = CommandSpecSet(ctx, redis_raft.config.ignored_commands);
    RedisModule_Assert(ret == RR_OK);
}
```

tezc:
@sjpotter What I meant, we don't have a test for that. So, currently, we don't know if it works or not if nobody has tried. I've just mentioned it as an idea that might work. 

So, I think either we should add a test in this PR or in a follow-up PR.

sjpotter:
hmm, trying to think how to test it.  easy enough to build a module (ala redis tests) that adds a command, but unsure how to test if parsed correctly without a "debug" mode that basically returns true/false if the command is in the table or not

tezc:
I think you can load that new the module and add a command `raft.debug commandspec` which returns the flags?

sjpotter:
ok, essentially what I proposed.  will add that.

sjpotter:
but will need some help on the cmake side to get this right.

tezc:
Ok, I'll take a look at Cmake part.

It'd nice if you can add this as something generic in the test suite. 

```python
n1 = cluster.add_node();
n1.load_module("test_commandspec.so"); 
n1.unload_module("test_commandspec");
```

Something like that.

tezc (approved):

PR #347, opened 2022-05-23
asking as an independent command

sjpotter:
Based on conversation the understanding of asking was incorrect, its an independent command that sets a flag on the client, this is to make it the same.

- saves client state when go into asking mode
- set RedisRaftCommandArray as asking if client put itself into asking state
- no longer have to parse multi specially
- serialize/deserialize the asking bool the log entry for RaftRedisCommandArray
- no longer parse asking in the apply path (as not appended to log as a command)

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -242,19 +242,3 @@ unsigned int CommandSpecGetAggregateFlags(RaftRedisCommandArray *array, unsigned
 
     return flags;
 }
-
-bool IsKeyCommands(RedisRaftCtx *rr,
-                   RedisModuleCtx *ctx,
-                   RaftRedisCommandArray *cmds) {
-    for (int i = 0; i < cmds->len; i++) {
-        RaftRedisCommand *cmd = cmds->commands[i];
-
-        /* Iterate command keys */
-        int num_keys = 0;
-        RedisModule_GetCommandKeys(rr->ctx, cmd->argv, cmd->argc, &num_keys);
-        if (num_keys) {
-            return true;
-        }
-    }
-    return false;
-}
--- a/src/commands.c
+++ b/src/commands.c
@@ -8,6 +8,7 @@
 /* ------------------------------------ Command Classification ------------------------------------ */
 
 #include <string.h>
+#include <strings.h>
 #include <ctype.h>
 #include "redisraft.h"
 
@@ -240,4 +241,20 @@ unsigned int CommandSpecGetAggregateFlags(RaftRedisCommandArray *array, unsigned
     }
 
     return flags;
+}
+
+bool IsKeyCommands(RedisRaftCtx *rr,
+                   RedisModuleCtx *ctx,
+                   RaftRedisCommandArray *cmds) {
+    for (int i = 0; i < cmds->len; i++) {
+        RaftRedisCommand *cmd = cmds->commands[i];
+
+        /* Iterate command keys */
+        int num_keys = 0;
+        RedisModule_GetCommandKeys(rr->ctx, cmd->argv, cmd->argc, &num_keys);
+        if (num_keys) {
+            return true;
+        }
+    }
+    return false;
 }
--- a/src/commands.c
+++ b/src/commands.c
@@ -219,19 +219,6 @@ unsigned int CommandSpecGetAggregateFlags(RaftRedisCommandArray *array, unsigned
     for (int i = 0; i < array->len; i++) {
         /* we get the flag for the command that is asking, not the asking */
         RedisModuleString * cmd = array->commands[i]->argv[0];
-        size_t cmd_len;
-        const char * cmd_str = RedisModule_StringPtrLen(cmd, &cmd_len);
-
-        if (!strncasecmp("asking", cmd_str, cmd_len)) {
-            /* TODO: when asking <magic> support is added, need to adjust to argv[2] */
-            if (array->commands[i]->argc > 1) {
-                cmd = array->commands[i]->argv[1];
-            } else {
-                /* necessary to protect/ignore a bare "asking" command */
-                continue;
-            }
-        }
-
         const CommandSpec *cs = CommandSpecGet(cmd);
         if (cs) {
             flags |= cs->flags;
--- a/src/common.c
+++ b/src/common.c
@@ -66,7 +66,7 @@ void replyRedirect(RedisModuleCtx *ctx, unsigned int slot, NodeAddr *addr)
 {
     char buf[sizeof(addr->host) + 256];
 
-    snprintf(buf, sizeof(buf), "MOVED %d %s:%u", slot, addr->host, addr->port);
+    snprintf(buf, sizeof(buf), "MOVED %u %s:%u", slot, addr->host, addr->port);
     RedisModule_ReplyWithError(ctx, buf);
 }
 
@@ -117,7 +117,7 @@ static const char *err_clusterdown = "CLUSTERDOWN No raft leader";
  * Note: it's possible to have an elected but unknown leader node, in which
  * case NULL will also be returned.
  */
-raft_node_t *getLeaderNodeOrReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
+raft_node_t *getLeaderRaftNodeOrReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
 {
     raft_node_t *node = raft_get_leader_node(rr->raft);
     if (!node) {
@@ -127,47 +127,32 @@ raft_node_t *getLeaderNodeOrReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
     return node;
 }
 
-/* Checks that we *have* a Raft leader and its identity.
+/* Returns the leader node (Node), or reply a -CLUSTERDOWN error
+ * and return NULL.
  *
- * If no leader is currently elected, replies a -CLUSTERDOWN error and
- * return RR_ERROR.
- *
- * If ret_leader is provided, it is used to return the current leader
- * by reference along with an RR_OK return value.
- *
- * If ret_leader is not provided and the current node is NOT the leader,
- * a -MOVED reply is generated with the current leader and RR_ERROR is
- * returned.
+ * Note: it's possible to have an elected but unknown leader node, in which
+ * case NULL will also be returned.
  */
-RRStatus checkLeader(RedisRaftCtx *rr, RedisModuleCtx *ctx, Node **ret_leader)
+Node *getLeaderNodeOrReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
 {
-    Node *node = NULL;
-
-    if (raft_is_leader(rr->raft)) {
-        return RR_OK;
+    raft_node_t *raft_node = getLeaderRaftNodeOrReply(rr, ctx);
+    if (!raft_node) {
+        return NULL;
     }
 
-    raft_node_t *leader = raft_get_leader_node(rr->raft);
-    if (leader) {
-        node = raft_node_get_udata(leader);
-    }
-
+    Node *node = raft_node_get_udata(raft_node);
     if (!node) {
         RedisModule_ReplyWithError(ctx, err_clusterdown);
-        return RR_ERROR;
     }
 
-    if (ret_leader) {
-        *ret_leader = node;
-        return RR_OK;
-    }
+    return node;
+}
 
-    /* One anomaly here is that may redirect a client to the leader even for
-     * commands have no keys, which is something Redis Cluster never does. We
-     * still need to consider how this impacts clients which may not expect it.
-     */
-    replyRedirect(ctx, 0, &node->addr);
-    return RR_ERROR;
+
+/* Checks if the current cluster is the leader */
+bool isLeader(RedisRaftCtx *rr) {
+
+    return raft_is_leader(rr->raft);
 }
 
 /* Check that we're not in REDIS_RAFT_LOADING state.  If not, reply with -LOADING
--- a/src/common.c
+++ b/src/common.c
@@ -89,6 +89,11 @@ void replyCrossSlot(RedisModuleCtx *ctx)
     RedisModule_ReplyWithError(ctx, "CROSSSLOT Keys in request don't hash to the same slot");
 }
 
+void replyClusterDownWithNoRaftLeader(RedisModuleCtx *ctx)
+{
+    RedisModule_ReplyWithError(ctx, "CLUSTERDOWN No raft leader");
+}
+
 void replyWithFormatErrorString(RedisModuleCtx *ctx, const char * fmt, ...)
 {
     va_list ap;
@@ -109,8 +114,6 @@ void replyWithFormatErrorString(RedisModuleCtx *ctx, const char * fmt, ...)
 
 }
 
-static const char *err_clusterdown = "CLUSTERDOWN No raft leader";
-
 /* Returns the leader node (raft_node_t), or reply a -CLUSTERDOWN error
  * and return NULL.
  *
@@ -121,7 +124,7 @@ raft_node_t *getLeaderRaftNodeOrReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
 {
     raft_node_t *node = raft_get_leader_node(rr->raft);
     if (!node) {
-        RedisModule_ReplyWithError(ctx, err_clusterdown);
+        replyClusterDownWithNoRaftLeader(ctx);
     }
 
     return node;
@@ -142,7 +145,7 @@ Node *getLeaderNodeOrReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
 
     Node *node = raft_node_get_udata(raft_node);
     if (!node) {
-        RedisModule_ReplyWithError(ctx, err_clusterdown);
+        replyClusterDownWithNoRaftLeader(ctx);
     }
 
     return node;
--- a/src/common.c
+++ b/src/common.c
@@ -62,7 +62,7 @@ void replyRaftError(RedisModuleCtx *ctx, int error)
 }
 
 /* Create a -MOVED reply. */
-void replyRedirect(RedisModuleCtx *ctx, int slot, NodeAddr *addr)
+void replyRedirect(RedisModuleCtx *ctx, unsigned int slot, NodeAddr *addr)
 {
     char buf[sizeof(addr->host) + 256];
 
@@ -71,7 +71,7 @@ void replyRedirect(RedisModuleCtx *ctx, int slot, NodeAddr *addr)
 }
 
 /* Create a -ASK reply. */
-void replyAsk(RedisRaftCtx *rr, RedisModuleCtx *ctx, int slot) {
+void replyAsk(RedisRaftCtx *rr, RedisModuleCtx *ctx, unsigned int slot) {
     ShardGroup *sg = rr->sharding_info->importing_slots_map[slot];
     if (!sg) {
         RedisModule_ReplyWithError(ctx, "ERR no importing shard group to ask");
@@ -79,7 +79,7 @@ void replyAsk(RedisRaftCtx *rr, RedisModuleCtx *ctx, int slot) {
     }
 
     char buf[sizeof(sg->nodes[0].addr.host) + 256];
-    snprintf(buf, sizeof(buf), "ASK %d %s:%u", slot, sg->nodes[0].addr.host, sg->nodes[0].addr.port);
+    snprintf(buf, sizeof(buf), "ASK %u %s:%u", slot, sg->nodes[0].addr.host, sg->nodes[0].addr.port);
     RedisModule_ReplyWithError(ctx, buf);
 }
 
--- a/src/common.c
+++ b/src/common.c
@@ -148,13 +148,6 @@ Node *getLeaderNodeOrReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
     return node;
 }
 
-
-/* Checks if the current cluster is the leader */
-bool isLeader(RedisRaftCtx *rr) {
-
-    return raft_is_leader(rr->raft);
-}
-
 /* Check that we're not in REDIS_RAFT_LOADING state.  If not, reply with -LOADING
  * and return an error.
  */
--- a/src/multi.c
+++ b/src/multi.c
@@ -86,13 +86,6 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
     size_t cmd_len;
     const char *cmd_str = RedisModule_StringPtrLen(cmd->argv[0], &cmd_len);
 
-    if (cmd_len == 6 && !strncasecmp(cmd_str, "ASKING", 6)) {
-        /* TODO: will have to be > 2 in future with magic value in asking for argv[2] */
-        if (cmd->argc > 1) {
-            cmd_str = RedisModule_StringPtrLen(cmd->argv[1], &cmd_len);
-        }
-    }
-
     if (cmd_len == 5 && !strncasecmp(cmd_str, "MULTI", 5)) {
         if (multiState) {
             RedisModule_ReplyWithError(ctx, "ERR MULTI calls can not be nested");
--- a/src/multi.c
+++ b/src/multi.c
@@ -44,17 +44,7 @@ void MultiClientStateReset(ClientState * clientState)
 bool MultiHandleCommand(RedisRaftCtx *rr,
                         RedisModuleCtx *ctx, RaftRedisCommandArray *cmds)
 {
-    ClientState *clientState;
-
-    /* MULTI/EXEC bundling takes place only if we have a single command. If we
-     * have multiple commands we've received this as a RAFT.ENTRY input and
-     * bundling, probably through a proxy, and bundling was done before.
-     */
-    if (cmds->len != 1) {
-        return false;
-    }
-
-    clientState = ClientStateGet(rr, ctx);
+    ClientState *clientState = ClientStateGet(rr, ctx);
 
     /* Is this a MULTI command? */
     RaftRedisCommand *cmd = cmds->commands[0];
--- a/src/multi.c
+++ b/src/multi.c
@@ -34,7 +34,7 @@
 
 #include "redisraft.h"
 
-void ResetMultiClientState(ClientState * clientState)
+void MultiClientStateReset(ClientState * clientState)
 {
     RaftRedisCommandArrayFree(&clientState->multi_state.cmds);
     clientState->multi_state.active = false;
@@ -82,7 +82,7 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
         }
 
         if (clientState->multi_state.error) {
-            ResetMultiClientState(clientState);
+            MultiClientStateReset(clientState);
             RedisModule_ReplyWithError(ctx, "EXECABORT Transaction discarded because of previous errors.");
             return true;
         }
@@ -90,16 +90,16 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
         /* Just swap our commands with the EXEC command and proceed. */
         RaftRedisCommandArrayFree(cmds);
         RaftRedisCommandArrayMove(cmds, &clientState->multi_state.cmds);
-        ResetMultiClientState(clientState);
+        MultiClientStateReset(clientState);
 
         return false;
     } else if (cmd_len == 7 && !strncasecmp(cmd_str, "DISCARD", 7)) {
-        if (!clientState || !clientState->multi_state.active) {
+        if (!clientState->multi_state.active) {
             RedisModule_ReplyWithError(ctx, "ERR DISCARD without MULTI");
             return true;
         }
 
-        ResetMultiClientState(clientState);
+        MultiClientStateReset(clientState);
         RedisModule_ReplyWithSimpleString(ctx, "OK");
 
         return true;
--- a/src/multi.c
+++ b/src/multi.c
@@ -35,39 +35,17 @@
 #include "redisraft.h"
 
 
-typedef struct MultiState {
-    RaftRedisCommandArray cmds;
-    bool error;
-} MultiState;
-
-void MultiInitClientState(RedisRaftCtx *rr)
+void ResetMultiClientState(ClientState * clientState)
 {
-    rr->multi_client_state = RedisModule_CreateDict(NULL);
-}
-
-uint64_t MultiClientStateCount(RedisRaftCtx *rr)
-{
-    return RedisModule_DictSize(rr->multi_client_state);
-}
-
-void MultiFreeClientState(RedisRaftCtx *rr, unsigned long long client_id)
-{
-    MultiState *state = NULL;
-
-    if (RedisModule_DictDelC(rr->multi_client_state, &client_id,
-                             sizeof(client_id), &state) == REDISMODULE_OK) {
-        if (state) {
-            RaftRedisCommandArrayFree(&state->cmds);
-            RedisModule_Free(state);
-        }
-    }
+    RaftRedisCommandArrayFree(&clientState->multiState.cmds);
+    clientState->multiState.active = false;
+    clientState->multiState.error = false;
 }
 
 bool MultiHandleCommand(RedisRaftCtx *rr,
                         RedisModuleCtx *ctx, RaftRedisCommandArray *cmds)
 {
-    unsigned long long client_id;
-    MultiState *multiState;
+    ClientState *clientState;
 
     /* MULTI/EXEC bundling takes place only if we have a single command. If we
      * have multiple commands we've received this as a RAFT.ENTRY input and
@@ -77,9 +55,7 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
         return false;
     }
 
-    client_id = RedisModule_GetClientId(ctx);
-    multiState = RedisModule_DictGetC(rr->multi_client_state,
-                                      &client_id, sizeof(client_id), NULL);
+    clientState = GetClientState(rr, ctx);
 
     /* Is this a MULTI command? */
     RaftRedisCommand *cmd = cmds->commands[0];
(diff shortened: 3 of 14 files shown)


src/raft.c
```diff
@@ -1138,8 +1136,9 @@ RRStatus RedisRaftInit(RedisModuleCtx *ctx, RedisRaftCtx *rr, RedisRaftConfig *c
         rr->resp_call_fmt = "v";
     }
 
-    /* Client state for MULTI support */
+    /* Client state for MULTI/ASKING support */
     rr->multi_client_state = RedisModule_CreateDict(ctx);
```
sjpotter, line 1140: combine into a single client_state dict that is a struct that is expandable.

sjpotter:
Q. re should asking be serialized as a bit/flag or should it be appended to the array and then parsed out on apply time.

src/redisraft.h
```diff
@@ -700,6 +701,17 @@ typedef struct JoinLinkState {
...
+typedef struct MultiState {
+    RaftRedisCommandArray cmds;
+    bool active;
+    bool error;
+} MultiState;
+
+typedef struct ClientState {
+    MultiState multiState;
+    bool asking;
+} ClientState;
```
tezc, line 713: I feel like we'll use client struct a lot in future, I'd use a shorter name, just "struct Client" probably.

src/redisraft.h
```diff
@@ -700,6 +701,17 @@ typedef struct JoinLinkState {
...
 } JoinLinkState;
 
+typedef struct MultiState {
+    RaftRedisCommandArray cmds;
+    bool active;
+    bool error;
+} MultiState;
+
+typedef struct ClientState {
+    MultiState multiState;
```
tezc, line 711: Just for style, can you use `multi_state` as variable name.

src/redisraft.h
```diff
@@ -778,6 +790,10 @@ void FreeImportKeys(ImportKeys *target);
 unsigned int keyHashSlot(const char *key, size_t keylen);
 unsigned int keyHashSlotRedisString(RedisModuleString *s);
 RRStatus parseHashSlots(char * slots, char * string);
+ClientState * GetClientState(RedisRaftCtx *rr, RedisModuleCtx * ctx);
```
tezc, line 793: Regarding naming functions, I did many mistakes and merged bunch of things previously but trying to follow a style recently. So, for exported functions, function names start with uppercase letter and with a common prefix. So, maybe these functions should start with `Client..()`

src/util.c
```diff
@@ -500,4 +500,34 @@ RRStatus parseHashSlots(char * slots, char * string)
...
+{
+    ClientState *clientState = RedisModule_Calloc(sizeof(ClientState), 1);
+    RedisModule_DictSetC(rr->client_state, &client_id, sizeof(client_id), clientState);
+}
+
+void FreeClientState(RedisRaftCtx *rr, unsigned long long client_id)
+{
+    ClientState *state = NULL;
+
+    if (RedisModule_DictDelC(rr->client_state, &client_id,
```
tezc, line 526: I think if we don't have a client struct here, we are in trouble (it shouldn't be possible?). Maybe, we should just assert here. We can also use assert when we add a client to dict, just to verify there is no existing client.

src/redisraft.c
```diff
@@ -1716,7 +1774,7 @@ static void handleInfo(RedisModuleInfoCtx *ctx, int for_crash_report)
     RedisModule_InfoAddFieldULongLong(ctx, "snapshots_created", rr->snapshots_created);
 
     RedisModule_InfoAddSection(ctx, "clients");
-    RedisModule_InfoAddFieldULongLong(ctx, "clients_in_multi_state", MultiClientStateCount(rr));
+    RedisModule_InfoAddFieldULongLong(ctx, "clients_state_tracking", ClientStateCount(rr));
```
tezc, line 1777: Right now, this field becomes "connected client count". So, if I'm not mistaken, this will be same with `connected_clients` field in INFO. 

Previously, it was "clients in multi state" and a test uses it. Maybe, we loop over clients and count clients in multi state ?

If we can run/verify test in some other way, it might also mean we don't need this field. @yossigo Do you think this is useful? Redis doesn't report clients in multi state if I'm not mistaken.

src/multi.c
```diff
@@ -35,39 +35,17 @@
...
-    bool error;
-} MultiState;
-
-void MultiInitClientState(RedisRaftCtx *rr)
-{
-    rr->multi_client_state = RedisModule_CreateDict(NULL);
-}
-
-uint64_t MultiClientStateCount(RedisRaftCtx *rr)
+void ResetMultiClientState(ClientState * clientState)
```
tezc, line 38: Looks like multi.c is not necessary anymore (it's getting shorter). If you want, we can delete the file and move the function here back into the redisraft.c.

src/redisraft.c
```diff
@@ -549,16 +549,65 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
     return false;
 }
 
+static bool AskingHandleCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRedisCommandArray *cmds)
```
tezc, line 552: I added a comment about style and I know it's not consistent in the codebase (I did bunch of mistakes) but let's try to follow the common style:
- Exported functions start with uppercase letter and a common prefix: 
   - RaftLogCreate(), RaftLogOpen() etc.
   - NodeAddrParse(), NodeAddrEqual() etc.
- Static functions start with lowercase letter. 
- Exception to this: common.c functions start with lowercase. (That's just the reality). So, these are just short helper functions . Probably, it's okay we keep it that way, start with lowercase letter without a common prefix as they are unrelated with each other anyway.

src/redisraft.c
```diff
@@ -604,6 +653,8 @@ static void handleRedisCommand(RedisRaftCtx *rr,
         if (rr->config->quorum_reads) {
             RaftReq *req = RaftReqInit(ctx, RR_REDISCOMMAND);
             RaftRedisCommandArrayMove(&req->r.redis.cmds, cmds);
+            /* need to save asking flag when we move the cmds */
+            req->r.redis.cmds.asking = cmds->asking;
```
tezc, line 657: Maybe this should be part of RaftRedisCommandArrayMove(). Is there an exception to this? Whenever we move a command into the array, we also adopt asking state?

src/redisraft.h
sjpotter: ok

src/redisraft.h
sjpotter: I personally like ClientState, I think the little bit more descriptive is fine.

src/redisraft.h
sjpotter: hmm.  So we want SubjectVerb not VerbSubject?

src/redisraft.h
tezc: ok

src/util.c
sjpotter: sure, its validating redis, which is fine.

src/redisraft.h
tezc: Normally, in C world, people use a common prefix in exported function names (C version of namespaces). So, I guess it naturally becomes SubjectVerb because of the prefix. Good thing is that when you see that function name, you know where to find it. If a function starts with Shard... I know it's in cluster.c or with Conn.. it's in connection.c etc.

src/multi.c
sjpotter: So I was thinking sort of the opposite, it be nicer if different commands were handled in different files.

src/redisraft.c
sjpotter: in this case, isn't your comment consistent with this?

src/redisraft.c
sjpotter: possibly.  we dont want to overwrite state each time from true to false (I think) thought that could be solved via a cmds.asking |= cmds->asking within it.

src/redisraft.c
tezc: Sorry I wasn't clear.  Yeah both are same.  The other comment(https://github.com/RedisLabs/redisraft/pull/347#discussion_r889877447) is consistent with this.

src/redisraft.c
tezc: > we dont want to overwrite state each time from true to false

Not sure I understand. What is the problem here? Also, we are going to set asking true once because if client is in asking mode, it will stay in asking mode forever, right?  So, once we are in asking mode, we'll never get "asking=false", no? (We may need to validate this somewhere though).

src/multi.c
tezc: In general, I find harder to read/navigate multiple files, especially on github.  (Also, it reminds me my Java days that you have billion of files even in the simplest project, [link](https://github.com/EnterpriseQualityCoding/FizzBuzzEnterpriseEdition)). 

If you have bunch of commands that you handle in 10 lines of code, creating a file for each seems overkill to me but anyway, I don't have a strong opinion.

src/redisraft.c
```diff
@@ -549,11 +549,60 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
...
+    ClientState *clientState = ClientStateGet(rr, ctx);
+    return clientState->asking;
+}
+
 static void handleRedisCommand(RedisRaftCtx *rr,
                                RedisModuleCtx *ctx, RaftRedisCommandArray *cmds)
 {
     Node *leader_proxy = NULL;
 
+    if (AskingHandleCommand(rr, ctx, cmds)) {
```
tezc, line 598:
```suggestion
    if (handleAsking(rr, ctx, cmds)) {
```

Style. Local functions start with lowercase letter.

src/redisraft.c
```diff
@@ -549,11 +549,60 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
...
+        return;
+    }
+
+    if (GetAskingState(rr, ctx)) {
```
tezc, line 602: Probably no need this function anymore and it can be part of handleAsking() above.

src/redisraft.c
tezc: This comment is still valid.

src/multi.c
fadidahanna: Agree with @sjpotter, I think we should separate logic into different files, as long as we can.

fadidahanna (commented):
Still in the middle of reviewing - will continue later on today

src/multi.c
```diff
@@ -77,81 +54,72 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
...
         RaftRedisCommandArrayFree(cmds);
-        RaftRedisCommandArrayMove(cmds, &multiState->cmds);
-        MultiFreeClientState(rr, client_id);
+        RaftRedisCommandArrayMove(cmds, &clientState->multi_state.cmds);
+        ResetMultiClientState(clientState);
 
         return false;
     } else if (cmd_len == 7 && !strncasecmp(cmd_str, "DISCARD", 7)) {
-        if (!multiState) {
+        if (!clientState || !clientState->multi_state.active) {
```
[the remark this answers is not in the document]
sjpotter: old code, before we created it on each connection.  to remove.

src/redisraft.c
sjpotter: right, I didn't remove it, because felt it needed discussion.  outdated is because I had to rebase.

src/redisraft.c
sjpotter: so, I'm not 100% sure about this, while they can be combined, they really are doing 2 different things.  1 is intercepting the asking command (and now that I think of it, should be part of the intercept function), the other is changing the current state of this command due to client state.

sjpotter: I actually took my idea right here and re factored this code a bit.

src/redisraft.c
```diff
@@ -549,11 +549,60 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
     return false;
 }
 
+static bool AskingHandleCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRedisCommandArray *cmds)
+{
+    if (cmds->len != 1) {
+        return false;
+    }
+
+    /* TODO: will have to be == 2 in future with magic value in asking for argv[2] */
```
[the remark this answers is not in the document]
sjpotter: https://github.com/RedisLabs/redisraft/issues/371

src/redisraft.c
```diff
@@ -549,11 +549,60 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
...
+    size_t cmd_len;
+    const char *cmd_str = RedisModule_StringPtrLen(cmd->argv[0], &cmd_len);
+
+    if (cmd_len != 6 || strncasecmp(cmd_str, "ASKING", 6) != 0) {
+        return false;
+    }
+
+    Node *ret;
+    if (checkLeader(rr, ctx, &ret) == RR_ERROR) {
+        /* change to replyAsk once that's committed */
```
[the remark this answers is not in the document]
sjpotter: so, I tried to change it, can't seem to change it "correctly" so just replying moved to it for now.

src/redisraft.c
```diff
@@ -549,11 +549,60 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
...
+        replyRedirect(ctx, -1, &ret->addr);
+        return true;
+    }
+
+    ClientState *clientState = ClientStateGet(rr, ctx);
+    clientState->asking = true;
```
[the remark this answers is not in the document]
sjpotter: unsure if there's much value to that?

src/redisraft.c
sjpotter: done

src/redisraft.c
```diff
@@ -574,6 +594,14 @@ static void handleRedisCommand(RedisRaftCtx *rr,
         return;
     }
 
+    /* update cmds array with the client's saved "asking" state */
+    cmds->asking = getAskingState(rr, ctx);
```
tezc, line 598: I realized one thing, what is our "asking" policy about the proxy? If these commands come via proxy, "asking" might be true here already and we'll overwrite it back to "false". This might be okay I guess, in the worst case is the request will be rejected. So, either we need a comment about proxy or we should not overwrite this value if is already set? Something like:

```c
/* Comment describing proxy case */
if (!cmds->asking) {
     cmds->asking = getAskingState(rr, ctx);
}
```

src/redisraft.c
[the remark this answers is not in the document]
fadidahanna: Well, if we have a getter `getAskingState` it makes sense to also have a setter, and keep `asking` field as a private

src/redisraft.c
```diff
@@ -640,6 +652,29 @@ static void handleRedisCommand(RedisRaftCtx *rr,
...
+     * if they have to.
+     */
+    if (handleInterceptedCommands(rr, ctx, cmds)) {
+        return;
+    }
+
+    /* update cmds array with the client's saved "asking" state */
+    if (!cmds->asking) {
+        cmds->asking = getAskingState(rr, ctx);
+    }
```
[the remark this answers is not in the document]
tezc: No, I mean `cmds->asking = getAskingState(rr,ctx);` line should be enough without the if check.

src/multi.c
```diff
@@ -34,124 +34,82 @@
...
-    bool error;
-} MultiState;
-
-void MultiInitClientState(RedisRaftCtx *rr)
-{
-    rr->multi_client_state = RedisModule_CreateDict(NULL);
-}
-
-uint64_t MultiClientStateCount(RedisRaftCtx *rr)
+void MultiClientStateReset(ClientState * clientState)
```
tezc, line 37:
```suggestion
void MultiClientStateReset(ClientState *clientState)
```

src/redisraft.c
```diff
@@ -509,6 +509,19 @@ static void handleMigrateCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRedi
     raft_entry_release(entry);
 }
 
+static void  handleAsking(RedisRaftCtx *rr, RedisModuleCtx *ctx)
```
tezc, line 512:
```suggestion
static void handleAsking(RedisRaftCtx *rr, RedisModuleCtx *ctx)
```

src/redisraft.c
sjpotter: removed as discussed in zoom meeting

src/redisraft.c
```diff
@@ -509,6 +509,19 @@ static void handleMigrateCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRedi
...
+{
+    Node *ret;
+    if (checkLeader(rr, ctx, &ret) == RR_ERROR) {
+        replyRedirect(ctx, -1, &ret->addr);
+    }
```
tezc, line 517:
```suggestion
    if (checkLeader(rr, ctx, NULL) == RR_ERROR) {
        return;
    }
```

This should be this way, right? `checkLeader()` already replies with "redirect".

src/redisraft.h
```diff
@@ -797,8 +812,8 @@ long long int RaftLogRewrite(RedisRaftCtx *rr, const char *filename, raft_index_
 void RaftLogRemoveFiles(const char *filename);
 void RaftLogArchiveFiles(RedisRaftCtx *rr);
 RRStatus RaftLogRewriteSwitch(RedisRaftCtx *rr, RaftLog *new_log, unsigned long new_log_entries);
-int RaftMetaRead(RaftMeta *meta, const char* filename);
-int RaftMetaWrite(RaftMeta *meta, const char* filename, raft_term_t term, raft_node_id_t vote);
+int RaftMetaRead(RaftMeta *meta, const char *filename);
+int RaftMetaWrite(RaftMeta *meta, const char*filename, raft_term_t term, raft_node_id_t vote);
```
tezc, line 815: These lines are unrelated but I see you fixed '*' alignment in the first line but you couldn't fix it on the second line :D

src/redisraft.c
sjpotter: actually, not, when passed with &ret, it doesn't redirect (i.e. for proxy mode), but need to remember what I was doing here, as you're right, if redirecting, no need for that.  I think think the point was that this in an ASKING handling, and hence we might want to reply with an ask response and hence would need the leader.  In our case, we don't right now (ask doesn't work for us right now, as the ASK doesn't include the slot we are asking about and hence can't return a well formed ask response, as an aside, part of me thinks that's also something missing from the protocol, an "ASK" response includes the slot its about, an ASKING should include it (and only work on said slot)

src/redisraft.c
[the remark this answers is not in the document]
sjpotter: ok, understood now, updated.

src/redisraft.h
sjpotter: I think I just noticed the first one for some reason, so got to it as a nit.

src/redisraft.c
```diff
@@ -509,6 +509,34 @@ static void handleMigrateCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRedi
...
+static void handleAsking(RedisRaftCtx *rr, RedisModuleCtx *ctx)
+{
+    Node *leader_proxy = NULL;
+
+    Node **ret = rr->config->follower_proxy ? &leader_proxy : NULL;
+    if (checkLeader(rr, ctx, ret) == RR_ERROR) {
+        if (!rr->config->follower_proxy) {
+            return;
+        }
+    }
```
tezc, line 533: This check won't work, please see the usage in `handleRedisCommand()`.

Other than that, probably we don't even need to handle this. Is there a reason we are handling this? We can just set the client "asking" whether we are leader or not.

src/util.c
```diff
@@ -500,4 +500,29 @@ RRStatus parseHashSlots(char * slots, char * string)
 exit:
     RedisModule_Free(string);
     return ret;
-}
\ No newline at end of file
+}
+
+ClientState *ClientStateGet(RedisRaftCtx *rr, RedisModuleCtx * ctx)
```
tezc, line 505:
```suggestion
ClientState *ClientStateGet(RedisRaftCtx *rr, RedisModuleCtx *ctx)
```
There were two misaligned pointer in this line, the previous suggestion had both. Consider accepting changes via github, you don't miss the changes and it's just one click. I find it easier than manual commit.

src/redisraft.c
sjpotter: so, the Q is, if we are not leader, how should we handle an ASKING.

do we 

a) accept it and handle leader later on a "real" command
b) reject it and reply moved back to migrating shardgroup
c) reject it and reply moved to our leader
d) reject it and reply ask to our leader

you're arguing for a. 

c. I'm not sure makes sense because the asking won't transfer with it.  (clients expect that a moved response means they are being redirected to primary owner, i.e. the migrating shard) 
b. is impossible because we dont know who it is (i.e. asking doesn't give us a slot)
d. would be the "best" (i.e. you're asking us, but you want our leader).  But unsure we can do it right (as dont know slot).

now if we do it later, we have 2 options

1) replyMoved back to migrating shard
2) since we know that we are an asking, replyAsk as appropriate (with correct slot info) to our leader.  I'm a bit skeptical on this.

but I'm ok with your suggestion, just wanted to enumerate everything.

sjpotter: another approach would be to try treat an ASK to a follower as an inherent follower proxy mode?  i.e. we will proxy it to right node, as we can't "nicely" redirect you to it.

tezc: As I said, if this scenario is super rare, we shouldn't optimize for it. What happens if we don't do this? If this doesn't cause any visible improvement, I'd say we shouldn't write code for it.

sjpotter: I'm not sure if its super rare, it be behavior I'd expect to see in real life usage.  I actually think as long as we have a proxy mode, it makes sense to leverage it for asking.

sjpotter: i.e. with the recent understanding that asking resets the state after every execution, every asking will be an independent command returned from an ask response.

with proxy we avoid a lot of ping ponging and just handle it naturally.  the reason to not support proxy normally is to teach the client where to go so it we can avoid the man in the middle.  there wont be ping poinging.  However, for ask/asking thats irrelvant.

With that said, can discuss it tomorrow

src/redisraft.c
```diff
@@ -679,6 +679,11 @@ static void handleRedisCommand(RedisRaftCtx *rr,
         return;
     }
 
+    /* reset asking state when we go to execute the command */
```
fadidahanna, line 676: You should reset the asking state *after* handling the current command.

src/redisraft.c
```diff
@@ -596,14 +586,18 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
     }
 
     /* Confirm that we're the leader and handle redirect or proxying if not. */
-    Node **ret = rr->config->follower_proxy ? &leader_proxy : NULL;
+    Node **ret = rr->config->follower_proxy || cmds->asking ? &leader : NULL;
```
fadidahanna, line 589: I suggest to change the `checkLeader` behaviour to check if we are the leader, or if ret exists fill it with value - without reply anything + having a wrapper method that based on the asking status or follower_proxy value reply :
!leader && asking -> reply ASK, 
!leader && !follower_proxy-> reply redirect)
I feel it's hard to follow the logic in the current implementation.

src/redisraft.c
tezc: Also, line 669 and line 677 contradict each other?

src/raft.c
```diff
@@ -187,9 +187,11 @@ static ShardGroup *GetSlotShardGroup(RedisRaftCtx *rr, int slot, bool asking)
 }
 
 static RRStatus validateRaftRedisCommandArray(RedisRaftCtx *rr, RedisModuleCtx *reply_ctx,
-                                              RaftRedisCommandArray *cmds, unsigned int slot)
+                                              RaftRedisCommandArray *cmds)
 {
-    RedisModule_Assert(slot <= REDIS_RAFT_HASH_MAX_SLOT);
+    RedisModule_Assert(0 <= cmds->slot && cmds->slot <= REDIS_RAFT_HASH_MAX_SLOT);
+
+    unsigned int slot = (unsigned int) cmds->slot;
```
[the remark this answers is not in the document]
tezc: `-1` is used as logical "no keys".  Maybe, we can define `NO_KEYS = -1` and use the variable unsigned. Then, check like `if (slot == NO_KEYS)`;

src/redisraft.c
```diff
@@ -575,14 +586,18 @@ static void handleRedisCommand(RedisRaftCtx *rr,
...
     if (checkLeader(rr, ctx, ret) == RR_ERROR) {
         return;
     }
 
     /* Proxy */
-    if (leader_proxy) {
-        if (ProxyCommand(rr, ctx, cmds, leader_proxy) != RR_OK) {
+    if (leader) {
+        if (cmds->asking) {
+            if (computeHashSlot(rr, ctx, cmds) == RR_OK){
```
tezc, line 597: What do we do if `computeHashSlot()` returns an error? We have to reply.

fadidahanna: `computeHashSlot` was changed to support replying in error, but I think we should avoid that.

tezc: Ok, I see this branch lags behind the master. @sjpotter has already changed this behavior: https://github.com/RedisLabs/redisraft/pull/331

src/raft.c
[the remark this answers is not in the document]
fadidahanna: actually, we don't use slot value for validation at all, since we return an error from `computeHashSlot` - `-1` is used only to mark that we are handling the first key in `computeHashSlot` - so I guess we can use a local flag for that...

src/redisraft.c
fadidahanna: Well, I'm not sure, because it's part of the changes in the PR... anyway, we all agree that `computeHashSlot` should not reply in case of error... 🙏

src/redisraft.c
```diff
@@ -673,12 +675,54 @@ static void handleRedisCommand(RedisRaftCtx *rr,
...
+     * commands have no keys, which is something Redis Cluster never does. We
+     * still need to consider how this impacts clients which may not expect it.
+     */
+    replyRedirect(ctx, 0, &leader->addr);
+}
+
+static RRStatus handleNonLeaderCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, Node *leader, RaftRedisCommandArray *cmds) {
+    RedisModule_Assert(!raft_is_leader(rr->raft));
+
+    // reply ASK for ASKING state commands
```
tezc, line 700: I guess this is one of our rare "consistent" coding convention, we use `/* */` for comments.

src/redisraft.c
```diff
@@ -673,12 +675,54 @@ static void handleRedisCommand(RedisRaftCtx *rr,
...
+    if (cmds->asking) {
+        RedisModule_Assert(cmds);
```
tezc, line 702: Assert won't be triggered ever as we deference it on line 701.

src/redisraft.c
```diff
@@ -673,12 +675,54 @@ static void handleRedisCommand(RedisRaftCtx *rr,
...
+            return RR_ERROR;
+        }
+
+        replyAsk(rr, ctx, (unsigned int) slot);
+        return RR_OK;
+    }
+
+    // forward commands to the leader in follower_proxy state
+    if (rr->config->follower_proxy) {
+        RedisModule_Assert(cmds);
```
tezc, line 715: Same here, line 701 deference it already.

src/commands.c
```diff
@@ -241,3 +242,19 @@ unsigned int CommandSpecGetAggregateFlags(RaftRedisCommandArray *array, unsigned
...
+
+bool IsKeyCommands(RedisRaftCtx *rr,
+                   RedisModuleCtx *ctx,
+                   RaftRedisCommandArray *cmds) {
+    for (int i = 0; i < cmds->len; i++) {
+        RaftRedisCommand *cmd = cmds->commands[i];
+
+        /* Iterate command keys */
+        int num_keys = 0;
+        RedisModule_GetCommandKeys(rr->ctx, cmd->argv, cmd->argc, &num_keys);
```
tezc, line 254: This functions returns something and you need to free it if I'm not mistaken.

src/raft.c
```diff
@@ -272,20 +270,23 @@ static RRStatus validateRaftRedisCommandArray(RedisRaftCtx *rr, RedisModuleCtx *
...
+    unsigned int slot;
+
     if (!isSharding(rr)) {
         return RR_OK;
     }
 
-    if (computeHashSlot(rr, ctx, cmds) != RR_OK) {
-        return RR_ERROR;
+    /* If commands have no keys, continue */
+    if (!IsKeyCommands(rr, ctx, cmds)) {
```
tezc, line 280: I think we duplicate the work here. We loop over the command array and pull the keys. We are going to do the same thing in `computeHashSlot()`. I know we did this to avoid cast for `slot` but I'm not sure it is worth the effort. Maybe, we should just go with cast?

src/common.c
```diff
@@ -127,47 +127,32 @@ raft_node_t *getLeaderNodeOrReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
...
 
-    /* One anomaly here is that may redirect a client to the leader even for
-     * commands have no keys, which is something Redis Cluster never does. We
-     * still need to consider how this impacts clients which may not expect it.
-     */
-    replyRedirect(ctx, 0, &node->addr);
-    return RR_ERROR;
+
+/* Checks if the current cluster is the leader */
+bool isLeader(RedisRaftCtx *rr) {
```
tezc, line 153: Sorry for nitpick, as this function doesn't do anything on top of `raft_is_leader()`, probably we don't need it.

src/redisraft.c
fadidahanna: yes, this is a leftover from previous impl. fixed.

src/redisraft.c
```diff
@@ -645,6 +661,72 @@ static void handleRedisCommand(RedisRaftCtx *rr,
     raft_entry_release(entry);
 }
 
+static void handleRedisCommand(RedisRaftCtx *rr,
+                               RedisModuleCtx *ctx, RaftRedisCommandArray *cmds) {
```
tezc, line 665:
```suggestion
static void handleRedisCommand(RedisRaftCtx *rr,
                               RedisModuleCtx *ctx, RaftRedisCommandArray *cmds) 
{
```

src/redisraft.c
```diff
@@ -645,6 +661,72 @@ static void handleRedisCommand(RedisRaftCtx *rr,
...
+    RedisModule_Assert(!raft_is_leader(rr->raft));
+    RedisModule_Assert(cmds);
+
+    /* reply ASK for ASKING state commands */
+    if (cmds->asking) {
+        int slot;
+        if (computeHashSlot(rr, ctx, cmds, &slot) != RR_OK) {
+            return RR_ERROR;
+        }
+        RedisModule_Assert(slot != -1); /* ASKING mode requests always have keys */
```
tezc, line 711: I think we should not crash even if a client sends an "asking" command without a key. Client might be buggy or it might be in the development phase etc. We should reply an error in this case.

src/redisraft.c
```diff
@@ -645,6 +661,72 @@ static void handleRedisCommand(RedisRaftCtx *rr,
...
+
+static RRStatus handleNonLeaderCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, Node *leader, RaftRedisCommandArray *cmds) {
+    RedisModule_Assert(!raft_is_leader(rr->raft));
+    RedisModule_Assert(cmds);
+
+    /* reply ASK for ASKING state commands */
+    if (cmds->asking) {
+        int slot;
+        if (computeHashSlot(rr, ctx, cmds, &slot) != RR_OK) {
+            return RR_ERROR;
```
tezc, line 709: We should also reply an error here.

tezc: I guess we can do something like: 
```
if slot = -1: "ERR asking without keys"
if not RR_OK: replyCrossSlot();
```

src/redisraft.c
```diff
@@ -645,6 +661,72 @@ static void handleRedisCommand(RedisRaftCtx *rr,
...
+    if (MultiHandleCommand(rr, ctx, cmds)) {
+        return;
+    }
+
+    handleRedisCommandAppend(rr, ctx, cmds);
+
+    /* reset asking state when after executing the command */
+    if (cmds->asking) {
+        setAskingState(rr, ctx, false);
+    }
```
tezc, line 687:
```suggestion
    setAskingState(rr, ctx, false);
```
I think we don't need an if check here.

src/redisraft.c
```diff
@@ -645,6 +661,72 @@ static void handleRedisCommand(RedisRaftCtx *rr,
...
+}
```
tezc, line 688:
```suggestion
    /* update cmds array with the client's saved "asking" state */
    cmds->asking = getAskingState(rr, ctx);
    setAskingState(rr, ctx, false);

    if (handleInterceptedCommands(rr, ctx, cmds)) {
        return;
    }

    /* Check if MULTI/EXEC bundling is required. */
    if (MultiHandleCommand(rr, ctx, cmds)) {
        return;
    }

    handleRedisCommandAppend(rr, ctx, cmds);
}
```

I guess we should clear "asking" state no matter what.

src/redisraft.c
```diff
@@ -639,10 +656,83 @@ static void handleRedisCommand(RedisRaftCtx *rr,
...
+        int slot;
+        RRStatus res = computeHashSlot(rr, ctx, cmds, &slot);
+        if (slot == -1) {
+            /* ASKING mode requests should always have keys */
+            RedisModule_ReplyWithError(ctx, "ERR cmd without keys in asking mode");
+            return RR_ERROR;
+        }
+
+        if (res == RR_ERROR) {
+            replyCrossSlot(ctx);
```
tezc, line 717:
```suggestion
            replyCrossSlot(ctx);
            return RR_ERROR;
```
We need to return here. 
Btw looks like this function does not need a return value.

fadidahanna: I prefer to have it right now, it may be useful for the error handling task.

tezc: sure

tezc (approved):

Issue #302, opened 2022-03-27
test_raft_sort test failure


tezc:
- `test_raft_sort_sets()` failed here: https://github.com/sjpotter/redisraft/runs/5648599700?check_suite_focus=true

- `test_raft_sort_hashes()` failed on my local:

```
____________________________________________________________________________________________________________________ test_raft_sort_hashes ____________________________________________________________________________________________________________________

cluster = <integration.sandbox.Cluster object at 0x7f71e45799d0>

    def test_raft_sort_hashes(cluster):
        cluster.create(3)
        for i in range(1000):
            key = ''.join(random.choices(string.ascii_uppercase, k=6))
            val = ''.join(random.choices(string.ascii_uppercase, k=6))
            cluster.execute("hset", "test", key, val)
            cluster.execute("hset", "test", key + "1", val)
    
        hgetall = cluster.execute("raft._sort_reply", "hgetall", "test")
        hkeys = cluster.execute("raft._sort_reply", "hkeys", "test")
        hvals = cluster.execute("raft._sort_reply", "hvals", "test")
    
>       assert len(hgetall) == 4000
E       assert 3996 == 4000
E         -3996
E         +4000

tests/integration/test_random.py:273: AssertionError
---------------------------------------------------------------------------------------------------------------------- Captured log call ----------------------------------------------------------------------------------------------------------------------
INFO     root:sandbox.py:195 starting node: args = ['/home/ozan/Desktop/redis/src/redis-server', '--port', '5001', '--bind', '0.0.0.0', '--dir', '/home/ozan/Desktop/redisraft/tests/tmp/c81b8eea-2c4c-470f-9030-e56003a1f54a', '--dbfilename', 'redis1.rdb', '--loglevel', 'debug', '--loadmodule', '/home/ozan/Desktop/redisraft/redisraft.so', 'id', '1', 'addr', 'localhost:5001', 'raft-log-filename', 'redis1.db', 'raft-log-fsync', 'no', 'loglevel', 'debug', 'tls-enabled', 'no']
DEBUG    sandbox:sandbox.py:52 c0/n1/stdout: 53211:C 27 Mar 2022 03:12:45.908 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
[...]

PR #351, opened 2022-05-30
Fix randomness issue in sort test

tezc:
`sort` tests use random but not necessarily unique values. We see some failures on CI. I've changed the values to be unique and hoping to fix the test failures. Also, reverted debugging info change from https://github.com/RedisLabs/redisraft/pull/298

Fixes: https://github.com/RedisLabs/redisraft/issues/302


tezc:
@sjpotter I've changed the values to integer. As far as I can see, it doesn't matter if it's string or int, we are actually comparing them as integer anyway. (maybe I'm mistaken, Python does something else).

sjpotter (approved):

PR #363, opened 2022-06-05
Fix linter warnings in util.c

tezc:
Fixed unused import and implicit cast issue in util.c

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -10,12 +10,13 @@
 #include <strings.h>
 #include <stdlib.h>
 #include <ctype.h>
-#include <assert.h>
+
 #include "redisraft.h"
 
 int RedisModuleStringToInt(RedisModuleString *str, int *value)
 {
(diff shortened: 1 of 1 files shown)


src/util.c
```diff
@@ -10,12 +10,13 @@
 #include <strings.h>
 #include <stdlib.h>
 #include <ctype.h>
-#include <assert.h>
+
 #include "redisraft.h"
 
 int RedisModuleStringToInt(RedisModuleString *str, int *value)
```
fadidahanna, line 16: Maybe for another PR - but do we want to keep the `RedisModules` prefix for internal lib functions?

fadidahanna (approved):

src/util.c
tezc: This can be part of another PR. Ideally, I'd like to see namespace usage per file, like util functions should start with `Util...`.

PR #364, opened 2022-06-05
Fix linter warnings in tests

tezc:
Fixed linter warnings in Python tests.

### DIFF
--- a/tests/integration/raftlog.py
+++ b/tests/integration/raftlog.py
@@ -134,12 +134,12 @@ class LogEntry(RawEntry):
     def parse_cmdlist(data):
         cmds = []
         cmd_count = int(data[0][1:])
-        idx = 1
+        i = 1
         for _ in range(cmd_count):
-            args_count = int(data[idx][1:])
-            i_end = idx + 1 + args_count * 2
-            cmds.append(' '.join(data[idx + 2:i_end:2]))
-            idx = i_end
+            args_count = int(data[i][1:])
+            i_end = i + 1 + args_count*2
+            cmds.append(' '.join(data[i+2:i_end:2]))
+            i = i_end
         return '|'.join(cmds)
 
     def data(self, decode=False):
@@ -178,10 +178,10 @@ class RaftLog(object):
     def read(self):
         while True:
             try:
-                ety = RawEntry.from_file(self.logfile)
+                entry = RawEntry.from_file(self.logfile)
             except EOFError:
                 break
-            self.entries.append(ety)
+            self.entries.append(entry)
         self.dump()
 
     def header(self):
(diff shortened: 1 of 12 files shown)


tests/integration/raftlog.py
```diff
@@ -178,10 +178,10 @@ def reset(self):
     def read(self):
         while True:
             try:
-                entry = RawEntry.from_file(self.logfile)
+                ety = RawEntry.from_file(self.logfile)
```
fadidahanna, line 181: what was wrong with `entry` ?

tests/integration/sandbox.py
```diff
@@ -641,7 +650,7 @@ def raft_retry(self, func):
                     if no_leader_first:
                         LOG.info("-CLUSTERDOWN response received, will retry"
                                  " for %s seconds", self.noleader_timeout)
-                        #no_leader_first = False
+                        # no_leader_first = False
```
fadidahanna, line 653: do we still need this comment?

tests/integration/test_sharding.py
```diff
@@ -135,7 +146,7 @@ def test_shard_group_validation(cluster):
             '12345678901234567890123456789012',
             '1', '1',
             '1001', '20000', '1',
-            '1234567890123456789012345678901234567890', '1.1.1.1:1111') == b'OK'
+            '1234567890123456789012345678901234567890', '1.1.1.1:1111')
```
fadidahanna, line 149: do we want to `assert` ?

tests/integration/test_sharding.py
```diff
@@ -144,7 +155,7 @@ def test_shard_group_validation(cluster):
             '12345678901234567890123456789012',
             '1', '1',
             '1000', '1001', '1',
-            '1234567890123456789012345678901234567890', '1.1.1.1:1111') == b'OK'
+            '1234567890123456789012345678901234567890', '1.1.1.1:1111')
```
fadidahanna, line 158: do we want to `assert` ?

tests/integration/raftlog.py
tezc: Looks like main function, `__main__`, does not define variable scope in Python. All variables in this procedure are global variables. So, linter was complaining about shadowing variables from the outer scope. What a horrible language. 

I added another commit to move `__main__` into another function, so, variables will be part of function scope. Probably, this is clear for everybody and less likely to fall in a trap in future.

tests/integration/sandbox.py
tezc: Not sure. If we delete the comment, we can also delete the if check. If we do that, we probably need to fix error message as well, it is not accurate. As this is becoming something unrelated, it can be part of another PR.

tests/integration/test_sharding.py
tezc: This command will throw an exception. It's inside `with raises()` statement which is waiting to catch exception. So, assert() is there, just hidden.

tests/integration/test_sharding.py
tezc: Same as above

fadidahanna (approved):

PR #375, opened 2022-06-08
add session migration key ("magic") to shardgroup replace command

### LINKED ISSUE #371: add migration session key (magic) to shardgroups replace slot definition
needed for consistent migration

sjpotter:
the slot range specification now takes an additional field for the shared session migration key, so its syntax is now

<start slot> <end slot> <slot type> <shared session migration key>

its been added to serialization / desrialization code and the unit tests have been extended to handle it.

closed #371


src/cluster.c
```diff
@@ -44,7 +44,8 @@ char *ShardGroupSerialize(ShardGroup *sg)
     for (unsigned int i = 0; i < sg->slot_ranges_num; i++) {
         /* individual slot ranges */
         ShardGroupSlotRange *sr = &sg->slot_ranges[i];
-        buf = catsnprintf(buf, &buf_size, "%u\n%u\n%u\n", sr->start_slot, sr->end_slot, sr->type);
+        /* storing magic as an unsigned long long as there's a strtoull() function, not a stroll() */
+        buf = catsnprintf(buf, &buf_size, "%u\n%u\n%u\n%llu\n", sr->start_slot, sr->end_slot, sr->type, sr->magic);
```
fadidahanna, line 48: actually there is `strtoll` but, in any case `magic` probably should be an `unsigned`, so let's change it in `ShardGroupSlotRange` definition. Beside that - I think we should change `magic` to some more meaningful name - to indicate its real meaning and values.

src/commands.c
```diff
@@ -222,7 +223,7 @@ unsigned int CommandSpecGetAggregateFlags(RaftRedisCommandArray *array, unsigned
         const char * cmd_str = RedisModule_StringPtrLen(cmd, &cmd_len);
 
         if (!strncasecmp("asking", cmd_str, cmd_len)) {
-            /* TODO: when asking <magic> support is added, need to adjust to argv[2] */
+            /* TODO: when asking <migration_session_key> support is added, need to adjust to argv[2] */
```
tezc, line 226: Why do we do something specific to migration and asking (on line 229) here? Is this leftover?

fadidahanna: Don't think it's leftover - in case of asking cmd, we should take the cmd name from argv[1] (in the current implementation), but agree we can remove this comment, because changing the ASKING spec is related to redis, and won't change in the nearest future.

fadidahanna: Agree, will remove it in the better-asking PR.

src/commands.c
```diff
@@ -222,7 +223,6 @@ unsigned int CommandSpecGetAggregateFlags(RaftRedisCommandArray *array, unsigned
         const char * cmd_str = RedisModule_StringPtrLen(cmd, &cmd_len);
 
         if (!strncasecmp("asking", cmd_str, cmd_len)) {
-            /* TODO: when asking <magic> support is added, need to adjust to argv[2] */
```
sjpotter, line 225: unsure why this comment was removed.

fadidahanna: Actually the whole `asking` section was removed in #347  - ASKING is a single command, with no args.

src/cluster.c
```diff
@@ -930,10 +939,11 @@ ShardGroup *ShardGroupParse(RedisModuleCtx *ctx, RedisModuleString **argv, int a
     sg->slot_ranges_num = num_slots;
     sg->slot_ranges = RedisModule_Calloc(num_slots, sizeof(ShardGroupSlotRange));
     for (int i = 0; i < num_slots; i++) {
-        long long start_slot, end_slot, type;
+        long long start_slot, end_slot, type, key;
         if (RedisModule_StringToLongLong(argv[argidx++], &start_slot) != REDISMODULE_OK ||
             RedisModule_StringToLongLong(argv[argidx++], &end_slot) != REDISMODULE_OK ||
             RedisModule_StringToLongLong(argv[argidx++], &type) != REDISMODULE_OK ||
+            RedisModule_StringToLongLong(argv[argidx++], &key) != REDISMODULE_OK ||
```
sjpotter, line 946: we talked about validating that key is >= 0 so it an be enforced as unsigned cleanly?

tezc (approved):

src/cluster.c
fadidahanna: we can add this validation, sure.

PR #379, opened 2022-06-12
add injected error for testing migration.

sjpotter:
add an emulated

"connection failure" error
"raft.import" failure
"leadership loss" failure

depends on #370

### DIFF
--- a/src/redisraft.h
+++ b/src/redisraft.h
@@ -347,7 +347,7 @@ typedef struct RedisRaftCtx {
     /* Debug - Testing */
     struct RaftReq *debug_req;                   /* Current RAFT.DEBUG request context, if processing one */
     long long debug_delay_apply;                 /* If not zero, sleep microseconds before the execution of a command */
-    migration_debug migration_debug;             /* for debugging migration, places to inject error */
+    MigrationDebug migration_debug;             /* for debugging migration, places to inject error */
 
     /* General stats */
     unsigned long client_attached_entries;       /* Number of log entries attached to user connections */
--- a/src/redisraft.h
+++ b/src/redisraft.h
@@ -308,6 +308,13 @@ void fsyncThreadStart(FsyncThread *th, void (*on_complete)(void *result));
 void fsyncThreadAddTask(FsyncThread *th, int fd, raft_index_t requested_index);
 void fsyncThreadWaitUntilCompleted(FsyncThread *th);
 
+typedef enum {
+    RAFT_DEBUG_MIGRATION_NONE                    = 0,
+    RAFT_DEBUG_MIGRATION_EMULATE_CONNECT_FAILED,
+    RAFT_DEBUG_MIGRATION_EMULATE_IMPORT_FAILED,
+    RAFT_DEBUG_MIGRATION_EMULATE_UNLOCK_FAILED,
+    RAFT_DEBUG_MIGRATION_MAX,
+} migration_debug;
 
 /* Global Raft context */
 typedef struct RedisRaftCtx {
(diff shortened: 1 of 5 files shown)


src/redisraft.h
```diff
@@ -308,6 +308,13 @@ void fsyncThreadStart(FsyncThread *th, void (*on_complete)(void *result));
 void fsyncThreadAddTask(FsyncThread *th, int fd, raft_index_t requested_index);
 void fsyncThreadWaitUntilCompleted(FsyncThread *th);
 
+typedef enum {
+    RAFT_DEBUG_MIGRATION_NONE                    = 0,
+    RAFT_DEBUG_MIGRATION_EMULATE_CONNECT_FAILED,
+    RAFT_DEBUG_MIGRATION_EMULATE_IMPORT_FAILED,
+    RAFT_DEBUG_MIGRATION_EMULATE_UNLOCK_FAILED,
+    RAFT_DEBUG_MIGRATION_MAX,
+} migration_debug;
```
tezc, line 317:
```suggestion
} MigrationDebug;
```

Style

src/redisraft.h
```diff
@@ -308,6 +308,13 @@ void fsyncThreadStart(FsyncThread *th, void (*on_complete)(void *result));
 void fsyncThreadAddTask(FsyncThread *th, int fd, raft_index_t requested_index);
 void fsyncThreadWaitUntilCompleted(FsyncThread *th);
 
+typedef enum {
+    RAFT_DEBUG_MIGRATION_NONE                    = 0,
+    RAFT_DEBUG_MIGRATION_EMULATE_CONNECT_FAILED,
+    RAFT_DEBUG_MIGRATION_EMULATE_IMPORT_FAILED,
+    RAFT_DEBUG_MIGRATION_EMULATE_UNLOCK_FAILED,
```
tezc, line 315:
```suggestion
    DEBUG_MIGRATION_EMULATE_UNLOCK_FAILED,
```

We use `RAFT_` and `raft_` namespace for raft library. Probably, it's better to get rid of the prefix just to avoid confusion/collusion.

tests/integration/test_migrate.py
```diff
@@ -135,9 +136,88 @@ def test_happy_migrate(cluster_factory):
...
+        with raises(ResponseError, match=f"MOVED {slot} localhost"):
+            cluster2.leader_node().client.get(key_name)
+
+        conn = cluster2.leader_node().client.connection_pool.get_connection('deferred')
+        conn.send_command('ASKING')
+        assert conn.read_response() == b'OK'
+        conn.send_command('get', key_name)
+        assert conn.read_response() == value
+
+    cluster1.execute("raft.debug", "migration_debug", 1)
```
tezc, line 217:
```suggestion
    cluster1.execute("raft.debug", "migration_debug", 'fail_connect')
```

```c
        int val;
        
        size_t len;
        const char *str = RedisModule_StringPtrLen(argv[2], &len);
        if (!strncasecmp(str, "fail_connect", len)) {
            val = DEBUG_MIGRATION_EMULATE_CONNECT_FAILED;
        } else if (!strncasecmp(str, "fail_import", len)) {
            val = DEBUG_MIGRATION_EMULATE_IMPORT_FAILED;
        } else if (!strncasecmp(str, "fail_unlock", len)) {
            val = DEBUG_MIGRATION_EMULATE_UNLOCK_FAILED;
        } else {
            RedisModule_ReplyWithError(ctx, "ERR invalid migration debug value");
            return REDISMODULE_OK;
        }
```

You can accept strings if you want, so it'd be easier to understand inside the test file. Just an idea.

tezc: I didn't try the code btw, fyi

tezc (approved):

PR #389, opened 2022-06-15
Add RAFT.DEBUG DELAY_APPLY

tezc:
Debug option to slow down execution of commands. Just for testing purposes.

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -124,6 +124,10 @@ void RaftExecuteCommandArray(RedisModuleCtx *ctx,
     RedisRaftCtx *rr = &redis_raft;
     RedisModuleCallReply *reply;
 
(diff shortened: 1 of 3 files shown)


src/raft.c
```diff
@@ -124,6 +124,10 @@ void RaftExecuteCommandArray(RedisModuleCtx *ctx,
     RedisRaftCtx *rr = &redis_raft;
     RedisModuleCallReply *reply;
 
+    if (rr->debug_delay_apply) {
```
fadidahanna, line 127: Don't we want to call it before each command exec?

tezc: I want to call it before each entry/read request. If the request is MULTI, still I want to call it once. So, this is per entry slow down.

fadidahanna (approved):

PR #393, opened 2022-06-16
tweak to restore error handling to help debug failures

sjpotter:
before we just aborted on failure, as didn't expect it to occur, as its occuring regularly, need output on what the error is to determine why.

### DIFF
--- a/src/migrate.c
+++ b/src/migrate.c
@@ -90,7 +90,7 @@ void importKeys(RedisRaftCtx *rr, raft_entry_t *entry)
 
         enterRedisModuleCall();
         RedisModuleCallReply *reply;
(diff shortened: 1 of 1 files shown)


src/migrate.c
```diff
@@ -90,7 +90,7 @@ void importKeys(RedisRaftCtx *rr, raft_entry_t *entry)
 
         enterRedisModuleCall();
         RedisModuleCallReply *reply;
-        RedisModule_Assert((reply = RedisModule_Call(rr->ctx, "restore", "v", temp, 3)) != NULL);
+        RedisModule_Assert((reply = RedisModule_Call(rr->ctx, "restore", "v",git  temp, 4)) != NULL);
```
tezc, line 93:
```suggestion
        RedisModule_Assert((reply = RedisModule_Call(rr->ctx, "restore", "v", temp, 4)) != NULL);
```

Issue #386, opened 2022-06-14
Consider adding Valgrind build


tezc:
RedisRaft already has address sanitizer build but there is one case that address sanitizer doesn't catch: Use of uninitialized memory. 

**Solution-1**: Memory sanitizer

There is "memory sanitizer" for this case but GCC does not support it. Clang supports it but we need to add memory sanitizer to Redis first. So, to use this sanitizer, we need to cleanup Redis code base from memory sanitizer issues and then, add it to the RedisRaft. Also, memory sanitizer future is not clear whether it'll exist in future.

**Solution-2**: Valgrind

Probably, it is easier to add Valgrind to RedisRaft. We can already run test suite with valgrind by passing `--valgrind` config to the pytest but we need to surpress some errors and catch failures that we want. Downside is it will cause stability issues as valgrind slows down the execution a lot, we might see lots of test failures.

fadidahanna:
JIRA ticket: https://redislabs.atlassian.net/browse/RR-295

PR #394, opened 2022-06-16
Delete unused files

tezc:
Delete unused files


fadidahanna (commented):
I'm not sure we want to remove it, maybe to update.

tezc:
@fadidahanna If someone wants to add valgrind build, I see some value on it. (https://github.com/RedisLabs/redisraft/issues/386) It's not that easy though, requires some work. This file won't help you because it's not accurate. There is no point keeping it until we add valgrind to the CI. 

For the other file, it's not accurate as well. I'm also %100 against it as it relies on your function names, flow control etc. It'll become invalid each time you change something in the code. (Time has proven it, it's been invalid for the last 2 years). I don't think we have time to maintain it. Also, I'm not sure what it brings to the project, so, I don't see any value in it.

fadidahanna:
@tezc, suggest discussing that on a wider forum, adding it to our upcoming meeting agenda.

sjpotter (approved):

tezc:
@fadidahanna ping ping. Added one more file (I like when we delete things).

fadidahanna:
@tezc - agree with deleting the makefile, can we please do it in a separate PR? I still want to keep the other file as part of our documentation (and hopefully soon we can update them) 🙏

fadidahanna:
@tezc, as was discussed today, we can remove these files, and consider moving to mermaid diagrams for flow documentation.

fadidahanna (approved):

PR #395, opened 2022-06-16
Fix tls and sanitizer build on MacOS

tezc:
Added tls and sanitizer compile on MacOS. 

We are not running tests on for this build because MacOS machines are relatively slower than Linux and TLS + Sanitizer makes it even more slower, causing test failures. Leaving this as a future task. For now, we just verify RedisRaft builds with TLS and Sanitizer on MacOS.

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -151,6 +151,38 @@ jobs:
           cd build
           make tests
 
+  test-macos-tls-sanitizer-compile:
+    runs-on: macos-latest
+    if: |
+      (github.event_name == 'workflow_dispatch' ||
+      (github.event_name == 'schedule' && github.repository == 'redislabs/redisraft')) && !contains(github.event.inputs.skipjobs, 'macos')
+    timeout-minutes: 14400
+    steps:
+      - name: prep
+        if: github.event_name == 'workflow_dispatch'
+        run: |
+          echo "GITHUB_REPOSITORY=${{github.event.inputs.use_repo}}" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


tezc:
@fadidahanna With this change, sanitizer and TLS build works on MacOS (on github CI). Let me know if there are other issues.

fadidahanna (commented):
@tezc, did you test that on Mac?

tezc:
@fadidahanna Yes, it works on MacOS on Github CI. Do you get more errors on your local? :)

tezc:
I tried to add sanitizer + TLS on MacOS build to the CI, so we wouldn't break MacOS support but MacOS machines are slower than Linux. Sanitizer+TLS makes it slower and some tests are failing because they rely on timing(I didn't investigate fully). Maybe, we spend some time later on this, if we manage to get it work, you wouldn't get MacOS issues on your local I guess.

tezc:
Hmm, maybe we add just build to the CI. MacOS + Sanitizer + TLS build without running tests. At least, we'll know build works, otherwise it might be annoying for you to get build failures over time just because we don't build sanitizer + tls on MacOS.

tezc:
https://github.com/tezc/redisraft/runs/6954482956?check_suite_focus=true

Added compile for MacOS, hope it helps to detect issues in future.

fadidahanna:
Thanks @tezc, we can continue with enabling testing for MacOS later on, I have an ubuntu machine that I can use for testing. 🙏

fadidahanna (approved):

PR #397, opened 2022-06-19
Version check for Redis 7.0

tezc:


src/redisraft.c
```diff
@@ -1653,29 +1653,11 @@ __attribute__((__unused__)) int RedisModule_OnLoad(RedisModuleCtx *ctx, RedisMod
...
-
-    /* Sanity check Redis version */
-    if (RedisModule_SubscribeToServerEvent == NULL ||
-            RedisModule_RegisterCommandFilter == NULL ||
-            RedisModule_GetCommandKeys == NULL ||
-            RedisModule_GetDetachedThreadSafeContext == NULL ||
-            RedisModule_MonotonicMicroseconds == NULL) {
-        RedisModule_Log(ctx, REDISMODULE_LOGLEVEL_NOTICE,
-                        "RedisRaft requires Redis build from unstable branch!");
+    const int v7_0_0 = 0x00070000;
```
fadidahanna, line 1656: Let's have it as a global, and rename it to something like -  `MIN_REDIS_SUPPORTED_VERSION`

tezc: We are not using this variable anywhere else and I doubt if we ever going to use it. So, defining it where we use makes reading easier. Otherwise, you need to jump somewhere else to see the actual value.

tezc: Most of the time (probably like %99 of the cases), you don't need to define variables for literals/constants because you use them only once. Just because of a possible future minor problem, we can't make code less readable. So, I'm against that to be honest. I've renamed the variable but keeping it local for this reason.

PR #401, opened 2022-06-19
wasn't freeing RM_Call reply in key serialization correctly

sjpotter:


src/migrate.c
```diff
@@ -329,6 +329,7 @@ void MigrateKeys(RedisRaftCtx *rr, RaftReq *req)
 
             if (reply && RedisModule_CallReplyType(reply) == REDISMODULE_REPLY_STRING) {
                 req->r.migrate_keys.keys_serialized[i] = RedisModule_CreateStringFromCallReply(reply);
+                RedisModule_FreeCallReply(reply);
```
tezc, line 332:
```c
        if (!reply || RedisModule_CallReplyType(reply) != REDISMODULE_REPLY_STRING) {
            RedisModule_ReplyWithError(req->ctx, "ERR dump failed");
            if (reply) {
                LOG_WARNING("unexpected response type = %d", RedisModule_CallReplyType(reply));
            }
            goto exit;
        }

        req->r.migrate_keys.keys_serialized[i] = RedisModule_CreateStringFromCallReply(reply);
```
Inverting check might be easier on eyes (success path on the main flow), maybe helps with catching these kind of issues, just an idea.

sjpotter: used

PR #403, opened 2022-06-19
now that shardgroup replace requires session migration key, need it migration

sjpotter:
we were setting it to 0 automatically, but that broke once the shardgroup replace pr went in.

this fetches the value for use during migration

### DIFF
--- a/src/migrate.c
+++ b/src/migrate.c
@@ -9,11 +9,18 @@ static bool validSlotMigrationSessionKey(RedisRaftCtx *rr, unsigned int slot, un
 
     for (unsigned int i = 0; i < sg->slot_ranges_num; i++) {
         ShardGroupSlotRange *sr = &sg->slot_ranges[i];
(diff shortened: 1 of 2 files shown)


src/migrate.c
```diff
@@ -316,6 +339,9 @@ void MigrateKeys(RedisRaftCtx *rr, RaftReq *req)
         goto exit;
     }
     req->r.migrate_keys.migrate_term = raft_get_current_term(rr->raft);
+    if (getMigrationSessionKey(rr, req, &req->r.migrate_keys.migration_session_key) != RR_OK) {
+        goto exit;
```
tezc, line 343: Should we reply here?

sjpotter: we should, I meant to include it in the function itself.  thought if better inside or out?

tezc: I see you already added. That makes sense because you also reply on some other places inside MigrateKeys(), not inside another sub function. So, I guess now it is inline with the rest of the function.

tezc (approved):

PR #405, opened 2022-06-20
Populate command specs by Redis introspection.

yossigo:
After this change, only `CMD_SPEC_UNSUPPORTED` and `CMD_SPEC_DONT_INTERCEPT` are explicitly specified for core Redis commands. The other flags are derived on the fly from Redis command introspection.

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -13,68 +13,128 @@
 
 static RedisModuleDict *commandSpecDict = NULL;
 
+/* Look up the specified command in the command spec table and return the
+ * CommandSpec associated with it. If create is true, a new entry will
+ * be created if one does not exist. Otherwise, NULL is returned.
+ */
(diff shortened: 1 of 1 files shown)


src/commands.c
```diff
@@ -13,68 +13,142 @@
...
+
+    if (lcmd != buf) {
+        RedisModule_Free(lcmd);
+    }
+
+    return cs;
+}
+
+/* Use COMMAND to fetch all Redis commands and update the CommandSpec. */
+static void CommandSpecPopulateFromRedis(RedisModuleCtx *ctx)
```
tezc, line 54:
```suggestion
static void commandSpecPopulateFromRedis(RedisModuleCtx *ctx)
```

tezc (dismissed):

tezc (approved):

PR #409, opened 2022-06-21
Refactor key validation

tezc:
I mentioned this before, so just to be clear, here is my suggestion. I realized that we don't need enum, when we take it out, decision making part becomes simpler (the part we decide if it is successful or whether we should send ask / tryagain). Otherwise, it just takes a while for me to understand the switch condition there. Do you find this alternative easier to understand or is it just me? :) Wdyt?

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -160,8 +160,12 @@ static RRStatus validateKeyExistence(RedisModuleCtx *ctx,
         for (int j = 0; j < num_keys; j++) {
             int nokey; /* Key should not be locked */
             RedisModule_DictGet(rr->locked_keys, cmd->argv[keyindex[j]], &nokey);
+            if (!nokey) {
+                RedisModule_Free(keyindex);
+                goto reply_tryagain;
+            }
 
-            if (nokey && RedisModule_KeyExists(rr->ctx, cmd->argv[keyindex[j]])) {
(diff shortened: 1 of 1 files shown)


src/raft.c
```diff
@@ -178,8 +137,57 @@ static ShardGroup *GetSlotShardGroup(RedisRaftCtx *rr, unsigned int slot, bool a
...
+    }
+
+    /* There are missing keys, reply an error */
+    if (ctx) {
+        if (num_keys_found == 0 && slot_type == SLOTRANGE_TYPE_MIGRATING) {
+            replyAsk(rr, ctx, slot);
+        } else {
+            RedisModule_ReplyWithError(ctx, "TRYAGAIN");
+        }
+    }
```
sjpotter, line 182: this isn't correct.  num_keys_found can == 0 with a locked key.  we don't reply ASK then.

sjpotter:
I'm also unsure if validateKeyExistence should be what's returning response to user.  In practice, the current mechansim overloads SomeExist with a locked key, perhaps we should have a new enum values (LOCKED_KEYS).

personally, I dont find the switch to be bad, as it makes the logic (to me) clear, its not hidden.  We determine what the key situation is and then based on the key situation and the type of slot we are dealing with, we return an answer.

src/raft.c
tezc: oops, right

PR #412, opened 2022-06-22
More comments + clean up locked key status

sjpotter:
missed a few comments + added a new LockedExist state to make that clearer and not overload SomeExist.

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -109,12 +109,14 @@ typedef enum KeysStatus{
     AllExist,
     SomeExist,
     NoneExist,
+    LockedExist,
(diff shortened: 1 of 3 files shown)


src/raft.c
```diff
@@ -109,12 +109,14 @@ typedef enum KeysStatus{
     AllExist,
     SomeExist,
     NoneExist,
+    LockedExist,
```
tezc, line 112: Could you use uppercase? `LOCKED_EXIST`, `ALL_EXIST` etc.

tezc (approved):

PR #413, opened 2022-06-22
did a bit of refactoring of handleRedisCommand path

sjpotter:
based on review that the functions were being treated as a void, did a bit of a refactor to try and clean it up

### DIFF
--- a/src/redisraft.c
+++ b/src/redisraft.c
@@ -658,9 +658,15 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
 
     if (!raft_is_leader(rr->raft)) {
         handleNonLeaderCommand(rr, ctx, cmds);
-    } else if (cmd_flags & CMD_SPEC_UNSUPPORTED) {
+        return;
+    }
+
+    if (cmd_flags & CMD_SPEC_UNSUPPORTED) {
         RedisModule_ReplyWithError(ctx, "ERR not supported by RedisRaft");
-    } else if (cmd_flags & CMD_SPEC_READONLY && !(cmd_flags & CMD_SPEC_WRITE)) {
+        return;
+    }
+
+    if (cmd_flags & CMD_SPEC_READONLY && !(cmd_flags & CMD_SPEC_WRITE)) {
         handleRedisCommandAppendReadOnly(rr, ctx, cmds);
     } else {
         handleRedisCommandAppendLog(rr, ctx, cmds);
@@ -711,12 +717,18 @@ static void handleNonLeaderCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRe
     if (!leader) {
         /* if no leader, cluster is down */
         replyClusterDown(ctx);
-    } else if (rr->config->follower_proxy) {
+        return;
+    }
+
+    if (rr->config->follower_proxy) {
         /* forward commands to the leader in follower_proxy state */
         if (ProxyCommand(rr, ctx, cmds, leader) != RR_OK) {
             RedisModule_ReplyWithError(ctx, "NOTLEADER Failed to proxy command");
         }
-    } else if (cmds->asking) {
+        return;
+    }
+
+    if (cmds->asking) {
         /* reply ASK for ASKING state commands */
         int slot;
         RRStatus res = HashSlotCompute(rr, cmds, &slot);
@@ -728,10 +740,11 @@ static void handleNonLeaderCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRe
         } else {
             replyAsk(rr, ctx, (unsigned int) slot);
         }
-    } else {
-        /* redirect otherwise */
-        redirectCommand(rr, ctx, leader);
+        return;
     }
+
+    /* redirect otherwise */
+    redirectCommand(rr, ctx, leader);
 }
 
 /* RAFT [Redis command to execute]
--- a/src/redisraft.c
+++ b/src/redisraft.c
@@ -38,7 +38,7 @@ void *__dso_handle;
 #define VALID_NODE_ID(x)    ((x) > 0)
 
 static void redirectCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, Node *leader);
-static void handleNonLeaderCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, Node *leader, RaftRedisCommandArray *cmds);
+static void handleNonLeaderCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRedisCommandArray *cmds);
 
 /* Parse a node address from a RedisModuleString */
 static RRStatus getNodeAddrFromArg(RedisModuleCtx *ctx, RedisModuleString *arg, NodeAddr *addr)
@@ -594,10 +594,6 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
 /* Handle the special case of read-only commands here: if quorum reads
  * are enabled schedule the request to be processed when we have a guarantee
  * we're still a leader. Otherwise, just process the reads.
- *
- * Normally we can expect a single command in the request, unless it is a
- * MULTI/EXEC transaction in which case all queued commands are handled at
- * once.
  */
 static void handleRedisCommandAppendReadOnly(RedisRaftCtx *rr, RedisModuleCtx *ctx,
                                              RaftRedisCommandArray *cmds)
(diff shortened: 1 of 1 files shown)


src/redisraft.c
```diff
@@ -38,7 +38,7 @@ void *__dso_handle;
 #define VALID_NODE_ID(x)    ((x) > 0)
 
 static void redirectCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, Node *leader);
-static RRStatus handleNonLeaderCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx,  Node *leader, RaftRedisCommandArray *cmds);
+static void handleNonLeaderCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx,  Node *leader, RaftRedisCommandArray *cmds);
```
tezc, line 41:
```suggestion
static void handleNonLeaderCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, Node *leader, RaftRedisCommandArray *cmds);
```

src/redisraft.c
```diff
@@ -591,76 +591,79 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
...
-            if (term != rr->snapshot_info.last_applied_term) {
-                replyClusterDown(ctx);
-                return RR_ERROR;
-            }
-
+/* Handle the special case of read-only commands here: if quorum reads
+ * are enabled schedule the request to be processed when we have a guarantee
+ * we're still a leader. Otherwise, just process the reads.
+ *
+ * Normally we can expect a single command in the request, unless it is a
```
tezc, line 598: This comment belongs to where we call `CommandSpecGetAggregateFlags()`

src/redisraft.c
```diff
@@ -591,76 +591,79 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
...
+ * MULTI/EXEC transaction in which case all queued commands are handled at
+ * once.
+ */
+static void handleRedisCommandAppendReadOnly(RedisRaftCtx *rr, RedisModuleCtx *ctx,
+                                             RaftRedisCommandArray *cmds) {/* read only command, quorum or not */
```
tezc, line 603:
```suggestion
static void handleRedisCommandAppendReadOnly(RedisRaftCtx *rr, RedisModuleCtx *ctx,
                                             RaftRedisCommandArray *cmds)
{
```

src/redisraft.c
```diff
@@ -591,76 +591,79 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
...
+        RaftReq *req = RaftReqInit(ctx, RR_REDISCOMMAND);
+        RaftRedisCommandArrayMove(&req->r.redis.cmds, cmds);
+        raft_queue_read_request(rr->raft, handleReadOnlyCommand, req);
+    } else {
+        /* Wait until the new leader applies an entry from the current term.
+         * Otherwise, we might process a request before replaying logs.
+         * The state machine will be in an older state. Reading from it
+         * might look like going backward in time. */
+        raft_term_t term = raft_get_current_term(rr->raft);
+        if (term == rr->snapshot_info.last_applied_term) {
```
tezc, line 615: I think this makes code less readable. Common pattern is having an if check for error case and returning inside it. So, you keep success path in the main flow. Previous version of this code is easier to understand.

src/redisraft.c
```diff
@@ -591,76 +591,79 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
...
             RaftExecuteCommandArray(rr, ctx, ctx, cmds);
+        } else {
+            replyClusterDown(ctx);
         }
-        return RR_OK;
     }
+}
 
+static void handleRedisCommandAppendLog(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRedisCommandArray *cmds) {
```
tezc, line 623:
```suggestion
static void handleRedisCommandAppendLog(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRedisCommandArray *cmds)
{
```

src/redisraft.c
```diff
@@ -591,76 +591,79 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
...
-        return RR_ERROR;
+    } else {
+        raft_entry_release(entry);
     }
+}
 
-    raft_entry_release(entry);
-    return RR_OK;
+static void handleRedisCommandAppend(RedisRaftCtx *rr,
+                                      RedisModuleCtx *ctx, RaftRedisCommandArray *cmds)
```
tezc, line 644:
```suggestion
                                     RedisModuleCtx *ctx, RaftRedisCommandArray *cmds)
```

src/redisraft.c
```diff
@@ -591,76 +591,79 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
...
+     * joining or loading data.
+     */
+    if (checkRaftState(rr, ctx) == RR_ERROR) {
+        return;
+    }
+
+    if (!raft_is_leader(rr->raft)) {
+        Node *leader = getLeaderNodeOrReply(rr, ctx);
+
+        return handleNonLeaderCommand(rr, ctx, leader, cmds);
```
tezc, line 658: A bit weird to return with a void function

src/redisraft.c
sjpotter: ok, changed back, was trying to avoid extra returns

src/redisraft.c
```diff
@@ -591,76 +591,79 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
...
+
+    /* Check that we're part of a bootstrapped cluster and not in the middle of
+     * joining or loading data.
+     */
+    if (checkRaftState(rr, ctx) == RR_ERROR) {
+        return;
+    }
+
+    if (!raft_is_leader(rr->raft)) {
+        Node *leader = getLeaderNodeOrReply(rr, ctx);
```
tezc, line 656: We are not doing anything with it here, we can call this inside `handleNonLeaderCommand()`.

src/redisraft.c
sjpotter: the other option was

```
handleNonLeaderCommand(rr, ctx, leader, cmds);
return;
```

is that better?

sjpotter: combined with the next if/else block to avoid need for return

src/redisraft.c
```diff
@@ -591,76 +591,80 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
...
+
     int e = raft_recv_entry(rr->raft, entry, &req->r.redis.response);
     if (e != 0) {
         replyRaftError(ctx, e);
         entryDetachRaftReq(rr, entry);
         raft_entry_release(entry);
         RaftReqFree(req);
-        return RR_ERROR;
+    } else {
+        raft_entry_release(entry);
```
tezc, line 639: As both paths are calling `raft_entry_release(entry)`, we can call this unconditionally but I think I find more readable when we separate error paths from the success path and keep success path in the main flow. (Not in an "if check" if possible).  Previously, when we get an error, we were calling bunch of functions and we were returning from the error case. I find it easier to understand.

sjpotter: ok, reverted else and put void return back in

src/redisraft.c
```diff
@@ -697,42 +701,36 @@ static void redirectCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, Node *leader)
...
-static RRStatus handleNonLeaderCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, Node *leader, RaftRedisCommandArray *cmds)
+static void handleNonLeaderCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, Node *leader, RaftRedisCommandArray *cmds)
 {
     RedisModule_Assert(!raft_is_leader(rr->raft));
     RedisModule_Assert(cmds);
 
-    /* reply ASK for ASKING state commands */
-    if (cmds->asking) {
+
+    if (!leader) {
```
tezc, line 711: Going back to same discussion :) I'd prefer previous version. I find it harder to understand when there are bunch of if checks. It's common to have "fallthrough" if checks for error cases/conditions and I find it easier to read:

```c
if (!leader) {
    .....
    return; 
}

if (rr->config->follower_proxy) {
    /* forward commands to the leader in follower_proxy state */
   .......
    return;
}

if (asking) {
    ......
    return;
}

redirect();
return;
```

src/redisraft.c
```diff
@@ -591,76 +591,80 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
...
+        return handleNonLeaderCommand(rr, ctx, leader, cmds);
+    }
+
+    if (cmd_flags & CMD_SPEC_UNSUPPORTED) {
+        RedisModule_ReplyWithError(ctx, "ERR not supported by RedisRaft");
+    } else if (cmd_flags & CMD_SPEC_READONLY && !(cmd_flags & CMD_SPEC_WRITE)) {
+        handleRedisCommandAppendReadOnly(rr, ctx, cmds);
+    } else {
+        handleRedisCommandAppendLog(rr, ctx, cmds);
+    }
```
tezc, line 667:
```suggestion
    if (cmd_flags & CMD_SPEC_UNSUPPORTED) {
        RedisModule_ReplyWithError(ctx, "ERR not supported by RedisRaft");
        return;
    } 
    
    if (cmd_flags & CMD_SPEC_READONLY && !(cmd_flags & CMD_SPEC_WRITE)) {
        handleRedisCommandAppendReadOnly(rr, ctx, cmds);
    } else {
        handleRedisCommandAppendLog(rr, ctx, cmds);
    }
```

For example, I find this one easier to understand.

src/redisraft.c
sjpotter: ok. redid

src/redisraft.c
sjpotter: ok, done

src/redisraft.c
sjpotter: reverted above comment's change.

tezc (approved):

PR #425, opened 2022-06-26
fix crash on shardgroup replace command

sjpotter:
was freeing non existent shardgroups as wasn't tracking them correctly when added

+ add tests to validate that invalid shardgroups error out correctly

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -985,7 +985,7 @@ ShardGroup **ShardGroupsParse(RedisModuleCtx *ctx,
     argc--;
     int argv_idx = 1;
 
-    int created_shard_count = 0;
+    unsigned int created_shard_count = 0;
     for (int i = 0; i < num_shards; i++) {
         ShardGroup *sg;
 
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -214,13 +214,10 @@ ShardGroup *ShardGroupCreate() {
     return sg;
 }
 
+/* Unlike C heap function, this should only be passed non NULL (valid) pointers */
 void ShardGroupFree(ShardGroup *sg) {
-    if (!sg) {
-        /* FIXME: Would it be good to have a non fatal backtrace be printed here.
-         * Calling ShardGroupFree on a a null seems to be a logic error
-         */
-        return;
-    }
+    RedisModule_Assert(sg != NULL);
+
     ShardGroupTerm(sg);
     RedisModule_Free(sg);
 }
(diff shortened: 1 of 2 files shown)


tezc:
I think you can add null check to `ShardGroupFree()` instead of tracking them seperately. It might save us now/in future if we have similar assumptions somewhere else. (Maybe we call ShardGroupFree() on some other NULL pointers somewhere?).

```c
void ShardGroupFree(ShardGroup *sg) 
{
    if (!sg) {
       return;
    }
    ShardGroupTerm(sg);
    RedisModule_Free(sg);
}
```

sjpotter:
> I think you can add null check to `ShardGroupFree()` instead of tracking them seperately. It might save us now/in future if we have similar assumptions somewhere else. (Maybe we call ShardGroupFree() on some other NULL pointers somewhere?).
> 
> ```c
> void ShardGroupFree(ShardGroup *sg) 
> {
>     if (!sg) {
>        return;
>     }
>     ShardGroupTerm(sg);
>     RedisModule_Free(sg);
> }
> ```

while I have no problem with defense in depth like that, I think it perhaps goes to a logic problem if we would call it on a null pointer.

perhaps a combination where we also print a backtrace when that occurs?  Do we have reusable code to print a backtrace?

tezc:
As `free()` accepts NULL pointers, I think it makes sense to make some free-like functions to accept NULL pointers. (Best way is avoiding allocation but leaving this discussion aside). This approach is useful when you know you don't have a logic problem. Then, it saves you tracking the allocation or adding if checks before calling `free()`. Anyway, if you feel like it will hide a problem, then sure, just keep the PR as it is. 

Regarding backtrace, not sure I understand. Do you want to print bracktrace unconditionally? `RedisModule_Assert(0);` should work.

sjpotter:
I'd want to print a backtrace but not crash (i.e. handle the free as you said, but warn that there is an issue elsewere.

tezc:
This is becoming complicated :) I think you are mixing two cases. Either it should be valid or not valid. It is a bit weird to accept NULL pointer and still print a warning just in case something is wrong. We should pick one behavior. If verification is required, it must be done somewhere else. 

Accepting NULL inside free()-like functions saves you writing boilerplate code(like in this PR) but if we are not sure our logic is correct, then maybe we shouldn't do it for now. Anyway, it's up to you, pick whatever you like best.

src/cluster.c
```diff
@@ -1064,7 +1068,7 @@ ShardGroup **ShardGroupsParse(RedisModuleCtx *ctx,
 fail:
     *len = 0;
 
-    for (int j = 0; j <= i; j++) {
+    for (unsigned int j = 0; j < created_shard_count; j++) {
```
tezc, line 1071:
```suggestion
    for (int j = 0; j < created_shard_count; j++) {
```
I think linter will complain about this

sjpotter: hmm, I mean for the created_shard_count to be unsigned.  will fix it there

PR #426, opened 2022-06-26
fix snapshotting shardgroups

sjpotter:
we weren't saving the migration session key and the max_importing_term array

### DIFF
--- a/tests/integration/test_migrate.py
+++ b/tests/integration/test_migrate.py
@@ -49,6 +49,51 @@ def test_raft_import(cluster):
     assert conn.read_response() == b'value'
 
 
+def test_import_with_snapshot(cluster):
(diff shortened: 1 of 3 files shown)


tests/integration/test_migrate.py
```diff
@@ -49,6 +49,51 @@ def test_raft_import(cluster):
...
+
+    cluster.node(1).kill()
+    cluster.node(1).start()
+    cluster.node(1).wait_for_info_param('raft_state', 'up')
+
+    # incorrect session migration key
+    with raises(ResponseError, match="invalid migration_session_key"):
+        cluster.execute('raft.import', '2', '0', 'key', serialized)
+
+    # older term than should be accepted
```
tezc, line 89:
```suggestion
    # older term should not be accepted
```

tezc (approved):

PR #428, opened 2022-06-28
Change CLUSTER NODES and CLUSTER SLOTS replies according to the new migration support

fadidahanna:


sjpotter:
went over the code, didn't see anything, will look at the tests later

tests/integration/test_sharding.py
```diff
@@ -583,3 +623,329 @@ def test_asking_follower(cluster):
...
+                assert node_data[8] == b"0-500"
+                assert node_data[9] == b"800-1000"
+            elif node_data[0] == node_id_2:
+                assert node_data[8] == b"600-700"
+            elif node_data[0] == node_id_3:
+                assert node_data[8] == b"1001-16383"
+            else:
+                assert False, "failed to match id {}".format(node_data[0])
+
+    validate_nodes(cluster.node(1).execute('CLUSTER', 'NODES').splitlines())
```
tezc, line 951:
```suggestion
    validate_nodes(cluster.node(1).execute('CLUSTER', 'NODES').splitlines())

```
New line at the end.

tezc:
@fadidahanna So, we've concluded that we just skip "importing" nodes in the reply? The documentation describes something else for "cluster nodes" but I guess we don't need to follow it:  https://redis.io/commands/cluster-nodes/ 

Just out of curiosity, what is the reasoning behind this decision? Could you describe a bit in the top comment? Are we trying to prevent listing same slot twice?

tezc:
Sure, makes sense. Requests for importing slot without asking will be rejected anyway, so, clients can't use that info anyway.

tests/integration/test_sharding.py
```diff
@@ -583,3 +623,329 @@ def test_asking_follower(cluster):
...
+            if node_id == stable_node_id:
+                assert start_slot == 0, cluster_slots
+                assert end_slot == 0, cluster_slots
+            elif node_id == migrate_node_id:
+                assert start_slot == 501, cluster_slots
+                assert end_slot == 501, cluster_slots
+            else:
+                assert False, "failed to match id {}".format(node_id)
+
+    validate_slots(cluster.node(1).client.execute_command('CLUSTER', 'SLOTS'))
```
tezc, line 674: Btw, if you feel like using in future, just fyi, we have shorter alternatives to execute a command:

`cluster.execute('CLUSTER', 'SLOTS')`    (sends it to the leader node, retries command on failure)
or
`cluster.node(1).execute('CLUSTER', 'SLOTS')` .

tezc (approved):

tests/integration/test_sharding.py
sjpotter: except if you want to catch MOVED as an expected situation cluster.execute() handles MOVED itself

tests/integration/test_migrate.py
```diff
@@ -15,10 +15,12 @@ def test_raft_import(cluster):
     assert cluster.execute(
         'RAFT.SHARDGROUP', 'REPLACE',
         '2',
+
         '12345678901234567890123456789013',
         '1', '1',
         '0', '16383', '3', '123',
         '1234567890123456789012345678901334567890', '3.3.3.3:3333',
+
```
sjpotter, line 23: same here

tests/integration/test_migrate.py
```diff
@@ -15,10 +15,12 @@ def test_raft_import(cluster):
     assert cluster.execute(
         'RAFT.SHARDGROUP', 'REPLACE',
         '2',
+
```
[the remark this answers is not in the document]
sjpotter: actually it seems no test changes were made here, so perhaps this is just unnecessary?

tests/integration/test_sharding.py
```diff
@@ -148,24 +150,58 @@ def validate_slots(cluster_slots):
...
+            elif node_id == b'3' * 40:
                 assert start_slot == 8, cluster_slots
                 assert end_slot == 16383, cluster_slots
             else:
-                assert False, "failed to match id {}".format(shargroup_id)
+                assert False, "failed to match id {}".format(node_id)
 
     validate_slots(cluster.node(3).client.execute_command('CLUSTER', 'SLOTS'))
 
+    def validate_nodes(cluster_nodes):
```
sjpotter, line 171: based on this test / validation, might want to define a shardgroup with a single slot and no slots.  i.e. validate that the no slot doesn't exist.

sjpotter: see single slot validation below

tests/integration/test_migrate.py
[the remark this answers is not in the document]
fadidahanna: did it during reviewing the prev tests, and found it easy for me to read the commands with these breaks. Wondered if to do it all over our tests, and decided not to do it in this PR.

PR #429, opened 2022-06-29
Add basic migration test

tezc:
Added a test just to simulate migration process fully. Test generates some keys in the source cluster and moves half of the keys to the destination cluster by using raft.scan.

### DIFF
--- a/tests/integration/test_migrate.py
+++ b/tests/integration/test_migrate.py
@@ -127,7 +127,7 @@ def test_migration_basic(cluster_factory):
     assert cluster2.execute('dbsize') == key_count
     assert cluster2.node(1).execute('get', '{key}1') == b'value'
     with raises(ResponseError, match='MOVED 4998'):
-        cluster2.node(1).execute('get', '{key2}')
+        cluster2.node(1).execute('get', '{key2}1')
 
 
 def test_raft_import(cluster):
--- a/tests/integration/test_migrate.py
+++ b/tests/integration/test_migrate.py
@@ -2,6 +2,134 @@ from _pytest.python_api import raises
 from redis import ResponseError
 
 
+def test_migration_basic(cluster_factory):
+    """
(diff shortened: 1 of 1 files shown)


tests/integration/test_migrate.py
```diff
@@ -2,6 +2,134 @@
...
+    assert cluster1.execute('dbsize') == key_count
+    assert cluster1.node(1).execute('get', '{key2}1') == b'value'
+    with raises(ResponseError, match='MOVED 12539'):
+        cluster1.node(1).execute('get', '{key}1')
+
+    # cluster-2 has slots 8001-16383
+    assert cluster2.execute('dbsize') == key_count
+    assert cluster2.node(1).execute('get', '{key}1') == b'value'
+    with raises(ResponseError, match='MOVED 4998'):
+        cluster2.node(1).execute('get', '{key2}')
```
fadidahanna, line 130:
```suggestion
        cluster2.node(1).execute('get', '{key2}1')
```

tests/integration/test_migrate.py
```diff
@@ -2,6 +2,134 @@
...
+        'sharding': 'yes',
+        'external-sharding': 'yes'})
+
+    key_count = 11500
+
+    # Fill two specific slots with keys. One of these slots will be migrated.
+    for i in range(key_count):
+        cluster1.execute('set', '{key}' + str(i), 'value')   # slot 12539
+        cluster1.execute('set', '{key2}' + str(i), 'value')  # slot 4998
+
```
fadidahanna, line 24: maybe add a validation here on the size: 
```assert cluster1.execute('dbsize') == 2 * key_count```

fadidahanna: Right. I usually prefer to have guards before and after to catch an issue asap, but as you wish. `Assert`s are not only to validate our changes, but declare an assumption.

fadidahanna (approved):

Issue #240, opened 2022-02-03
Limit maximum size of appendentries message


tezc:
Limit maximum size of appendentries message and make it configurable.

fadidahanna:
Waiting for test PR to close this issue. @tezc

PR #435, opened 2022-07-04
Limit appendentries max size

tezc:
Add support to limit appendentries message size

Currently, there is no limit on appendentries payload size. If there are many entries to be sent to the followers, we create a huge appendentries message. This might cause performance issues or even OOM.

Using the callback `get_entries_to_send()` introduced in https://github.com/RedisLabs/raft/pull/101, this PR adds support to limit appendentries message size. Rather than creating a single large appendentries message, RedisRaft will send multiple smaller appendentries messages. 

Configuration parameter changes: 
- Added `append-req-max-size` configuration. RedisRaft will try to create append req messages smaller than this limit. This is just a best-effort limit. (e.g if a single entry is larger than the limit, the message will be larger than the limit. ) 
- Renamed existing configuration parameter `max-append-req-in-flight` to `append-req-max-count` (just for clarity and to have common prefix in related configs). 

Info output changes:
- Added `stats` sections and two other metrics, required for testing but can be useful in general:
```
# raft_stats
raft_appendreq_received:0               /* Number of received appendreq messages */
raft_appendreq_with_entry_received:0    /* Number of received appendreq messages with at least one entry */
```

Fixes: https://github.com/RedisLabs/redisraft/issues/240

### DIFF
--- a/src/redisraft.c
+++ b/src/redisraft.c
@@ -901,6 +901,9 @@ static int cmdRaftAppendEntries(RedisModuleCtx *ctx, RedisModuleString **argv, i
         msg.entries[i] = e;
     }
 
+    rr->appendreq_received++;
+    rr->appendreq_with_entry_received += (msg.n_entries > 0) ? 1 : 0;
+
     raft_appendentries_resp_t resp = {0};
     raft_node_t *node = raft_get_node(rr->raft, src_node_id);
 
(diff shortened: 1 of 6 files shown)


tezc:
I'll have an idea for testing this feature, I'll update the PR soon.

src/redisraft.c
```diff
@@ -1756,6 +1759,10 @@ static void handleInfo(RedisModuleInfoCtx *ctx, int for_crash_report)
     RedisModule_InfoAddFieldULongLong(ctx, "proxy_failed_reqs", rr->proxy_failed_reqs);
     RedisModule_InfoAddFieldULongLong(ctx, "proxy_failed_responses", rr->proxy_failed_responses);
     RedisModule_InfoAddFieldULongLong(ctx, "proxy_outstanding_reqs", rr->proxy_outstanding_reqs);
+
+    RedisModule_InfoAddSection(ctx, "stats");
+    RedisModule_InfoAddFieldULongLong(ctx, "appendreq_received", rr->appendreq_received);
```
tezc, line 1764: Not sure about naming, I'm open to suggestions. 

We should expand this pattern to other messages over time. Stats for these might be useful:

appendreq
appendresp
snapshotreq
snapshotresp
reqvotereq
revqoteresp

fadidahanna (approved):

PR #436, opened 2022-07-04
Delete unused variable

tezc:
We've moved this variable into RedisRaftConfig struct in https://github.com/RedisLabs/redisraft/pull/360. This is just a leftover, result of a rebase mistake.


sjpotter (approved):
assuming if this compiles that in tls mode that its correct, and just an artifact from older code

tezc:
Yes, just a leftover from a rebase error I've made in https://github.com/RedisLabs/redisraft/pull/360

PR #442, opened 2022-07-07
TLS CA directory support

fadidahanna:


.github/workflows/daily.yml
```diff
@@ -189,6 +192,9 @@ jobs:
       (github.event_name == 'workflow_dispatch' ||
       (github.event_name == 'schedule' && github.repository == 'redislabs/redisraft')) && !contains(github.event.inputs.skipjobs, 'tls')
     timeout-minutes: 14400
+    strategy:
+      matrix:
+        tls_mode: [ "None", "File", "Dir", "Both" ]
```
tezc, line 197: I think running whole suite with different TLS configurations doesn't add much value because it is still just TLS :) After loading the certificates/keys and connecting to each other, rest will be same anyway.

 I think we just want to test if we can configure TLS with a file, a directory or both. So, I suggest adding "dir" and "both" cases as a single test case. Maybe, we can pass a parameter to `cluster.create(tls_cacert_location)` and depending on this parameter, we can use `dir` or `file` or `both`.

So, something like:

```python
# Skip if not TLS
test_tls_cacert_dir(cluster):
   cluster.create(3, tls_cacert_location=dir)
   cluster.execute("set x 1")
   assert cluster.execute("get x") == 1

# Skip if not TLS
test_tls_cacert_both(cluster):
   cluster.create(3, tls_cacert_location=both)
   cluster.execute("set x 1")
   assert cluster.execute("get x") == 1
```

tezc: If `dir` will be used more, maybe we can make `dir` as the default option here and add single test cases for `file` and `both` cases.

docs/TLS.md
```diff
@@ -5,7 +5,7 @@ RedisRaft's TLS support mirrors Redis's native TLS support.
 One runs the underlying redis server in the same manner that one would run a redis server with TLS
 
 ```
-redis-server --tls-port <por> --port 0 --tls-cert-file <file> --tls-key-file <file> --tls-ca-cert-file <file> 
+redis-server --tls-port <por> --port 0 --tls-cert-file <file> --tls-key-file <file> --tls-ca-cert-file <file> | --tls-ca-cert-dir <dir>
```
tezc, line 8:
```suggestion
redis-server --tls-port <por> --port 0 --tls-cert-file <file> --tls-key-file <file> --tls-ca-cert-file <file> --tls-ca-cert-dir <dir> 
```

there is a `|` here

.github/workflows/daily.yml
fadidahanna: I like as much as possible tests, but sure, can have a small dedicated test only to verify connection creation.

tests/integration/test_sanity.py
```diff
@@ -429,3 +429,24 @@ def test_tls_reconfig(cluster):
...
+
+@pytest.mark.skipif("not config.getoption('tls')")
+def test_tls_ca_cert_file(cluster):
+    cluster.create(3, tls_ca_cert_location='file')
+    assert cluster.execute('set', 'key', 'value')
+    assert cluster.execute('get', 'key') == b'value'
+
+
+@pytest.mark.skipif("not config.getoption('tls')")
+def test_tls_ca_cert_file(cluster):
```
tezc, line 449: Should function name be different than above? It might be confusing if one fails other one succeeds, we may spend extra time to figure out which one failed. I'm surprised Python is okay with this :o

fadidahanna, line 449:
```suggestion
def test_tls_ca_cert_both(cluster):
```

fadidahanna:
Daily run: https://github.com/fadidahanna/redisraft/runs/7237117253?check_suite_focus=true

Issue #449, opened 2022-07-11
possible crash in raft.snapshot


sjpotter:
@danni-m saw a crash in what appears to be a RAFT.SNAPSHOT command, but it seems weird situation that I don't know how to diagnose.

[raft_crash-10-07-22.log](https://github.com/RedisLabs/redisraft/files/9084022/raft_crash-10-07-22.log)

tezc:
I think leader sends another message while we are loading the snapshot. 

1- Leader sends the last snapshot chunk and we call `raft_recv_snapshot()` -> `rdbLoad()`. Redis starts loading the data.
2- Leader sends another message.
3- As we didn't block the leader on step 1, Redis tries to process new message while loading rdb file and hits that assert. 

Probably this is not an explored path for Redis as`rdbLoad()` is not part of the module API.

After verifying this is the case, I think we can call `RM_BlockClient()` around `raft_recv_snapshot()` to prevent processing messages while loading the rdb.

PR #450, opened 2022-07-13
Block leader's connection while loading the snapshot

tezc:
Block leader's connection while loading the snapshot

`raft_recv_snapshot()` might trigger loading the received RDB file by calling `raftLoadSnapshot()`  and then `rdbLoad()`. While loading, Redis may process incoming messages to reply `-LOADING`. So, we must block the leader's connection to prevent processing further messages, as we haven't replied to the current command yet. Otherwise, it triggers an assert inside Redis.

This is just a workaround. We need to think about having `rdbLoad()` in the module API and revisit this issue. 

Fixes https://github.com/RedisLabs/redisraft/issues/449

### DIFF
--- a/src/redisraft.c
+++ b/src/redisraft.c
@@ -1000,9 +1000,15 @@ static int cmdRaftSnapshot(RedisModuleCtx *ctx, RedisModuleString **argv, int ar
 
     rr->snapshotreq_received++;
 
+    /* raft_recv_snapshot() might trigger loading the received RDB file.
+     * While loading, Redis may process incoming messages to reply -LOADING.
+     * So, we must block the leader's connection to prevent processing further
+     * messages, as we haven't replied to the current command yet. */
+    RedisModuleBlockedClient *c = RedisModule_BlockClient(ctx, NULL, 0, 0, 0);
+
     if (raft_recv_snapshot(rr->raft, node, &req, &resp) != 0) {
         RedisModule_ReplyWithError(ctx, "ERR operation failed");
-        return REDISMODULE_OK;
+        goto out;
     }
 
     RedisModule_ReplyWithArray(ctx, 5);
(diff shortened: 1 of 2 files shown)


src/redisraft.c
```diff
@@ -1000,9 +1000,15 @@ static int cmdRaftSnapshot(RedisModuleCtx *ctx, RedisModuleString **argv, int ar
 
     rr->snapshotreq_received++;
 
+    /* raft_recv_snapshot() might trigger loading the received RDB file.
+     * While loading, Redis may process incoming messages to reply -LOADING.
+     * So, we must block the leader's connection to prevent processing further
+     * messages, as we haven't replied to the current command yet. */
+    RedisModuleBlockedClient *c = RedisModule_BlockClient(ctx, NULL, 0, 0, 0);
```
sjpotter, line 1007: I'm confused by the purpose of this.  I thought BlockClient was blocking the client (i.e. essentially allowing it to be backgrounded), not the server.

tezc: `rdbLoad()` is synchronous but inside that function, it still processes messages from the network. It hits here: https://github.com/redis/redis/blob/599e59ebc57283f52c60a8de56ec5f44d053109a/src/rdb.c#L2784

So, `rdbLoad()` is blocking from caller's point of view. Loading RDB is non-blocking from client's point of view. 
Clients get `-LOADING` reply.

In our case, stacktrace becomes something like:

```c
processCommandFromLeader()
raft_recv_snapshot();
rdbLoad();
startProcessingAnotherCommandFromLeader();
assert(false);
```

We need to prevent processing the next message of the leader node at this point to avoid that assert. Blocking the connection will prevent it as far as I can see.

tezc: @sjpotter Do you have any other comments? Otherwise, let's merge this as a workaround for now.

PR #451, opened 2022-07-14
Clang format

tezc:
Added clang-format to help us using a single style in the project. Actually, this is a very loose clang-format file, basically just a helper to catch simple issues. 

Later, I'll prepare a suggested style document for other things like 
- This clang-format file does not cover ex: Prefer 80 column limit, use `/* */` for comments etc. 
- Clang-format cannot cover: ex: how to name functions/variables etc.

Changes: 

- Added clang-format file under `src/`. Clang-format will check the files inside that directory.
- Moved third party code under `deps/common`: This files are `crc16.h crc16.c queue.h redismodule.h`. So, clang-format will not format these files.
- Moved C test files under `tests/unit`. There is another clang-format file there, slightly different. 
- CI will check formatting issues, sample output: https://github.com/RedisLabs/redisraft/runs/7337382717?check_suite_focus=true
- Your IDE should recognize clang-format. If you wan to run it locally, install `clang-format` and then:
   ```
   clang-format -n -Werror -style=file:src/.clang-format src/*
   ```

We cannot make everyhing as we wish, I marked some minor trade-offs as a review comment. Please play with clang-format file if you want, let me know your feedback.


src/redisraft.c
```diff
@@ -24,28 +25,28 @@ const char *redisraft_loglevels[] = {
...
-        LOG_LEVEL_DEBUG,
-        LOG_LEVEL_VERBOSE,
-        LOG_LEVEL_NOTICE,
-        LOG_LEVEL_WARNING,
-        LOG_LEVEL_COUNT
+    LOG_LEVEL_DEBUG,
+    LOG_LEVEL_VERBOSE,
+    LOG_LEVEL_NOTICE,
+    LOG_LEVEL_WARNING,
+    LOG_LEVEL_COUNT,
```
tezc, line 36: See the trailing comma. Without it, clang-format will not format this way. We need to add it when required.

src/connection.c
```diff
@@ -66,9 +66,9 @@ Connection *ConnCreate(RedisRaftCtx *rr, void *privdata,
 
     int timeout_millis = rr->config.connection_timeout;
 
-    conn->timeout = (struct timeval) {
-            .tv_sec = timeout_millis / 1000,
-            .tv_usec = (timeout_millis % 1000) * 1000
+    conn->timeout = (struct timeval){
+        .tv_sec = timeout_millis / 1000,
+        .tv_usec = (timeout_millis % 1000) * 1000,
```
tezc, line 71: See the trailing comma. Without it, clang-format will not format this way. We need to add it when required for arrays and designated struct initializers.

src/redisraft.h
```diff
@@ -308,92 +307,92 @@ void fsyncThreadAddTask(FsyncThread *th, int fd, raft_index_t requested_index);
...
-    int join_timeout;                   /* Milliseconds the node will continue to try joining a cluster */
-    int reconnect_interval;             /* Milliseconds to wait to reconnect to a node if connection drops */
-    int proxy_response_timeout;         /* Milliseconds to wait for a response to a proxy request */
-    int response_timeout;               /* Milliseconds to wait for a response to a Raft message */
-    long long append_req_max_count;     /* Max in-flight appendreq message count between two nodes. */
-    long long append_req_max_size;      /* Max appendreq message size in bytes. Just an approximation. */
-    long long snapshot_req_max_count;   /* Max in-flight snapshotreq message count between two nodes. */
-    long long snapshot_req_max_size;    /* Max snapshotreq message size in bytes. Just an approximation. */
-    long long scan_size;                /* how many keys to fetch at a time internally for raft.scan */
+    int periodic_interval;            /* raft_periodic() interval */
```
tezc, line 331: Consecutive comment lines are aligned but we cannot align with the upper section anymore. (There is a new line between them).

src/config.c
```diff
@@ -711,13 +710,15 @@ RRStatus ConfigInit(RedisRaftConfig *c, RedisModuleCtx *ctx)
...
-        .str_conf_ref = RedisModule_CreateString(NULL, "", 0)
+    *c = (RedisRaftConfig){
+        .str_conf_ref = RedisModule_CreateString(NULL, "", 0),
     };
 
     if (validateRedisConfig(ctx) != RR_OK) {
         goto err;
     }
+
+    /* clang-format off */
```
tezc, line 721: This is how you disable clang-format for a part: 
```c
/* clang-format off */
code you formatted yourself
/* clang-format on */ 
```

We should avoid using this. This is the only place right now.

PR #456, opened 2022-07-20
Fix use-after-free bug for shardgroup replace

tezc:
Fix use-after-free bug for shardgroup replace

It's unlikely to hit this path. Currently, only case is if leader receives a shardgroup replace command when leader transfer is in progress.


sjpotter (approved):
wouldn't so much call this as use after free (though that is accurate), but a brainfart. RaftReqFree should always be after the Reply.

PR #458, opened 2022-07-24
Commands spec dictionary flows clean up

fadidahanna:
Fixes: #457


src/redisraft.h
```diff
@@ -787,6 +787,7 @@ RedisModuleString **RaftRedisLockKeysDeserialize(const void *buf, size_t buf_siz
 
 /* raft.c */
 RRStatus RedisRaftInit(RedisRaftCtx *rr, RedisModuleCtx *ctx);
+void RedisRaftFree( RedisModuleCtx *ctx);
```
sjpotter, line 790: I wonder if this is really needed to capture things?

src/commands.c
```diff
@@ -236,59 +236,98 @@ static void freeCommandSpecDict(RedisModuleDict *dict)
...
+            continue;
+        }
+
+        cs = RedisModule_Calloc(1, sizeof(*cs));
+        cs->name = RedisModule_Strdup(tok);
+        cs->flags = CMD_SPEC_DONT_INTERCEPT;
+
+        int ret = RedisModule_DictSetC(command_spec_dict, tok, strlen(tok), cs);
+        if (ret != REDISMODULE_OK) {
+            RedisModule_Free(cs->name);
```
tezc, line 301: Previously, we were rebuilding a separate the map on each config change and then replacing it atomically if we are successful. So, it had a few advantages. 
- If user passes a bad config, we wouldn't have partial configs. With this PR, if user passes `ignored_commands`, say `cmd1, cmd2, cmd1, cmd3`, we are going to add `cmd1,cmd2` to the map and also return an error. (see the duplicate command). 
- Currently, we don't delete entries from the map. If `ignored` commands are `cmd1,cmd2` and new one is empty string, entries stay in the map. 
- Similar problem for module commands. If we unload a module, we will not delete its commands from the dictionary. Previously, that wasn't the case as we create a new dict each time. 

Overall, I'm not sure what is the benefit of different function calls for setting the ignored commands and other commands. I guess rebuilding the dict fully is simpler in our case. 

----
If this is the way we should do this (not sure), maybe we should have `CMD_SPEC_IGNORED` flag and mark user specified configs as `CMD_SPEC_IGNORED | CMD_SPEC_DONT_INTERCEPT`. 

Then, if user wants to update `ignored-commands`, we can create a temp dict, copy `non - CMD_SPEC_IGNORED` ones there, create new cmdspecs for the new `ignored-commands` and then atomically replace the map.

Similar for redis commands, on update, we copy `CMD_SPEC_IGNORED` ones to a new map and then fetch new redis commands to this map, an then atomically replace the map.

src/redisraft.c
```diff
@@ -1874,6 +1874,123 @@ static int registerRaftCommands(RedisModuleCtx *ctx)
...
+    }
+}
+
+void RedisRaftFreeGlobals()
+{
+    if (redisraft_log_ctx) {
+        RedisModule_FreeThreadSafeContext(redisraft_log_ctx);
+    }
+    RedisRaftCtxClear(&redis_raft);
+    CommandSpecFree();
```
tezc, line 1991: I think we should delete statics/globals from the codebase eventually. Static variables just cause confusion and hurt readability. We should define commandspec in the RedisRaftCtx, just like any other variable, maybe something like:

```c
typedef struct CommandSpecMap {
     RedisModuleDict *cmds;
} CommandSpecMap;

typedef struct RedisRaftCtx {
     ...
     CommandSpecMap cmd_map;
     ...
}

int CommandSpecInit(CommandSpecMap *map);

Comment (tezc) [reply src/redisraft.c:1991]: Probably, we should do the same thing for `redisraft_log_ctx`, move it into RedisRaftCtx. Then, we can have one function to free RedisRaftCtx: `RedisRaftCtxFree()` .

File: src/commands.c (hunk 2 above)
Comment (fadidahanna) [reply src/commands.c:301]: 1. regarding the "bad config" - actually even in the orig implementation we ignored duplication (see https://github.com/RedisLabs/redisraft/blob/72b4741a1fa30b28ce800749c77e59bf29c66984/src/commands.c#L263). The difference is in case of error (oom): here we add part of the cmds to the dict, instead releasing the whole dict, and I think we should keep this behaviour. 
2. Right, there is an assumption here that `ignored_commands` should be valid commands only, which is part of the command dict anyway. I agree we have several issues here: 
- Does empty "ignored-commands" means "removing the previous restrictions"? - if yes we should support it (it's during config time, so no performance concerns)
- Agree that for module commands we have no other way but to update the whole dict.

Considering the above issues, we should update the dict each time - will update my code accordingly.

File: src/redisraft.c (hunk 3 above)
Comment (fadidahanna) [reply src/redisraft.c:1991]: Agree with that, but wonder if `RedisRaftCtx` is the right container - from what I see `RedisRaftCtx` is mostly used to keep data for raft lib (except the migration data that were added lately). WDYT?

File: src/commands.c (hunk 2 above)
Comment (tezc) [reply src/commands.c:301]: 1- Just to be clear, in case of an error, we dismiss the temporary dict, so we don't store the new/bad config partially. 

2- 
> Does empty "ignored-commands" means "removing the previous restrictions"? - if yes we should support it (it's during config time, so no performance concerns)

Yes. 

Again just to be clear, what I meant is , with the current PR, we don't ever delete commanspecs from the dict. Let's say previous config is `cmd1,cmd2` and new config is `cmd3,cmd4`. With this PR, `cmd1,cmd2` will stay in the dict? This wasn't the case previously. 

----

Overall, I feel like we just need a single function rebuild the dict each time. (We already have it, `CommandSpecSet()` )
I guess we can rename  `CommandSpecSet()` as `CommandSpecRebuild()` for clarity, maybe. 

Because if `CommandSpecInit()`, `CommandSpecUpdateIgnoredCommands()` and `CommandSpecUpdateRedisCommands()` will do the same thing, rebuild the dict, we don't need three functions. At least, this was my conclusion. Maybe, I'm wrong. 

I know it is ugly currently as we call this in the config path but don't know if there is a better way to do it.

File: src/redisraft.c (hunk 3 above)
Comment (tezc) [reply src/redisraft.c:1991]: Yes, it is the right one. `RedisRaftCtx` is our root object, one object to rule others :) Ideally, we should just have this one as static(currently, it is) and reach others from it.

File: src/commands.c (hunk 2 above)
Comment (fadidahanna) [reply src/commands.c:301]: For sure we don't want to call it *only* in the config path, because `ignored-commands` is not a mandatory configuration, meaning command spec will not be created at all. We should have a single function, which we call on 3 different paths - onLoad, on config change and on module load state change.

Comment (fadidahanna) [reply src/commands.c:301]: > 1- Just to be clear, in case of an error, we dismiss the temporary dict, so we don't store the new/bad config partially.

Right, but in the current implementation duplication is not an error. 

Anyway, it seems like we treat differently to a `RedisModule_DictSetC()` error - in some cases we assert on that (see https://github.com/RedisLabs/redisraft/blob/72b4741a1fa30b28ce800749c77e59bf29c66984/src/commands.c#L128) on other cases, like in `CommandSpecSet` (https://github.com/RedisLabs/redisraft/blob/72b4741a1fa30b28ce800749c77e59bf29c66984/src/commands.c#L248), we return an error, and ignoring... we should be consistent...

Comment (tezc) [reply src/commands.c:301]: > For sure we don't want to call it only in the config path, because ignored-commands is not a mandatory configuration, meaning command spec will not be created at all

No, it will be called for the default value even if the config is missing. 

> Right, but in the current implementation duplication is not an error.

Ahh, I see, sorry. Yeah, duplication is not an error, just image another error case instead :) 

> Anyway, it seems like we treat differently to a RedisModule_DictSetC() error - in some cases we assert on that 

Depends if we can handle the error not I guess.

File: src/cluster.c [hunk 4]
```diff
@@ -1083,24 +1083,17 @@ ShardGroup **ShardGroupsParse(RedisModuleCtx *ctx,
  * ShardGroup.
  */
 
-void ShardingInfoInit(RedisRaftCtx *rr)
+void ShardingInfoInit(RedisModuleCtx *ctx, ShardingInfo **si)
```
sjpotter, line 1086: so I'm not quite sure why the prototype changed, it still only taking things from the RedisRaftCtx

sjpotter: ok, now I think understand better (needs a better description to go over all that changes here).  I think is about simply making it separation of  concerns.  while, yes, it could just pass a single rr pointer, passing the RedisModuleCtx and a double pointer to the pointer we would alloc and handle is a "cleaner" api.

sjpotter:
seems fine, I'm wondering if we made a mistake in retrospect by combining two things.  changing the command spec dict and how memory is init'd/freed in general.  made it harder to review than it probably should have been.

sjpotter (approved):

PR #464, opened 2022-07-26
poc max memory support with eval / evalsha

sjpotter:
depends on https://github.com/sjpotter/redis/pull/3

### DIFF
--- a/src/commands.c
+++ b/src/commands.c
@@ -27,6 +27,10 @@ static const CommandSpec commands[] = {
     {"save",                        CMD_SPEC_UNSUPPORTED                     },
     {"bgsave",                      CMD_SPEC_UNSUPPORTED                     },
 
+ /* Commands that need further processing for DENY_OOM */
+    {"eval",                        CMD_SPEC_FLAGS                           },
+    {"evalsha",                     CMD_SPEC_FLAGS                           },
+
  /* Blocking commands not supported */
     {"brpop",                       CMD_SPEC_UNSUPPORTED                     },
     {"brpoplpush",                  CMD_SPEC_UNSUPPORTED                     },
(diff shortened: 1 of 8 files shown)


src/commands.c
```diff
@@ -307,13 +311,30 @@ const CommandSpec *CommandSpecGet(const RedisModuleString *cmd)
...
+                if (!strncmp(cs->name, "eval", strlen("eval")) && array->commands[i]->argc > 1) {
+                    uint64_t sub_flags;
+                    if (RedisModule_GetScriptBodyFlags(ctx, array->commands[i]->argv[1], &sub_flags) == REDISMODULE_OK) {
+                        if (!(sub_flags && (REDISMODULE_SCRIPT_FLAG_NO_WRITES | REDISMODULE_SCRIPT_FLAG_ALLOW_OOM))) {
+                            flags |= CMD_SPEC_DENYOOM;
+                        }
+                    }
+                } else if (!strncmp(cs->name, "evalsha", strlen("evalsha")) && array->commands[i]->argc > 1) {
+                    uint64_t sub_flags;
+                    if (RedisModule_GetScriptSHAFlags(ctx, array->commands[i]->argv[1], &sub_flags) == REDISMODULE_OK) {
```
tezc, line 331: It looks good to me. Just have two questions:
- Do we need to pass `ctx` to these new functions? Maybe, we won't ever need it. We can discuss with others.
- As you mentioned before, do functions require any special treatment?

sjpotter: ok, realized we probably don't need context (was from error message returing, but perhaps not needed anymore, we would just return 0 for flags in case of error and later code would handle the error).

functions would need their own Get*Flags functions which haven't been written, but would work similarly I'd expect.

src/commands.c
```diff
@@ -303,19 +307,41 @@ const CommandSpec *CommandSpecGet(const RedisModuleString *cmd)
...
+
+    return !(sub_flags & (REDISMODULE_SCRIPT_FLAG_NO_WRITES | REDISMODULE_SCRIPT_FLAG_ALLOW_OOM));
+}
+
 /* For a given RaftRedisCommandArray, return a flags value that represents
  * the aggregate flags of all commands. If a command is not listed in the
  * command table, use default_flags.
  */
-unsigned int CommandSpecGetAggregateFlags(RaftRedisCommandArray *array, unsigned int default_flags)
+unsigned int CommandSpecGetAggregateFlags(RedisModuleCtx *ctx, RaftRedisCommandArray *array, unsigned int default_flags)
```
fadidahanna, line 331: ctx can be removed

tests/integration/test_sanity.py
```diff
@@ -522,3 +522,25 @@ def test_maxmemory(cluster):
...
+
+def test_max_memory_eval(cluster):
+    cluster.create(3)
+
+    val = ''.join('1' for _ in range(2000000))
+
+    cluster.execute('set', 'key1', val)
+    cluster.execute('config', 'set', 'maxmemory', 100000)
+
+    with (raises(ResponseError, match="OOM command not allowed when used memory > 'maxmemory'")):
```
fadidahanna, line 535: reply for scripts should be `-OOM allow-oom flag is not set on the script, can not run it when used memory > 'maxmemory'`, see OSS code

tests/integration/test_sanity.py
```diff
@@ -522,3 +522,25 @@ def test_maxmemory(cluster):
...
+        cluster.execute("eval", """
+            redis.call('SET','key1','value1');
+            redis.call('SET','key2','value2');
+            redis.call('SET','key3','value3');
+            return 1234;""", "0")
+
+    assert cluster.execute("eval", """#!lua flags=allow-oom
```
fadidahanna, line 542: it's worth to test also `no-writes` for coverage

tests/integration/test_sanity.py
fadidahanna: k, seems like we have two paths in redis that check OOM for scripts - can you please verify it? see https://github.com/redis/redis/blob/00097bf4aa9e28477ee5d9361569f3212fb7493a/tests/unit/moduleapi/misc.tcl#L200

fadidahanna: suggest talking with Oran about that, he made some changes lately in this area

sjpotter: ok, can probably do that with our new FLAGS flag, as that would tell us which error message to return

tests/integration/test_sanity.py
sjpotter: ok

tests/integration/test_multi.py
```diff
@@ -189,3 +189,4 @@ def test_multi_with_maxmemory(cluster):
     assert node.execute('MULTI') == b'OK'
     assert node.execute('GET', 'key1') == b'QUEUED'
     assert node.execute('EXEC') == [val.encode()]
+
```
fadidahanna, line 192: please remove

sjpotter:
rebased wth @fadidahanna changes merged in master

fadidahanna:
@sjpotter can we close this poc PR?  thanks!

PR #466, opened 2022-07-27
[RR-22] ACL support

sjpotter:
* check ACLs at append time to see if they are valid using dry run
* if a "script" (i.e. need apply time acl checking), 
  * at append time 
    * extract current client's ACL and include in log entry
  * at apply time
    * create a ModuleUser at apply time if a user doesn't already exist for this ACL
    * assign existing/new user to ctx with RM_SetContextUser()
    * execute RM_Call with user enforcement.


src/commands.c
```diff
@@ -27,6 +27,9 @@ static const CommandSpec commands[] = {
     {"save",                        CMD_SPEC_UNSUPPORTED                     },
     {"bgsave",                      CMD_SPEC_UNSUPPORTED                     },
 
+    {"eval",                        CMD_SPEC_SEND_ACL                        },
```
fadidahanna, line 30: This should be done also for `eval_ro` etc.

src/raft.c
```diff
@@ -346,8 +366,13 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
         }
 
         enterRedisModuleCall();
-        reply = RedisModule_Call(ctx, cmd,
-                                 rr->resp_call_fmt, &c->argv[1], c->argc - 1);
+        if (user) {
+            reply = RedisModule_CallWithUser(ctx, user, cmd,
```
tezc, line 370: Shoul we allow null user in this function? It'd effectively be `RedisModule_Call()` when user is null.

src/redisraft.c
```diff
@@ -670,8 +670,20 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
         return;
     }
 
+    if (cmd_flags & CMD_SPEC_SEND_ACL) {
+        RedisModuleString *user_name = RedisModule_GetCurrentUserName(ctx);
+        RedisModuleUser *user = RedisModule_GetModuleUserFromUserName(user_name);
+        RedisModule_FreeString(ctx, user_name);
+        cmds->acl = RedisModule_GetModuleUserACLString(ctx, user);
+        RedisModule_FreeModuleUser(user);
```
tezc, line 678: Probably, we should consider something like `RedisModule_GetCurrentUserAcl()` just to avoid these calls and allocations.

src/redisraft.h
```diff
@@ -682,6 +684,8 @@ typedef struct {
 #define CMD_SPEC_SORT_REPLY     (1 << 5) /* Command output should be sorted within a lua script */
 #define CMD_SPEC_RANDOM         (1 << 6) /* Commands that are always random */
 #define CMD_SPEC_DENYOOM        (1 << 7) /* Commands that can't be run when in oom */
+#define CMD_SPEC_FLAGS          (1 << 8) /* Commands that have script/function flags */
+#define CMD_SPEC_SEND_ACL       (1 << 9) /* Commands that have require we serialize ACL */
```
tezc, line 688: I'd prefer calling this something like `CMD_SPEC_SCRIPT`. Sending ACL is what we will do to implement  raft logic correctly. I see cmdspec flags as the type of a command rather than the action we need to take.

src/raft.c
sjpotter: hmm.  reasonable, makes code simpler.  I'd agree.

src/commands.c
sjpotter: agreed

src/redisraft.c
sjpotter: ok.  change to redis, easy enough.

src/redisraft.h
sjpotter: So, I sort of agree, but was wondering if it was correct.  it currently is, but if in the future we need some apply time acl checking, it wouldn't be.

tezc: Apply time ACL checks for other commands? I think we can handle it by then if it becomes a problem. We can rename the flag or add another flag for those new commands. We don't have to maintain that behavior (deciding ACL serialization) under single flag anyway. 

So, until then, I feel like `CMD_SPEC_SCRIPT` is easier understand. Otherwise, whenever we see that flag in the commands.c, we'll have to jump to redisraft.c to see what does this flag actually mean.

src/redisraft.c
```diff
@@ -651,6 +651,11 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
         return;
     }
 
+    if (!(cmd_flags & CMD_SPEC_SCRIPTS) && validateAllCommandsACL(ctx, cmds) != REDISMODULE_OK) {
+        RedisModule_ReplyWithError(ctx, "NOPERM this user doesn't have proper permissions");
```
fadidahanna, line 655: `RM_ACLCheckCommandPermissions` returns only OK/ERROR, which doesn't reflect the real root cause of the error, so we cannot reply in the same reply msg as redis. We maybe can consider changing the api to return also the acl return val - `ACL_DENIED_CMD\ACL_DENIED_KEY\ACL_DENIED_CHANNEL`, it should be easy because `ACLCheckAllUserCommandPerm()`  already returns this value. @sjpotter , WDYT?

fadidahanna:
Do we want to enable `Script ACL check` redis test?

src/commands.c
```diff
@@ -27,6 +27,11 @@ static const CommandSpec commands[] = {
     {"save",                        CMD_SPEC_UNSUPPORTED                     },
     {"bgsave",                      CMD_SPEC_UNSUPPORTED                     },
 
+    {"eval",                        CMD_SPEC_SCRIPTS,                        },
+    {"evalsha",                     CMD_SPEC_SCRIPTS,                        },
+    {"eval_ro",                     CMD_SPEC_SCRIPTS,                        },
+    {"evalsha_ro",                  CMD_SPEC_SCRIPTS,                        },
+
```
fadidahanna, line 34:
```suggestion
    {"eval",                        CMD_SPEC_SCRIPTS                         },
    {"evalsha",                     CMD_SPEC_SCRIPTS                         },
    {"eval_ro",                     CMD_SPEC_SCRIPTS                         },
    {"evalsha_ro",                  CMD_SPEC_SCRIPTS                         },

```

src/redisraft.c
sjpotter: the better response would be to either return the key (as the underlying call does) along with the error or just pass in a RedisModuleString ** and if not null, it generates the proper error message.  I thought that was a bit overkill for now.  In redis itself, it had a default error message when it can't parse.

sjpotter:
rebased wth @fadidahanna changes merged in master

sjpotter:
this has been updated to use latest redis and the small fixes needed to get it to work based on oran's changes.

src/multi.c
```diff
@@ -102,7 +102,10 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
          */
         unsigned int cmd_flags = CommandSpecTableGetAggregateFlags(rr->commands_spec_table, cmds, 0);
 
-        if (cmd_flags & CMD_SPEC_UNSUPPORTED) {
+        if (validateCommandACL(ctx, cmds->commands[0]) == REDISMODULE_ERR) {
```
tezc, line 105: Why are we manually validating here? Dry-run will validate commands in the array anyway, no?

src/redisraft.c
```diff
@@ -649,6 +649,10 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
         RaftRedisCommand *cmd = cmds->commands[i];
         size_t cmdlen;
         const char *cmdstr = RedisModule_StringPtrLen(cmd->argv[0], &cmdlen);
+        /* skip multi */
+        if (i == 0 && cmdlen == 5 && !strncasecmp(cmdstr, "MULTI", 5)) {
+            continue;
+        }
```
tezc, line 655: Why do we skip MULTI? I remember we discussed this and decide we shouldn't skip it, no? Because dry-run can also reject MULTI.

src/common.c
```diff
@@ -305,3 +305,25 @@ void raftNodeIdToString(char *output, const char *dbid, raft_node_id_t raft_id)
...
+int validateAllCommandsACL(RedisModuleCtx *ctx, RaftRedisCommandArray *cmds)
+{
+    for (int i = 0; i < cmds->len; i++) {
+        if (validateCommandACL(ctx, cmds->commands[i]) == REDISMODULE_ERR) {
+            return REDISMODULE_ERR;
+        }
+    }
+
+    return REDISMODULE_OK;
+}
```
tezc, line 329:
```suggestion
}

```

src/common.c
```diff
@@ -305,3 +305,25 @@ void raftNodeIdToString(char *output, const char *dbid, raft_node_id_t raft_id)
...
+    RedisModuleString *user_name = RedisModule_GetCurrentUserName(ctx);
+    RedisModuleUser *user = RedisModule_GetModuleUserFromUserName(user_name);
+    int ret = RedisModule_ACLCheckCommandPermissions(user, cmd->argv, cmd->argc);
+    RedisModule_FreeModuleUser(user);
+    RedisModule_FreeString(ctx, user_name);
+
+    return ret;
+}
+
+int validateAllCommandsACL(RedisModuleCtx *ctx, RaftRedisCommandArray *cmds)
```
tezc, line 320: Seems like this functions is unused

src/raft.c
```diff
@@ -306,6 +306,26 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
...
+            char user_name[64];
+            uint64_t count = RedisModule_DictSize(rr->acl_dict);
+            count++;
+            snprintf(user_name, 64, "redis_raft%lu", count);
+            user = RedisModule_CreateModuleUser(user_name);
+            RedisModule_Assert(user != NULL);
+
+            size_t acl_len;
+            const char *acl_str = RedisModule_StringPtrLen(cmds->acl, &acl_len);
+            RedisModule_Assert(RedisModule_SetModuleUserACLString(ctx, user, acl_str, NULL) == REDISMODULE_OK);
```
tezc, line 324:
```suggestion
            int ret = RedisModule_SetModuleUserACLString(ctx, user, acl_str, NULL);
            RedisModule_Assert(ret == REDISMODULE_OK);
```

Function call in an assert statement feels like something not important. Even though `RedisModule_Assert()` is a runtime check, traditionally, `assert()` is stripped out in the release build. So, I personally tend to ignore statements in an assert while reading the code. Let's have it in another line to help me while reading the code :)

src/serialization.c
```diff
@@ -218,6 +227,23 @@ RRStatus RaftRedisCommandArrayDeserialize(RaftRedisCommandArray *target, const v
...
     target->asking = asking;
 
+    /* Read ACL */
+    RedisModuleString *acl;
+    if ((n = decodeString(p, buf_size, &acl)) < 0) {
+        return RR_ERROR;
+    }
+    p += n;
+    buf_size -= n;
+    {
```
tezc, line 237: Scope is unnecessary here I guess.

src/multi.c
sjpotter: I believe (might be wrong) that multi evaluates at "append" to queue time as well, but perhaps I confused myself on that.  will check again

src/raft.c
```diff
@@ -341,23 +361,43 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
...
+                resp_call_fmt = "0v";
+            }
+        } else {
+            if (cmds->acl) {
+                /* We don't have a client so only check ACL (C) */
+                resp_call_fmt = "Cv";
+            } else {
+                resp_call_fmt = "v";
+            }
+        }
```
tezc, line 385: 1- As we don't support older Redis versions, I think we can delete this: https://github.com/RedisLabs/redisraft/blob/722e48ec70fddde8fe3584d5e8ab6862dcdf8e6d/src/redisraft.c#L1931

2- Why don't we go with `CE0v` here all the time?

src/raft.c
```diff
@@ -341,23 +361,43 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
...
         if (reply_ctx) {
-            if (reply) {
-                RedisModule_ReplyWithCallReply(reply_ctx, reply);
+            int type = RedisModule_CallReplyType(reply);
+            if (type != REDISMODULE_REPLY_ERROR) {
+                RedisModule_ReplyWithCallReply(ctx, reply);
             } else {
-                replyRMCallError(reply_ctx, ret_errno, cmd, cmdlen);
+                const char *str = RedisModule_CallReplyStringPtr(reply, NULL);
+                RedisModule_ReplyWithError(ctx, str);
```
tezc, line 400: Why do we need these if checks? If we go with `CE0v` flags, can we just do:

```c


if (reply_ctx) {
    RedisModule_ReplyWithCallReply(reply_ctx, reply);
}

```

src/raft.c
```diff
@@ -306,6 +306,26 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
...
+        if (nokey) {
+            char user_name[64];
+            uint64_t count = RedisModule_DictSize(rr->acl_dict);
+            count++;
+            snprintf(user_name, 64, "redis_raft%lu", count);
+            user = RedisModule_CreateModuleUser(user_name);
+            RedisModule_Assert(user != NULL);
+
+            size_t acl_len;
+            const char *acl_str = RedisModule_StringPtrLen(cmds->acl, &acl_len);
```
fadidahanna, line 323:
```suggestion
            const char *acl_str = RedisModule_StringPtrLen(cmds->acl, NULL);
```

we can ignore acl size

src/multi.c
sjpotter: chekecd the code, and it validates acl twice, at enqueue time and at "execution" time (in case ACL changed per comment in code)

src/redisraft.c
sjpotter: in regular redis multi isn't enqueued, hence not skipping it would cause a problem as a change in multi acl shouldn't impact anything now.

src/raft.c
```diff
@@ -341,23 +361,43 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
         int old_entered_eval = rr->entered_eval;
 
         if ((cmdlen == 4 && !strncasecmp(cmd, "eval", 4)) ||
-            (cmdlen == 7 && !strncasecmp(cmd, "evalsha", 7))) {
+            (cmdlen == 7 && !strncasecmp(cmd, "evalsha", 7)) ||
+            (cmdlen == 5 && !strncasecmp(cmd, "fcall", 5))) {
```
fadidahanna, line 365: what about `X_ro` cmds?

src/raft.c
fadidahanna: calling `RM_call` with `C` flag when we have no ctx user will try to use client user, which is not what we want here.

src/multi.c
tezc: You're right and I'm not happy with that :laughing:

It is just weird. I wonder if there is any use-case for this. User won't see the ACL error until EXEC in any case. So, the only difference between "ACL check twice" and   "ACL check on EXEC" is when there is an ACL change while accumulating.

This scenario doesn't make any sense to me. Maybe I'm missing something but I personally don't think it is worth the overhead. In RR, now we have extra code for it (`validateCommandACL()`) and that function also adds performance overhead unnecessarily (due to allocations, lookup etc. Redis doesn't have any of these overheads). 

Let's keep it this way for now but discuss it tomorrow. It'd be great if we can get away without doing it twice.

src/raft.c
tezc: Hmm, can we at least do:

```c
char *resp_call_fmt = cmds->acl ? "CE0v" : "E0v";
```

src/redisraft.c
tezc: You are right again and I'm still not happy :) 

Btw we should have one flag variable in the entry to encode these. `multi`, `asking` etc can be a bit of a single variable. If we have it, our entry size will be smaller and we can avoid this check here and also on apply time. We can do it later ofcourse as part of another PR.

tests/integration/test_multi.py
```diff
@@ -189,3 +189,41 @@ def test_multi_with_maxmemory(cluster):
     assert node.execute('MULTI') == b'OK'
     assert node.execute('GET', 'key1') == b'QUEUED'
     assert node.execute('EXEC') == [val.encode()]
+
```
tezc, line 192: Should we have "multi with scripts" tests? As far as I can see it should work according to the current code but let's have a test for it.

src/common.c
sjpotter: ok, removed

src/serialization.c
sjpotter: ok, removed

src/raft.c
sjpotter: pretty sure I did get rid of it, don't see it here anymore.  confused.

sjpotter: ok, tried to simplify it a bit with an explanation, also caught an issue I think

src/common.c
```diff
@@ -305,3 +305,25 @@ void raftNodeIdToString(char *output, const char *dbid, raft_node_id_t raft_id)
...
+{
+    for (int i = 0; i < cmds->len; i++) {
+        if (validateCommandACL(ctx, cmds->commands[i]) == REDISMODULE_ERR) {
```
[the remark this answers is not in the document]
sjpotter: removed as not used, so not needed anymore

src/raft.c
sjpotter: removed, as unused

src/raft.c
tezc: Really? It should not crash, sounds like a bug in Redis. I thought the whole purpose of  `E` flag is being able to reply to the client without doing any additional checks.

Can you check it again with a clean build? I just make this change and it doesn't crash for me:

```c 
reply = RedisModule_Call(ctx, "some_non_existing_command", "E0v", &c->argv[1], c->argc - 1);
RedisModule_ReplyWithCallReply(reply_ctx, reply);
```

sjpotter: I ran into the problem with my dry run tests and the module, I'll try again.  If the above is working, perhaps was a temporary condition that was fixed (and yes, I agree, I'd expect it to just work).

src/raft.c
sjpotter: ok, yea, this is a bit old, I'm wondering if we it be worth to do something like

```
        if (((cmdlen == 4 || cmdlen == 7 || cmdlen == 10) && !strncasecmp(cmd, "eval", 4)) ||
            ((cmdlen == 5 || cmdlen == 8) && !strncasecmp(cmd, "fcall", 5))) {
            rr->entered_eval = 1;
        }
```

i.e. 4 is eval, 7 is evalsha and eval_ro and 10 is evalsha_ro and the same for fcall.

sjpotter: not neccessary

src/common.c
```diff
@@ -305,3 +305,25 @@ void raftNodeIdToString(char *output, const char *dbid, raft_node_id_t raft_id)
 {
     snprintf(output, RAFT_SHARDGROUP_NODEID_LEN + 1, "%.32s%08x", dbid, raft_id);
 }
+
+int validateCommandACL(RedisModuleCtx *ctx, RaftRedisCommand *cmd)
+{
+    RedisModuleString *user_name = RedisModule_GetCurrentUserName(ctx);
+    RedisModuleUser *user = RedisModule_GetModuleUserFromUserName(user_name);
+    int ret = RedisModule_ACLCheckCommandPermissions(user, cmd->argv, cmd->argc);
```
fadidahanna, line 313: Can we use dry run mode here?

tests/integration/test_sanity.py
```diff
@@ -520,10 +520,68 @@ def test_maxmemory(cluster):
...
+    cluster.execute('set', 'abc', 1)
+    cluster.execute('acl', 'setuser', 'default', 'resetkeys', '(+set', '~key*)', '(+get', '~key*)')
+    cluster.execute('get', 'key')
+    with (raises(ResponseError, match="acl verification failed, can't access at least one of the keys mentioned in the command arguments")):
+        cluster.execute('get', 'abc')
+
+    cluster.execute('set', 'key', 2)
+    commit_idx = cluster.leader_node().commit_index()
+    assert cluster.node(1).raft_debug_exec('get', 'key') == b'2'
+    cluster.node(2).wait_for_commit_index(commit_idx, gt_ok=True)
```
tezc, line 540: Maybe I'm missing something but can we replace `wait_xx` calls with a single `cluster.wait_for_unanimity()`?

src/common.c
sjpotter: modified multi.c to use dry run mode (though only for ACL, not for OOM, oom is different in multi, i.e. even non deny oom commands are rejected when OOM (as dont want multi queue to grow without bounds)

src/raft.c
sjpotter: ok, made it like this, seems to be passing now, unsure why I had issues before.

src/serialization.c
fadidahanna: The comment is still relevant...

src/common.c
fadidahanna: The comment is still relevant

tests/integration/test_sanity.py
sjpotter: ok, done

src/raft.c
```diff
@@ -306,6 +306,27 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
...
+        if (nokey) {
+            char user_name[64];
+            uint64_t count = RedisModule_DictSize(rr->acl_dict);
+            count++;
+            snprintf(user_name, 64, "redis_raft%lu", count);
+            user = RedisModule_CreateModuleUser(user_name);
+            RedisModule_Assert(user != NULL);
+
+            size_t acl_len;
+            const char *acl_str = RedisModule_StringPtrLen(cmds->acl, &acl_len);
```
tezc, line 323:
```suggestion
            const char *acl_str = RedisModule_StringPtrLen(cmds->acl, NULL);
```
One less line :)

src/raft.c
```diff
@@ -306,6 +306,27 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
...
+            int ret = RedisModule_SetModuleUserACLString(ctx, user, acl_str, NULL);
+            RedisModule_Assert(ret == REDISMODULE_OK);
+
+            RedisModule_DictSet(rr->acl_dict, cmds->acl, user);
```
tezc, line 327: So, dict might grow over time, we might be okay for now(don't know if people play with ACLs a lot) but it'd be nice to have an issue to remember it.

sjpotter: https://redislabs.atlassian.net/browse/RR-93

src/raft.c
```diff
@@ -341,23 +362,39 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
         int old_entered_eval = rr->entered_eval;
 
         if ((cmdlen == 4 && !strncasecmp(cmd, "eval", 4)) ||
-            (cmdlen == 7 && !strncasecmp(cmd, "evalsha", 7))) {
+            (cmdlen == 7 && !strncasecmp(cmd, "evalsha", 7)) ||
+            (cmdlen == 5 && !strncasecmp(cmd, "fcall", 5))) {
             rr->entered_eval = 1;
         }
 
+        char *resp_call_fmt;
```
tezc, line 369:
```c
        char *resp_call_fmt = cmds->acl ? "CE0v" : "E0v";

        enterRedisModuleCall();
        RedisModule_SetContextUser(ctx, user);
        exitRedisModuleCall();

        rr->entered_eval = old_entered_eval;

        if (reply_ctx) {
            RedisModule_ReplyWithCallReply(reply_ctx, reply);
        }

	if (reply) {
        }
```
Starting from this line and to the end, can we do this? 

Note: I think crashes you see are because you reply to the `ctx`, not to `reply_ctx`. `reply_ctx` occasionally will be NULL.

sjpotter: less allocations on followers.  so "need" is relative.

tezc: Yes I see but it only helps on error cases (I assume error cases are rare). 

My only concern is readability. Flags are already hard to read and explain :) so, at least we can skip this extra check as it is not going to help us anyway.

tezc (approved):

src/raft.c
sjpotter: I still don't understand why redis went with a "flags" as characters method vs binary bits with #defines.  I get it for denoting the arguments, but not for the non arguments.

tezc: Yes, it is a bit counter-intuitive.

tezc: Yes, format string gets uglier this way when you start adding new flags.

tezc: Yes, format string gets uglier this way when you start adding new flags.

src/raft.c
```diff
@@ -341,24 +361,25 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
...
         if (reply_ctx) {
-            if (reply) {
-                RedisModule_ReplyWithCallReply(reply_ctx, reply);
-            } else {
-                replyRMCallError(reply_ctx, ret_errno, cmd, cmdlen);
-            }
+            RedisModule_ReplyWithCallReply(ctx, reply);
         }
 
         if (reply) {
```
fadidahanna, line 385:
```suggestion
```
As @tezc pointed out - we call RM_Call with "E" to reply cannot be NULL.

src/redisraft.c
```diff
@@ -649,13 +649,16 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
...
 
         enterRedisModuleCall();
         RedisModuleCallReply *reply = RedisModule_Call(ctx, cmdstr, "DCEMv", cmd->argv + 1, cmd->argc - 1);
```
fadidahanna, line 658: Redis OSS documentation for Dry-Run mode: 
``` *     * 'D' -- A "Dry Run" mode. Return before executing the underlying call().
 *              If everything succeeded, it will return with a NULL, otherwise it will
 *              return with a CallReply object denoting the error, as if it was called with
 *              the 'E' code.
 ```

But it seems we will return an error reply only when calling with `E` flag. Need to change the documentation there.

File: src/multi.c [hunk 26]
```diff
@@ -105,17 +105,37 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
...
+            /* Can only enqueue if not in an OOM situation, even for non deny oom commands */
             RedisModule_ReplyWithError(ctx, "OOM command not allowed when used memory > 'maxmemory'.");
             multiState->error = true;
-        } else {
-            RaftRedisCommandArrayMove(&multiState->cmds, cmds);
-            RedisModule_ReplyWithSimpleString(ctx, "QUEUED");
+            return true;
         }
 
+        /* "Multi Dry Run" - only check for ACL, not for OOM, as OOM is checked above */
```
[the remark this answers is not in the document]
sjpotter: its alreday handled above.  In multi even "get" is denied.  So we check ratio earlier.  so "Dry run" doesn't help us for OOM

src/raft.c
sjpotter: pretty sure this code was already modified?

src/redisraft.c
sjpotter: cofused

src/multi.c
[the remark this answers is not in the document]
fadidahanna: 👍

src/raft.c
sjpotter: ok, I see, because its consistent now between leader and follower.

src/raft.c
```diff
@@ -341,24 +361,25 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
...
 
         rr->entered_eval = old_entered_eval;
 
         if (reply_ctx) {
-            if (reply) {
-                RedisModule_ReplyWithCallReply(reply_ctx, reply);
-            } else {
-                replyRMCallError(reply_ctx, ret_errno, cmd, cmdlen);
-            }
+            RedisModule_ReplyWithCallReply(ctx, reply);
```
fadidahanna, line 382:
```suggestion
            RedisModule_ReplyWithCallReply(reply_ctx, reply);
```
Although in the current implementation having `reply_ctx` means `ctx == reply_ctx` I think we should use  `reply_ctx` for readability and future uses.

src/raft.c
fadidahanna: explanation, please? do we have a scenario when `reply == NULL`?

sjpotter: previously we executed RM_Call 2 ways at apply time, with and without E (only reply_ctx node got E).  Ozan argued for it to be the same, so it was simplified.  In the non E (follower/log replay) case, that would have meant errors (say acl) would be reply == NULL (and just ignored as on follower and didnt have an impact).

but I agree now, it can be simplified further

fadidahanna (approved):

tezc:
@sjpotter There are some test failures: https://github.com/RedisLabs/redisraft/actions/runs/3170634817

sjpotter:
the macos are build errors because of this

```
/Users/runner/work/redisraft/redisraft/src/raft.c:318:54: error: format specifies type 'unsigned long' but the argument has type 'uint64_t' (aka 'unsigned long long') [-Werror,-Wformat]
```

unsure how to fix it, as %llu will cause it to fail to build on linux.  (are these 32 bit mac builds?)

can't figure out the test undefined sanitizer error that seems to blow up most tests

sjpotter:
think I figured out the undefined error

DEBUG    sandbox:sandbox.py:54 c0/n1/stderr: /home/spotter/CLionProjects/redisraft/src/serialization_utils.c:160:5: runtime error: null pointer passed as argument 2, which is declared to never be null

tezc:
`uint64_t` is `unsigned long` on Linux. `unsigned long long` on MacOS. 
This is C,  people can define things as they want on their own operation system :) 

I guess we can do 

```c
snprintf(user_name, 64, "redis_raft%lu", (unsigned long) count);
```

tezc:
> DEBUG sandbox:sandbox.py:54 c0/n1/stderr: /home/spotter/CLionProjects/redisraft/src/serialization_utils.c:160:5: runtime error: null pointer passed as argument 2, which is declared to never be null

Can we use empty string instead of null for ACL string? Does it make things worse? Maybe we can avoid null checks in serialization.c ?

sjpotter:
this is the fix for encodeString

```
diff --git a/src/serialization_utils.c b/src/serialization_utils.c
index df75257..a575e96 100644
--- a/src/serialization_utils.c
+++ b/src/serialization_utils.c
@@ -142,6 +142,7 @@ int encodeString(char *p, size_t sz, RedisModuleString *str)
 
     if (str == NULL) {
         len = 0;
+        e = "";
     } else {
         e = RedisModule_StringPtrLen(str, &len);
     }
```

which I think is correct, as the == NULL portion should set both len and e, as the != portion does.

Issue #468, opened 2022-08-03
Redisraft crashes when getting interface addresses of VMs


doflink:
I'm deploying using VMs. However it crashes when loading the *redisraft.so* module.
The following the log.

```
[root@host-192-168-3-100 redisraft]# redis-server --loadmodule ./redisraft.so   --loglevel verbose
1355442:C 03 Aug 2022 17:39:32.542 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
1355442:C 03 Aug 2022 17:39:32.542 # Redis version=7.0.4, bits=64, commit=4ce3fd51, modified=1, pid=1355442, just started
1355442:C 03 Aug 2022 17:39:32.542 # Configuration loaded
1355442:M 03 Aug 2022 17:39:32.542 * monotonic clock: POSIX clock_gettime
                _._
           _.-``__ ''-._
      _.-``    `.  `_.  ''-._           Redis 7.0.4 (4ce3fd51/1) 64 bit
  .-`` .-```.  ```\/    _.,_ ''-._
 (    '      ,       .-`  | `,    )     Running in standalone mode
 |`-._`-...-` __...-.``-._|'` _.-'|     Port: 6379
 |    `-._   `._    /     _.-'    |     PID: 1355442
  `-._    `-._  `-./  _.-'    _.-'
 |`-._`-._    `-.__.-'    _.-'_.-'|
 |    `-._`-._        _.-'_.-'    |           https://redis.io
  `-._    `-._`-.__.-'_.-'    _.-'
 |`-._`-._    `-.__.-'    _.-'_.-'|
 |    `-._`-._        _.-'_.-'    |
  `-._    `-._`-.__.-'_.-'    _.-'
      `-._    `-.__.-'    _.-'
          `-._        _.-'
              `-.__.-'

1355442:M 03 Aug 2022 17:39:32.543 # Server initialized
1355442:M 03 Aug 2022 17:39:32.543 * <raft> RedisRaft version 255.255.255 [cd281252]


=== REDIS BUG REPORT START: Cut & paste starting from here ===
1355442:M 03 Aug 2022 17:39:32.548 # Redis 7.0.4 crashed by signal: 11, si_code: 1
1355442:M 03 Aug 2022 17:39:32.548 # Accessing address: (nil)
1355442:M 03 Aug 2022 17:39:32.548 # Crashed running the instruction at: 0x7f7c58c4eea6

------ STACK TRACE ------
EIP:
[...]

PR #469, opened 2022-08-03
Fix accessing nil address bug when getting interface addresses (#468)

doflink:
This merge request fixes the accessing null address bug when getting interface addresses insides a VM

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -639,8 +639,9 @@ static RRStatus getInterfaceAddr(NodeAddr *addr)
     }
 
     for (ent = addrs; ent != NULL; ent = ent->ifa_next) {
-        if (ent->ifa_addr == NULL)
+        if (ent->ifa_addr == NULL) {
             continue;
+        }    
 
         /* Skip loopback and non-IP interfaces */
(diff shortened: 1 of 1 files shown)


fadidahanna (dismissed):

src/config.c
```diff
@@ -639,6 +639,9 @@ static RRStatus getInterfaceAddr(NodeAddr *addr)
     }
 
     for (ent = addrs; ent != NULL; ent = ent->ifa_next) {
+        if (ent->ifa_addr == NULL)
+            continue;
```
tezc, line 643:
```suggestion
        if (ent->ifa_addr == NULL) {
            continue;
        }    
```

src/config.c
```diff
@@ -639,6 +639,10 @@ static RRStatus getInterfaceAddr(NodeAddr *addr)
     }
 
     for (ent = addrs; ent != NULL; ent = ent->ifa_next) {
+        if (ent->ifa_addr == NULL) {
+            continue;
+        }
```
tezc, line 644:
```suggestion
        }
```

tezc (approved):

tezc:
@doflink Thank you very much for the fix!

Issue #470, opened 2022-08-04
Misleading assert in raftNotifyMembershipEvent


jhershberg:
In the function raftNotifyMembershipEvent there is a line:
` assert(entry->type == RAFT_LOGTYPE_ADD_NODE || entry->type == RAFT_LOGTYPE_ADD_NONVOTING_NODE);`
This line gives the impression that raftNotifyMembershipEvent is invoked for both of these kinds of LOGTYPES. This caused me to wonder how there is no mem leak in the continuation of the code since a node obj is allocated and if this function were invoked twice, once for non-voting and once for add_node the second udata would overwrite the first and hence never be cleaned. Poking around the code it seems that raftNotifyMembershipEvent is in fact NEVER invoked for nonvoting.

I submit that the assert should be changed to only check for add_node since the logic that follows in this function does not support this function being invoked for nonvoting and add_node.

PR #471, opened 2022-08-08
Fix misleading assert in raftNotifyMembershipEvent()

tezc:
raftNotifyMembershipEvent() will be called for `RAFT_LOGTYPE_ADD_NODE` only.

Fixes: https://github.com/RedisLabs/redisraft/issues/470

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -973,8 +973,9 @@ void raftNotifyMembershipEvent(raft_server_t *raft, void *user_data,
                 break;
             }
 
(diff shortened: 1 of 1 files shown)


tezc:
Good to see our tests are failing :) So, yeah I was confused about it but `raftNotifyMembershipEvent()` can be called for both non-voting and voting entry types. Important thing is it will be called once per node. A node can be added as voter directly by calling `raft_add_node()` or as non-voter first by calling `raft_add_non_voting_node()`.

PR #473, opened 2022-08-10
RR-24 CLUSTER SHARDS implementation

sjpotter:
implements cluster shards, only outputs "slot ranges" for migrating/stable slots.

outputs all ndoes, currently treats every node as always "healthy" (unsure how to determine otherwise, perhaps if we would record no response within an election timeout, but might need more support for librafT), as wll as every node being fully in sync (offset == 0, as also unsure how to translate that)

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1569,11 +1569,11 @@ static void addClusterShardsReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
             if (type != SLOTRANGE_TYPE_STABLE && type != SLOTRANGE_TYPE_MIGRATING) {
                 continue;
             }
-            slot_count += 2;
+            slot_count += 1;
             RedisModule_ReplyWithLongLong(ctx, sg->slot_ranges[i].start_slot);
             RedisModule_ReplyWithLongLong(ctx, sg->slot_ranges[i].end_slot);
         }
-        RedisModule_ReplySetArrayLength(ctx, slot_count);
+        RedisModule_ReplySetArrayLength(ctx, slot_count * 2);
 
         RedisModule_ReplyWithCString(ctx, "nodes");
         if (sg->local) {
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1554,7 +1554,7 @@ static void addClusterShardsReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
 
         int slot_count = 0;
         RedisModule_ReplyWithArray(ctx, REDISMODULE_POSTPONED_LEN);
-        for (unsigned int i=0; i < sg->slot_ranges_num; i++) {
+        for (unsigned int i = 0; i < sg->slot_ranges_num; i++) {
             /* we only include slot ranges for a shard that are stable / migration */
             SlotRangeType type = sg->slot_ranges[i].type;
             if (type != SLOTRANGE_TYPE_STABLE && type != SLOTRANGE_TYPE_MIGRATING) {
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1483,14 +1483,23 @@ static void addClusterSlotsReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
     RedisModule_ReplySetArrayLength(ctx, num_slots);
 }
 
-static void addClusterShardsNodeReply(RedisModuleCtx *ctx, char *id, uint16_t port, char *host, char *role, int offset, char *health)
+static void addClusterShardsNodeReply(RedisRaftCtx *rr, RedisModuleCtx *ctx, char *id, uint16_t port, char *host, char *role, int offset, char *health)
 {
-    RedisModule_ReplyWithMap(ctx, 6);
+    RedisModule_ReplyWithMap(ctx, 8);
     RedisModule_ReplyWithCString(ctx, "id");
     RedisModule_ReplyWithCString(ctx, id);
-    RedisModule_ReplyWithCString(ctx, "port");
-    RedisModule_ReplyWithLongLong(ctx, port);
+    if (!rr->config.tls_enabled) {
+        RedisModule_ReplyWithCString(ctx, "port");
+        RedisModule_ReplyWithLongLong(ctx, port);
+    } else {
+        RedisModule_ReplyWithCString(ctx, "tls-port");
+        RedisModule_ReplyWithLongLong(ctx, port);
+    }
     RedisModule_ReplyWithCString(ctx, "ip");
+    RedisModule_ReplyWithCString(ctx, "");
+    RedisModule_ReplyWithCString(ctx, "hostname");
+    RedisModule_ReplyWithCString(ctx, host);
+    RedisModule_ReplyWithCString(ctx, "endpoint");
     RedisModule_ReplyWithCString(ctx, host);
     RedisModule_ReplyWithCString(ctx, "role");
     RedisModule_ReplyWithCString(ctx, role);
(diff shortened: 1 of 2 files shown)


src/cluster.c
```diff
@@ -1483,11 +1483,127 @@ static void addClusterSlotsReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
...
+    RedisModule_ReplyWithCString(ctx, "port");
+    RedisModule_ReplyWithLongLong(ctx, port);
+    RedisModule_ReplyWithCString(ctx, "ip");
+    RedisModule_ReplyWithCString(ctx, host);
+    RedisModule_ReplyWithCString(ctx, "role");
+    RedisModule_ReplyWithCString(ctx, role);
+    RedisModule_ReplyWithCString(ctx, "replication-offset");
+    RedisModule_ReplyWithLongLong(ctx, offset);
+    RedisModule_ReplyWithCString(ctx, "health");
+    RedisModule_ReplyWithCString(ctx, health);
```
tezc, line 1500: I think we need to verify if clients are handling this response correctly. Maybe we should try py-redis? 
I see some issues might be problematic, just listing here:

1- `endpoint` field is missing. It isn't optional according to doc. Here, we should return our hostname and port if I understand correctly.

2- Currently, we do not have support for NAT. Apparently, `endpoint` is also used for this purpose according to doc. (Maybe, we can ignore it for now )

3- We are returning hostname for the `ip` field. Not sure clients will be happy with this. (Note: I also don't understand the reasoning behind `ip` field. Hope I'm missing something, otherwise it doesn't look like a good API. Should this be an IP? Why can't this be a hostname?). 

4- Will TLS work with current clients? We are not setting `tls-port` field. Will clients open tls connection even when they don't see `tls-port` field?

tezc: Hmm, this is introduced in 7.0. I guess we have to implement this blindy? Without trying with a client.

sjpotter: good catch, I did it based on documented examples, but I see they list those additional fields in the text portion of the documentation.

https://redis.io/commands/cluster-shards/

tezc: 1- Regarding `endpoint`, problem is in the PR and docs.

For `cluster shards`: Redis adds `endpoint` field to the response.
For `cluster slots`: Endpoint is either `hostname` or `ip` field. There is no separate `endpoint` field.

I'm not sure if this is documented well or the intended behavior. So, I guess we should either fix the code or the documentation after deciding the correct behavior. 

--------
2-
> ip -> do a resolution on hostname

I'm having hard time to understand this field. Are we trying to help the client so it wouldn't need to resolve a hostname? Why should we return this when our cluster is configured with hostnames?. We need to resolve all hostnames in the cluster (including other shard groups)? I think we should decide if it is okay to return empty string if this field is mandatory.  Then, we can just return hostname in the `endpoint` field. Client can resolve the hostname itself.

fadidahanna: * Endpoint - yes, it was added to `cluster slot` in Redis7.0 - we should update this command too (will open a separate ticket for that). Anyway, this field should depend on `cluster-preferred-endpoint-type` configuration in both cases.
* I guess we can stick to the Redis code and suggest fixing to the documentation.Please add Harkrishn Patro (who added `cluster shards`) and Madelyn Olson (`cluster slots` update) to this discussion to ensure the documentation should be updated.
*  Regarding `ip` - why not to keep the ip too? any reason why `NodeAddr.host` can be hostname or ip?

src/cluster.c
```diff
@@ -1483,11 +1483,136 @@ static void addClusterSlotsReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
...
+
+        int slot_count = 0;
+        RedisModule_ReplyWithArray(ctx, REDISMODULE_POSTPONED_LEN);
+        for (unsigned int i = 0; i < sg->slot_ranges_num; i++) {
+            /* we only include slot ranges for a shard that are stable / migration */
+            SlotRangeType type = sg->slot_ranges[i].type;
+            if (type != SLOTRANGE_TYPE_STABLE && type != SLOTRANGE_TYPE_MIGRATING) {
+                continue;
+            }
+            slot_count += 2;
```
fadidahanna, line 1572: For readability, I think it's better to increase only in `1`, and when we set the array length, just to multiply in `2`.

fadidahanna (dismissed):

fadidahanna (approved):

src/cluster.c
```diff
@@ -1483,11 +1483,136 @@ static void addClusterSlotsReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
...
+        RedisModule_ReplyWithCString(ctx, "port");
+        RedisModule_ReplyWithLongLong(ctx, port);
+    } else {
+        RedisModule_ReplyWithCString(ctx, "tls-port");
+        RedisModule_ReplyWithLongLong(ctx, port);
+    }
+    RedisModule_ReplyWithCString(ctx, "ip");
+    RedisModule_ReplyWithCString(ctx, "");
+    RedisModule_ReplyWithCString(ctx, "hostname");
+    RedisModule_ReplyWithCString(ctx, host);
```
tezc, line 1501: IIRC we've decided to put our address in "ip" field as it is a mandatory field. 
I think we shouldn't return "hostname" field as our address can be IP. (this is not a mandatory field anyway).

In summary, we should reply with just these two fields:
1- "endpoint"
2- "ip"

src/cluster.c
```diff
@@ -1483,11 +1483,136 @@ static void addClusterSlotsReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
...
+    RedisModule_ReplyWithCString(ctx, "endpoint");
+    RedisModule_ReplyWithCString(ctx, host);
+    RedisModule_ReplyWithCString(ctx, "role");
+    RedisModule_ReplyWithCString(ctx, role);
+    RedisModule_ReplyWithCString(ctx, "replication-offset");
+    RedisModule_ReplyWithLongLong(ctx, offset);
+    RedisModule_ReplyWithCString(ctx, "health");
+    RedisModule_ReplyWithCString(ctx, health);
```
tezc, line 1509: offset and health are constant, we can have them here hardcoded.

src/cluster.c
```diff
@@ -1483,11 +1483,136 @@ static void addClusterSlotsReply(RedisRaftCtx *rr, RedisModuleCtx *ctx)
...
+                node_count += addClusterShardsLocalNodeReply(rr, ctx, raft_node, leader_node);
+            }
+            RedisModule_ReplySetArrayLength(ctx, node_count);
+        } else {
+            RedisModule_ReplyWithArray(ctx, sg->nodes_num);
+            for (unsigned int j = 0; j < sg->nodes_num; j++) {
+                char *node_id = sg->nodes[j].node_id;
+                uint16_t port = sg->nodes[j].addr.port;
+                char *host = sg->nodes[j].addr.host;
+                char *role = (j == 0) ? "primary" : "replica";
```
tezc, line 1597: I guess something wrong in the documentation.
- Doc mention "primary" but in Redis, it uses "master". 
- Also, doc says `health: Either online, failed, or loading.` . Example uses `fail`.

tezc:
Some test failures with TLS: https://github.com/RedisLabs/redisraft/runs/8295638622?check_suite_focus=true

tezc:
Seems like one failure is not related with TLS: https://github.com/RedisLabs/redisraft/runs/8299341042?check_suite_focus=true#step:10:157

sjpotter:
> Seems like one failure is not related with TLS: https://github.com/RedisLabs/redisraft/runs/8299341042?check_suite_focus=true#step:10:157



> Some test failures with TLS: https://github.com/RedisLabs/redisraft/runs/8295638622?check_suite_focus=true

have a fix for this using the pytestconfig fixture

sjpotter:
> Seems like one failure is not related with TLS: https://github.com/RedisLabs/redisraft/runs/8299341042?check_suite_focus=true#step:10:157

this is an output ordering issue.  investigating

PR #474, opened 2022-08-10
RR-23 cluster nodes special migrtion info lines

sjpotter:
adds the [slot-<direction>-node_id] lines to cluster nodes output.

modifies test to validate

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1305,7 +1305,7 @@ static void appendSpecialClusterNodeString(RedisModuleString *ret, unsigned int 
     size_t len;
     const char *temp;
 
-    RedisModuleString *str = RedisModule_CreateStringPrintf(NULL, "%u-%s-%.*s\r\n", j, direction, 40, node->node_id);
+    RedisModuleString *str = RedisModule_CreateStringPrintf(NULL, "[%u-%s-%.*s]\r\n", j, direction, 40, node->node_id);
     temp = RedisModule_StringPtrLen(str, &len);
     RedisModule_StringAppendBuffer(NULL, ret, temp, len);
     RedisModule_FreeString(NULL, str);
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1314,7 +1314,7 @@ static void appendSpecialClusterNodeString(RedisModuleString *ret, unsigned int 
             direction = ">";
             break;
         default:
-            return;;
+            return;
     }
 
     RedisModuleString *str = RedisModule_CreateStringPrintf(NULL, "[%u-%s-%.*s]\r\n", j, direction, 40, node->node_id);
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1300,10 +1300,22 @@ static void appendClusterNodeString(RedisModuleString *ret, char node_id[41], No
     RedisModule_FreeString(NULL, str);
 }
 
-static void appendSpecialClusterNodeString(RedisModuleString *ret, unsigned int j, char *direction, ShardGroupNode *node)
+static void appendSpecialClusterNodeString(RedisModuleString *ret, unsigned int j, SlotRangeType type, ShardGroupNode *node)
 {
     size_t len;
     const char *temp;
+    char *direction;
+
+    switch (type) {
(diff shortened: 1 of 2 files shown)


tests/integration/test_sharding.py
```diff
@@ -1,4 +1,4 @@
-"""
+f"""
```
fadidahanna, line 1:
```suggestion
"""
```

tests/integration/test_sharding.py
```diff
@@ -211,6 +215,7 @@ def test_shard_group_validation(cluster):
         'sharding': 'yes',
         'slot-config': '0:1000'})
 
+    cluster_dbid = cluster.leader_node().info()['raft_dbid']
```
fadidahanna, line 218: can be removed

src/cluster.c
```diff
@@ -1346,6 +1361,28 @@ static void addClusterNodeReplyFromNode(RedisRaftCtx *rr,
...
+                    if (migrating_map[j] != NULL && migrating_map[j]->nodes_num > 0) {
+                        appendSpecialClusterNodeString(ret, j, "<", &migrating_map[j]->nodes[0]);
+                    }
+                }
+            } else if (sr.type == SLOTRANGE_TYPE_MIGRATING) {
+                for (unsigned int j = sr.start_slot; j <= sr.end_slot; j++) {
+                    RedisModule_Assert(j < REDIS_RAFT_HASH_SLOTS);
+                    if (importing_map[j] != NULL && importing_map[j]->nodes_num > 0) {
+                        appendSpecialClusterNodeString(ret, j, ">", &importing_map[j]->nodes[0]);
+                    }
```
fadidahanna, line 1381: Actually the behavior for these two cases is the same, the only difference is the angle bracket direction.

src/cluster.c
```diff
@@ -1300,12 +1300,27 @@ static void appendClusterNodeString(RedisModuleString *ret, char node_id[41], No
     RedisModule_FreeString(NULL, str);
 }
 
+static void appendSpecialClusterNodeString(RedisModuleString *ret, unsigned int j, char *direction, ShardGroupNode *node)
+{
+    size_t len;
+    const char *temp;
+
+    RedisModuleString *str = RedisModule_CreateStringPrintf(NULL, "[%u-%s-%.*s]\r\n", j, direction, 40, node->node_id);
```
fadidahanna, line 1308: In Redis max length of the node_id is 40 (it's true for us too, but let's keep the same format)

src/cluster.c
sjpotter: I don't understand the Q?  do you mean instead of passing in the "char", we hsould pass in the type and appendSpecialClusterNodeString should handle the convesion to > or <?

src/cluster.c
sjpotter: I dont understand the Q, aren't I limiting it to 40 here?

fadidahanna (approved):

fadidahanna (approved):

src/cluster.c
```diff
@@ -1300,12 +1300,39 @@ static void appendClusterNodeString(RedisModuleString *ret, char node_id[41], No
...
+            direction = "<";
+            break;
+        case SLOTRANGE_TYPE_MIGRATING:
+            direction = ">";
+            break;
+        default:
+            return;
+    }
+
+    RedisModuleString *str = RedisModule_CreateStringPrintf(NULL, "[%u-%s-%.*s]\r\n", j, direction, 40, node->node_id);
```
tezc, line 1320: If I understand correctly, separator should be space, not `\r\n`: https://github.com/redis/redis/blob/36abc0fa8f50a0c5bbe8b5ddcca657d6ada94363/src/cluster.c#L4775

Also, looks like whole `cluster node` format separated with `\n`, not `\r\n`: 
https://github.com/redis/redis/blob/36abc0fa8f50a0c5bbe8b5ddcca657d6ada94363/src/cluster.c#L4858

I think, the best way is to get raw output from Redis and mimic it byte by byte (seems like documentation might be misleading).
Probably, we should do it for other commands(cluster slots, cluster shards) as well, not to miss anything.

PR #476, opened 2022-08-20
Delete unnecessary raft_apply_all() call

tezc:
Raft library already handles when to apply the entries. Explicit call is unnecessary.


fadidahanna:
@tezc, which path do you refer to? please pay attention that  auto_flush configuration is set to 0 in RR...

tezc:
@fadidahanna `raft_flush()` call will apply() in our case.

Whether `auto_flush()` is set or not, libraft users should never call `raft_apply()` family functions. It's libraft's job to decide when it is safe to apply an entry. (Libraft should not even expose `raft_apply()`).

fadidahanna (approved):
Please don't forget to clean the lib raft api, and remove `raft_apply_all()` from the API.

PR #489, opened 2022-09-07
[RR-20] move oom to use dry run

sjpotter:
changes from manually doing it to using dry run, would be a performance regression but will probably need it for acl in future anyways.

didn't require any test changes

### DIFF
--- a/src/redisraft.c
+++ b/src/redisraft.c
@@ -651,20 +651,9 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
         RaftRedisCommand *cmd = cmds->commands[i];
         size_t cmdlen;
         const char *cmdstr = RedisModule_StringPtrLen(cmd->argv[0], &cmdlen);
-        /* skip multi */
-        if (i == 0 && cmdlen == 5 && !strncasecmp(cmdstr, "MULTI", 5)) {
-            continue;
-        }
-
-        char *rm_call_flags;
-        if (RedisModule_GetUsedMemoryRatio() > 1.0) {
-            rm_call_flags = "DCEMv";
-        } else {
-            rm_call_flags = "DCEv";
-        }
 
         enterRedisModuleCall();
-        RedisModuleCallReply *reply = RedisModule_Call(ctx, cmdstr, rm_call_flags, cmd->argv + 1, cmd->argc - 1);
+        RedisModuleCallReply *reply = RedisModule_Call(ctx, cmdstr, "DCEMv", cmd->argv + 1, cmd->argc - 1);
         exitRedisModuleCall();
         if (reply != NULL) {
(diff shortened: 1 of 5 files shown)


src/redisraft.c
```diff
@@ -646,9 +646,32 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
...
-    if (cmd_flags & CMD_SPEC_DENYOOM && (RedisModule_GetUsedMemoryRatio() > 1.0)) {
-        RedisModule_ReplyWithError(ctx, "OOM command not allowed when used memory > 'maxmemory'.");
-        return;
+    /* dry run */
+    for (int i = 0; i < cmds->len; i++) {
+        RaftRedisCommand *cmd = cmds->commands[i];
+        size_t cmdlen;
+        const char *cmdstr = RedisModule_StringPtrLen(cmd->argv[0], &cmdlen);
+        /* skip multi */
+        if (i == 0 && cmdlen == 5 && !strncasecmp(cmdstr, "MULTI", 5)) {
```
tezc, line 655: Why do we skip multi? If I'm not mistaken, ACL can reject it? Even it won't reject it, do we need this special check?

src/redisraft.c
```diff
@@ -646,9 +646,32 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
...
+            continue;
+        }
+
+        char *rm_call_flags;
+        if (RedisModule_GetUsedMemoryRatio() > 1.0) {
```
tezc, line 660: As we are going to call dry-run unconditionally, we don't need to do this check here. rm_call() does more or less same thing, checks usage ratio + one extra if check for DENY_OOM flag. So, I think we can pass "DCEMv" directly. 

Hopefully, we can improve rm_call() performance later.

tezc (approved):

tezc:
@sjpotter Maybe as part of this PR, maybe in another PR, it'd be nice if we can add OOM tests with scripts. With or without allow-oom flags etc.

sjpotter:
so we have a PR already for scripts, but depends on redis updates that haven't gone in yet.

tezc:
Okay, we'll merge those tests later with that PR then.

fadidahanna:
Do we still need `CMD_SPEC_DENYOOM`? I guess the answer is no, so maybe remove it too from our code, to keep it cleaner.

fadidahanna (approved):

tests/integration/test_config.py
```diff
@@ -278,17 +278,6 @@ def test_module_command_flags(cluster):
     dont_intercept = 1 << 4
     denyoom = 1 << 7
```
fadidahanna, line 279: can be removed too. Also in 'hellomodule.c`

tezc (approved):

tezc (approved):

PR #490, opened 2022-09-11
Delete unused variables to silence linter warning

tezc:
Delete unused variables and superfluous assignment to silence linter.

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -1620,23 +1620,19 @@ void replaceShardGroups(RedisRaftCtx *rr, raft_entry_t *entry)
     ShardingInfo *si = rr->sharding_info;
 
     if (si->shard_group_map != NULL) {
+        void *data;
         RedisModuleDictIter *iter = RedisModule_DictIteratorStartC(si->shard_group_map, "^", NULL, 0);
 
-        char *key;
-        size_t key_len;
-        ShardGroup *data;
-
-        while ((key = RedisModule_DictNextC(iter, &key_len, (void **) &data)) != NULL) {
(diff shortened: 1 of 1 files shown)


src/raft.c
```diff
@@ -1620,23 +1620,19 @@ void replaceShardGroups(RedisRaftCtx *rr, raft_entry_t *entry)
     ShardingInfo *si = rr->sharding_info;
 
     if (si->shard_group_map != NULL) {
+        void *data;
```
fadidahanna, line 1623: Better to keep it in the right type - `ShardGroup *`

src/raft.c
```diff
@@ -1620,23 +1620,19 @@ void replaceShardGroups(RedisRaftCtx *rr, raft_entry_t *entry)
...
-        size_t key_len;
-        ShardGroup *data;
-
-        while ((key = RedisModule_DictNextC(iter, &key_len, (void **) &data)) != NULL) {
+        while (RedisModule_DictNextC(iter, NULL, &data) != NULL) {
             ShardGroupFree(data);
         }
         RedisModule_DictIteratorStop(iter);
         RedisModule_FreeDict(rr->ctx, si->shard_group_map);
-        si->shard_group_map = NULL;
```
fadidahanna, line 1634: It's better to keep `si->shard_group_map = NULL;`, although in the current implementation we immediately set a new value, in order to avoid using freed memory in the future.

src/raft.c
tezc: sure

src/raft.c
tezc: sure

fadidahanna (approved):

PR #491, opened 2022-09-11
Explicitly compare strcmp result

tezc:
Explicitly compare strcmp result to silence linter

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -333,7 +333,7 @@ error:
 static int handleHeader(RaftLog *log, RawLogEntry *re)
 {
     if (re->num_elements != 6 ||
-        strcmp(re->elements[0].ptr, "RAFTLOG")) {
+        strcmp(re->elements[0].ptr, "RAFTLOG") != 0) {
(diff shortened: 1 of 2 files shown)


fadidahanna:
@tezc, there are a bunch of other places in the code that use `strcmp()` without explicit comparison (see, for example, https://github.com/RedisLabs/redisraft/blob/master/src/config.c#L833-L840, https://github.com/RedisLabs/redisraft/blob/master/src/node_addr.c#L59, etc). Don't see why it's so critical to change it, but if you do so, please change it all over the code.

tezc:
`!strcmp()` <-- See the starting `!` .  <-- So, your examples are considered as "explicit checking". 

Clang-tidy complains about this because it thinks you make a decision without checking the return code of `strcmp()`.
Maybe one day this warning might help if you forget checking the return code but I don't care about this specific check. I just want to have clang-tidy clean code base eventually.

fadidahanna:
k, if it's for a linter reason...

PR #493, opened 2022-09-12
Delete unnecessary checks for node promotion

tezc:
- `raftNodeHasSufficientLogs()` callback cannot be called on loading state as we start using libraft after loading is completed.
- `raft_node_is_active()` check seems unnecessary anymore. This callback won't be called if the node is inactive.
   - https://github.com/RedisLabs/redisraft/blob/f389775cf166076976f0c05e6b3ffc8851a5f6f4/deps/raft/src/raft_server.c#L765
   
    `raftNodeHasSufficientLogs()` callback is only called when there is no ongoing voting change. If a node is not active, it means `RAFT_LOGTYPE_REMOVE_NODE` config change has been appended and config change is in progress. It means, there is an ongoing voting change. So, libraft will not call this callback.

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -891,25 +891,6 @@ static raft_node_id_t raftLogGetNodeId(raft_server_t *raft, void *user_data,
 
 static int raftNodeHasSufficientLogs(raft_server_t *raft, void *user_data, raft_node_t *raft_node)
 {
-    RedisRaftCtx *rr = (RedisRaftCtx *) user_data;
-    if (rr->state == REDIS_RAFT_LOADING)
-        return 0;
-
-    /* Node may have sufficient logs to be promoted but be scheduled for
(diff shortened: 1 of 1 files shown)


tezc:
Before placing a validation check, I look for a possibility of a bug, future misuse etc. (not to make code longer unnecessarily).

In that sense, "loading" check does not make sense to me because if that's true, lots of things will fail before and after this function anyway. In another sense, we can add that check into all functions. There is nothing special about this function for this check.

Regarding, `raft_node_is_active()` check, libraft should guarantee it is not calling this for non-active nodes. Also, libraft users should not use this function (this function is part of private API). Second, `raft_recv_entry()` will fail if there is a inactive node. I can say we are covered very well but still adding that as assert just to find the middle ground :)

fadidahanna (approved):

PR #498, opened 2022-09-19
Update libraft

tezc:
Updated libraft.

On RedisRaft side, there is only one fix for snapshot testing. (required after https://github.com/RedisLabs/raft/pull/143)

Rather than manually sending the snapshot,
- We stop the target node, 
- Commit an entry
- Make the leader take a snapshot
- Restart the target instance.
- Leader will deliver the snapshot as the target node is missing the latest entry. 

This way, we can trigger snapshot delivery in a well-defined way 
(without relying on our debug "sendsnapshot" command).


src/redisraft.h
```diff
@@ -25,6 +25,7 @@
 #include <openssl/ssl.h>
 #endif
 #include "raft.h"
+#include "raft_private.h"
```
fadidahanna, line 28: well, if it's a private API, why does RR need to use it?

tezc: https://github.com/RedisLabs/redisraft/blob/a6bf17d3f07be8ff648ca4236f9d2a8f54e67793/src/redisraft.c#L1394

Just for testing, we use some private functions. For example, here we manually set node inactive. Normally, application should never set active status of the node. It will break the cluster safety. 

We may come up with something else, maybe we can get rid of these calls but in another sense, this a good example of private API. You use these functions if you know what you are doing. Wdyt?

fadidahanna: Let's first ignore the tests - I think we should have a single API, even for users that know exactly what they're doing. The logic should be kept in the raft library.
As for testing - need a deep investigation to see what exactly we are testing, and maybe some scenarios can move to the libraft itself, instead of testing in RR side. Meanwhile let's merge it, and open a ticket for future clean up of private API calls from RR.

fadidahanna (approved):

src/redisraft.h
tezc: Meanwhile, I've realized I shouldn't have moved `raft_node_set_voting()` function to the private API. (It's required after the snapshot: https://github.com/RedisLabs/raft/blame/master/docs/Using.md#L306

I'll create a PR in libraft to move this function back, maybe I'll move a few others back as well.

tezc: https://github.com/RedisLabs/raft/pull/144

I've moved all the functions we required back to the raft.h  . RR will not include private API.

fadidahanna (approved):

PR #499, opened 2022-09-23
Delete duplicate error reply on migration

tezc:
We already reply in the next line according to `raft_recv_entry()` result. 
Duplicate reply breaks the protocol.

### DIFF
--- a/src/migrate.c
+++ b/src/migrate.c
@@ -205,7 +205,6 @@ static void raftAppendRaftUnlockDeleteEntry(RedisRaftCtx *rr, RaftReq *req)
 
     int e = raft_recv_entry(rr->raft, entry, NULL);
     if (e != 0) {
-        RedisModule_ReplyWithError(req->ctx, "ERR Unable to unlock/delete migrated keys, try again");
         replyRaftError(req->ctx, e);
         entryDetachRaftReq(rr, entry);
         raft_entry_release(entry);


src/migrate.c
```diff
@@ -205,7 +205,6 @@ static void raftAppendRaftUnlockDeleteEntry(RedisRaftCtx *rr, RaftReq *req)
 
     int e = raft_recv_entry(rr->raft, entry, NULL);
     if (e != 0) {
-        RedisModule_ReplyWithError(req->ctx, "ERR Unable to unlock/delete migrated keys, try again");
         replyRaftError(req->ctx, e);
```
fadidahanna, line 208: I think we need to decide which error msg is more informative in this flow. We can probably have both, using `raft_get_error_str()` instead of`replyRaftError()`

tezc: We still need to check for `RAFT_ERR_SHUTDOWN`. libraft can return it when we have to shutdown, e.g something seriously wrong. 

`replyRaftError()` checks it:
https://github.com/RedisLabs/redisraft/blob/da0d3416a833ba665d63fabac349a51dd5b300dd/src/common.c#L45

So, overall, I'm not sure about `RAFT_ERR_SHUTDOWN` error. What is the point of returning it when only thing we have to do is shutting down? Probably, we should just trigger the assert inside libraft. I'm pretty sure either us or other libraft users will miss checking this value. Wdyt?

fadidahanna (approved):

src/migrate.c
sjpotter: as @fadidahanna said, I think the first message is more informative.  I think we perhaps need a replyRaftErrorWithMessage(ctx, msg), that doesn't take e.  and calls a common code that will also be used by replyRaftError() (or perhaps replyRaftError will now wrap replyRaftErroWithMessage.

tezc:
Closing in favor of https://github.com/RedisLabs/redisraft/pull/574

PR #502, opened 2022-10-03
fix daily testing errors

sjpotter:
fix encodeString null bug and and fix macos build failure

### DIFF
--- a/src/serialization_utils.c
+++ b/src/serialization_utils.c
@@ -142,6 +142,7 @@ int encodeString(char *p, size_t sz, RedisModuleString *str)
 
     if (str == NULL) {
         len = 0;
+        e = "";
     } else {
         e = RedisModule_StringPtrLen(str, &len);
     }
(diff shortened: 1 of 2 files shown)


src/serialization_utils.c
```diff
@@ -142,6 +142,7 @@ int encodeString(char *p, size_t sz, RedisModuleString *str)
 
     if (str == NULL) {
         len = 0;
+        e = "";
```
tezc, line 145: Just a question, can we avoid these null checks altogether that we've introduced for ACL? Can we use empty string for non-existing ACLs rather than NULL? Does it make things worse?

fadidahanna (approved):

src/serialization_utils.c
sjpotter: its not a "char *" its a RedisModuleString so would involve an allocation and free every single time, I think a NULL check is better

tezc (approved):

PR #510, opened 2022-10-17
Use SEEK_SET to set position of the snapshot file

tezc:
Use `SEEK_SET` to set file position while storing snapshot chunks.

In `raftStoreSnapshotChunk()`, `offset` indicates position relative 
to the head of the file. 

Currently, we use `SEEK_CUR` which sets the position relative to the
current position. As we reopen file each time, current position is zero
and it works correctly but better to use `SEEK_SET` which sets position
relative to the head of the file. (Just to prevent future bugs).

### DIFF
--- a/src/snapshot.c
+++ b/src/snapshot.c
@@ -139,7 +139,7 @@ int raftStoreSnapshotChunk(raft_server_t *raft, void *user_data,
         return -1;
     }
 
-    off_t ret_offset = lseek(fd, offset, SEEK_CUR);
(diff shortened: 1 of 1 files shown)


fadidahanna:
Agree with the fix. You probably want to change it also in `app_raft_store_snapshot_impl()` (example implementation in our documentation)

tezc:
How did you remember it? :)  

Okay, opened a PR for it: https://github.com/RedisLabs/raft/pull/150

fadidahanna (approved):

PR #511, opened 2022-10-17
[RR-94] Simplify RaftLogDelete()

tezc:
In `RaftLogDelete()`, no need to loop over entries, read or parse.

We can immediately truncate the file as truncating to the offset of the
first entry will delete the following entries.

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -652,47 +652,26 @@ raft_entry_t *RaftLogGet(RaftLog *log, raft_index_t idx)
 
 RRStatus RaftLogDelete(RaftLog *log, raft_index_t from_idx)
 {
-    off_t offset;
-    RRStatus ret = RR_OK;
-
     if (from_idx <= log->snapshot_last_idx) {
         return RR_ERROR;
     }
 
-    while (log->index >= from_idx) {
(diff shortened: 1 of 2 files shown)


src/log.c
```diff
@@ -652,47 +652,26 @@ raft_entry_t *RaftLogGet(RaftLog *log, raft_index_t idx)
...
             return RR_ERROR;
         }
 
-        RawLogEntry *re;
-        raft_entry_t *e;
-
-        if (readRawLogEntry(log->file, &re) < 0) {
-            ret = RR_ERROR;
-            break;
+        if (ftruncate(fileno(log->file), entry_offset) < 0) {
```
fadidahanna, line 666: Did we have in the past different types of log entries? what was the reason for `ENTRY` comparison?  what was the reason for the entry validation here?

tezc: I guess there was some sort of an experimentation with different types of entries. Not sure what was the experiment though, can't relate with Raft.

fadidahanna: k, so we can treat it as leftover probably...

fadidahanna (approved):

PR #513, opened 2022-10-18
[RR-95] adding elle infrastructure for redisraft tests

sjpotter:
Add a POC of using elle (not a real test, just a demo for now) with RedisRaft in its pytest tests


.github/workflows/ci.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Install build dependencies
-      run: sudo apt-get install -y build-essential autoconf automake libtool cmake lcov
+      run: sudo apt-get install -y build-essential autoconf automake libtool cmake lcov leiningen default-jdk-headless
```
tezc, line 11: I guess we want to run these new tests for a while. Probably, better to use daily for it, not to slow-down CI. 
We can have a job there just for elle. Similar to `test-macos`, `test-tls-sanitizer`, we can have `test-elle`.

tests/integration/conftest.py
```diff
@@ -42,8 +43,10 @@ def pytest_addoption(parser):
         '--tls', default=False, action='store_true',
         help='Use tls')
     parser.addoption(
-        "--runslow", action="store_true", default=False, help="run slow tests"
-    )
+        "--runslow", action="store_true", default=False, help="run slow tests")
+    parser.addoption(
+        '--elle-cli', default='../elle-cli/target/elle-cli-0.1.4-standalone.jar',
```
tezc, line 48: I think it'd be nice to run `elle` tests only when this option is given. Similar to above `--runslow` option and `--tls` option. Because, when we want to run tests on a new machine, we'll need to install those new packages. It'll be painful. Let's make `elle` test optional.

.github/workflows/ci.yml
sjpotter: created an elle test suite for daily

tests/integration/conftest.py
sjpotter: done

.github/workflows/daily.yml
```diff
@@ -214,7 +253,7 @@ jobs:
           ref: 'unstable'
           path: 'redis'
       - name: Build Redis
-        run: cd redis && make -j 4 SANITIZER=address BUILD_TLS=yes
+        run: cd redis && make -j 4 SANITIZER=address BUILD_TLS=yesa
```
tezc, line 256:
```suggestion
        run: cd redis && make -j 4 SANITIZER=address BUILD_TLS=yes
```

.github/workflows/ci.yml
tezc, line 11:
```suggestion
      run: sudo apt-get install -y build-essential autoconf automake libtool cmake lcov
```

So, this is not necessary anymore

.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install build dependencies
-        run: sudo apt-get install -y build-essential autoconf automake libtool cmake lcov
+        run: sudo apt-get install -y build-essential autoconf automake libtool cmake lcov leiningen default-jdk-headless
```
tezc, line 48:
```suggestion
        run: sudo apt-get install -y build-essential autoconf automake libtool cmake lcov
```

.github/workflows/daily.yml
```diff
@@ -19,6 +19,45 @@ on:
         default: 'master'
 
 jobs:
+  tests-with-elle:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install build dependencies
+        run: sudo apt-get install -y build-essential autoconf automake libtool cmake lcov leiningen
```
tezc, line 27:
```suggestion
        run: sudo apt-get install -y build-essential autoconf automake libtool cmake lcov leiningen default-jdk
```
I guess we'll need jdk here.

tests/integration/test_sharding.py
```diff
@@ -1119,9 +1118,9 @@ def validate_shards(cluster_shards):
             else:
                 assert False, "didn't match %s" % cluster_shards
 
-
     validate_shards(cluster.node(1).execute('CLUSTER', 'SHARDS'))
 
+
```
tezc, line 1123: Changes in this file are leftover right, not related

tests/integration/sandbox.py
```diff
@@ -718,6 +722,188 @@ def restart(self):
...
+            return ret
+
+
+class ElleWorker(object):
+    def __init__(self, elle: Elle, clients: typing.Dict[str, redis.Redis]):
+        self.clients = clients
+        self.elle = elle
+        self.id = threading.get_ident()
+
+    # might not want generate_ops, perhaps each multi should just be a single append/read of same key
```
tezc, line 833: I don't see any difference between single append vs single append + read. Read version is probably better, gives more opportunities to catch a bug. So, I guess we should make it append + read.

tezc:
Just to be clear, when we discussed whether Elle job can run in the background, I thought Elle job would not do anything specific to test and test case would not do anything specific to Elle job. If we can do it, we can pick some tests to run with Elle by using `mark`s. That's why I mentioned `mark` stuff. So, it'd be as simple as:

```python
@python.mark.elle ## <--- Only thing to add
def test_leader_removal_allowed(cluster):
    """
    Leader node can be removed.
    """

    cluster.create(3)
    assert cluster.leader == 1
    cluster.remove_node(1)
    cluster.node(2).wait_for_num_nodes(2)
```

```python
@python.mark.elle ## <--- Only thing to add
def test_migration_basic(cluster_factory):
....
```

That way we can reuse some tests and we can write new tests without hurting its readability (by writing Elle related stuff into it).  I thought that was our discussion a week ago, maybe I misunderstood. 

After we have that, ofcourse we can add an option to take parameters if we think it is useful for a test/tests, e.g: 
```python
@python.mark.elle("key-prefix={test}, worker-count=5")
```

I guess the last test in the PR is the example as it has less Elle specific thing. Currently, IIUC it won't work if we add/remove nodes or when we want to create the cluster with some config parameters?  Maybe, fixture is not the answer here? We need a way to update node list of Elle workers whenever a node is created/removed in any cluster, no?

.github/workflows/daily.yml
```diff
@@ -19,6 +19,45 @@ on:
...
+          cmake .. -DPYTEST_OPTS="--redis-executable=redis/src/redis-server --elle-cli=elle-cli/target/elle-cli-0.1.4-standalone.jar --elle -v"
+          make
+      - name: Checkout Redis
+        uses: actions/checkout@v2
+        with:
+          repository: 'redis/redis'
+          ref: 'unstable'
+          path: 'redis'
+      - name: Build Redis
+        run: cd redis && make -j 4 gcov
```
tezc, line 40:
```suggestion
        run: cd redis && make -j 4
```

I guess this is copy paste error from coverage build. Let's delete it

.github/workflows/daily.yml
```diff
@@ -19,6 +19,45 @@ on:
...
+  tests-with-elle:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install build dependencies
+        run: sudo apt-get install -y build-essential autoconf automake libtool cmake lcov leiningen
+      - name: Build
+        run: |
+          mkdir build && cd build
+          cmake .. -DPYTEST_OPTS="--redis-executable=redis/src/redis-server --elle-cli=elle-cli/target/elle-cli-0.1.4-standalone.jar --elle -v"
```
tezc, line 31:
```suggestion
          cmake .. -DPYTEST_OPTS="--redis-executable=redis/src/redis-server --elle -v"
```

If this is same with the default path, let's delete it to keep here a bit simpler

tests/integration/sandbox.py
```diff
@@ -718,6 +722,227 @@ def restart(self):
...
+        # if this throws an exception, we have a failure elsewhere
+        self.index[thread] += 1
+
+        output = self.generate_result_output(thread, result, results, ops)
+        self.lock.acquire()
+        self.logfile.write(output)
+        self.logfile.write("\n")
+        self.lock.release()
+
+    def log_comment(self, msg: str):
```
tezc, line 808: Is this unused?

tests/integration/sandbox.py
```diff
@@ -733,3 +958,51 @@ def assert_after(func, timeout, retry_interval=0.5):
...
+
+    crc = 0
+    for byte in data:
+        crc = ((crc << 8) & 0xff00) ^ x_mode_m_crc16_lookup[((crc >> 8) & 0xff) ^ byte]
+    return crc & 0xffff
+
+
+def key_hash_slot(key: str) -> int:
+    assert key is not None
+    return crc16(key.encode()) % 16384
```
tezc, line 1008:
```suggestion
    return crc16(key.encode()) % 16384

```

tests/integration/test_elle.py
```diff
@@ -0,0 +1,293 @@
...
+RedisRaft is licensed under the Redis Source Available License (RSAL).
+"""
+import time
+import pytest as pytest
+
+from .sandbox import ElleWorker, Elle, key_hash_slot
+
+
+@pytest.mark.skipif("not config.getoption('elle_threads')")
+def test_elle_sanity(created_clusters):
```
tezc, line 15: Is this a leftover?

tests/integration/test_elle.py
```diff
@@ -0,0 +1,293 @@
...
+        'external-sharding': 'yes'})
+    cluster2 = cluster_factory().create(3, raft_args={
+        'sharding': 'yes',
+        'external-sharding': 'yes'})
+
+    cluster1_dbid = cluster1.leader_node().info()["raft_dbid"]
+    cluster2_dbid = cluster2.leader_node().info()["raft_dbid"]
+
+    slot = -1
+    marker = request.node.get_closest_marker("key_hash_tag")
```
tezc, line 164: This prevents making errors but I don't think it is sustainable if we want to reuse tests without polluting it. 

Initially, I thought we would have 16k keys, one per slot. Then, test would run by picking keys randomly on each operation. So, we wouldn't need to specify anything. Maybe this is not so efficient for short running tests, also not sure if Elle is efficient with lots of keys. Manually marking keys make sense but I think we need to extend this to multi slot in future.

tests/integration/test_elle.py
```diff
@@ -0,0 +1,293 @@
...
+        'sharding': 'yes',
+        'external-sharding': 'yes'})
+
+    cluster1_dbid = cluster1.leader_node().info()["raft_dbid"]
+    cluster2_dbid = cluster2.leader_node().info()["raft_dbid"]
+    slot = key_hash_slot("test")
+
+    client_map = elle.map_addresses_to_clients([cluster1, cluster2])
+
+    worker = ElleWorker(elle, client_map)
```
tezc, line 34: Was this test case for POC? Do we need to do things manual?

tests/integration/sandbox.py
```diff
@@ -718,6 +722,227 @@ def restart(self):
...
+
+    def do_ops(self, ops: typing.List[ElleOp]):
+        # needs a way to handle asking
+        good_write = False
+        needs_asking = False
+
+        client = redis.Redis(host="localhost", port=self.clusters[0].leader_node().port)
+        self.elle.log_command(self.id, ops)
+
+        # conn.send_command('ASKING')
```
tezc, line 871: leftover?

tests/integration/pytest.ini
```diff
@@ -3,3 +3,8 @@ log_file=tests/tmp/tests.log
 log_file_format=%(message)s
 log_file_level=debug
 timeout=600
+markers = 
+    num_clusters: how many clusters should be created by created_clusters fixture
```
tezc, line 7: I guess this is not required anymore

tests/integration/conftest.py
```diff
@@ -133,3 +184,39 @@ def workload():
     _workload = Workload()
     yield _workload
     _workload.terminate()
+
+
+@pytest.fixture()
+def created_clusters(request, cluster_factory, elle):
```
tezc, line 190: I guess this is not required anymore

tezc:
@sjpotter I think this PR needs some cleanups, there are some leftovers from the previous version.

.github/workflows/ci.yml
sjpotter: why isn't it necessary? elle-cli needs leinigen and jdk to build/run

sjpotter: oh, i see, because in regular ci, while elle only runs in daily.  part of me thinks the environments should be the same, even if not directly used now, as would make it easier to move tests around without wondering its not running.

.github/workflows/ci.yml
sjpotter: see my comment above

.github/workflows/daily.yml
sjpotter: yes, but made headless and to be consistent with above comments

tests/integration/test_sharding.py
sjpotter: not related, but I think bug fixes.  though this extra line is removable.

sjpotter: actually, no this extra line was also needed for py linting

tests/integration/sandbox.py
sjpotter: ok, I think thats what it is now, right?

tests/integration/sandbox.py
sjpotter: for now, the thought was that it could add comment lines to the log file produced for elle so could add "marks" so can understand output better.

tests/integration/test_elle.py
sjpotter: no, this is the simplest example of an elle background test (i.e. test does nothing, just "elle" stuff running in background.

tests/integration/test_elle.py
sjpotter: I believe marks can take arrays, so we should be able to extend it with that (and have a different mark if we want to make sure it touches all slots)

tests/integration/test_elle.py
sjpotter: I think there's value in doing tests not just as background to an existing test.  but am open to removing it.

tests/integration/conftest.py
sjpotter: removed

tests/integration/pytest.ini
sjpotter: removed

tests/integration/sandbox.py
sjpotter: removed

.github/workflows/ci.yml
tezc: but jdk is often quite big and it comes with some other packages. I see it is around 250mb++.
Let's not do it. It's too big, it will slow-down the build and contribute to global warming unnecessarily :)

.github/workflows/daily.yml
```diff
@@ -19,6 +19,45 @@ on:
         default: 'master'
 
 jobs:
+  tests-with-elle:
```
tezc, line 22: Can you add manual dispatch stuff similar to other jobs?

To select jobs and to prevent running on forks:

```
    if: |
      (github.event_name == 'workflow_dispatch' ||
      (github.event_name == 'schedule' && github.repository == 'redislabs/redisraft')) && !contains(github.event.inputs.skipjobs, 'elle')
```

To set the repo:
```
    - name: prep
      if: github.event_name == 'workflow_dispatch'
      run: |
        echo "GITHUB_REPOSITORY=${{github.event.inputs.use_repo}}" >> $GITHUB_ENV
        echo "GITHUB_HEAD_REF=${{github.event.inputs.use_git_ref}}" >> $GITHUB_ENV
```

tests/integration/sandbox.py
```diff
@@ -762,6 +770,216 @@ def restart(self):
...
+        self.logfile.write("\n")
+        self.lock.release()
+
+    def log_comment(self, msg: str):
+        self.lock.acquire()
+        self.logfile.write("; " + msg + "\n")
+        self.lock.release()
+
+    @staticmethod
+    def map_addresses_to_clients(clusters: List[Cluster]) -> typing.Dict[str, redis.Redis]:
```
tezc, line 862: Is this leftover?

tests/integration/sandbox.py
tezc: I was just replying to your comment there :) Is that comment invalid now? Should we delete it? orshould we simplify this function?

tests/integration/sandbox.py
```diff
@@ -762,6 +770,216 @@ def restart(self):
...
+    def finish(self):
+        self.ended = True
+
+    def set_keys(self, keys: typing.List):
+        self.keys = keys
+
+    # might not want generate_ops, perhaps each multi should just be a single append/read of same key
+    def generate_ops(self, available_keys: typing.List[str]) -> typing.List[ElleOp]:
+        ops: typing.List[ElleOp] = []
+        for _ in range(random.randint(1, 1)):
```
tezc, line 903: Probably, related with the above. This is not a loop right? `random.randint(1,1)` will generate `1` only.

tests/integration/sandbox.py
```diff
@@ -762,6 +770,216 @@ def restart(self):
...
+                good_write = True
+                break
+            except ResponseError as e:
+                # handle MOVED/ASK as those aren't "failures"
+                m = re.search(r"^MOVED \d* (.*):(.*)$", str(e))
+                a = re.search(r"^ASK \d* (.*):(.*)$", str(e))
+                if m:
+                    client = redis.Redis(host=m.group(1), port=m.group(2))
+                elif a:
+                    # don't know how to handle submitting an ASKING yet + how to control it
```
tezc, line 956: Is this comment still valid? Looks like you handle asking already.

tests/integration/test_elle.py
```diff
@@ -0,0 +1,297 @@
...
+
+RedisRaft is licensed under the Redis Source Available License (RSAL).
+"""
+import time
+import pytest as pytest
+
+from .sandbox import ElleWorker, Elle, key_hash_slot
+
+
+@pytest.mark.skipif("not config.getoption('elle_threads')")
```
tezc, line 14:
```suggestion
```

We don't need this one right? You added `elle_test()` for this purpose?

tests/integration/test_elle.py
tezc: My point was we should be able to write tests unaware of Elle as we discussed. So, it means, tests should not do things like `key_hash_slot()` or `get_closest_marker()` etc. This is not the suggested way to write tests right? 

For certain tests, we may want to let Elle know the key prefix that is used in the test. So, it is okay to have a tag `@pytest.mark.key_hash_tag("{test}")`. 

Anyway, we can revisit this after exploring "one key per slot" setup. I think it should be the default. Then, if we see `@pytest.mark.key_hash_tag("{test}")` marker, Elle can operate with that slot / slots only.

tests/integration/test_elle.py
```diff
@@ -0,0 +1,297 @@
...
+@pytest.mark.elle_test()
+def test_elle_sanity(cluster_factory):
+    cluster1 = cluster_factory().create(3, raft_args={
+        'sharding': 'yes',
+        'external-sharding': 'yes'})
+    time.sleep(5)
+
+
+#@pytest.mark.skipif("not config.getoption('elle_threads')")
+#@pytest.mark.elle_test()
```
tezc, line 24: Please fix this mark. It requires `@pytest.mark.elle_test()` right?

tests/integration/sandbox.py
sjpotter: I agree that append/read is what we want.

tests/integration/test_elle.py
sjpotter: no, as this is the "manual" test, the mark are only needed for the background/automatic tests.  we might include the skipif though.

tests/integration/sandbox.py
sjpotter: yes, handle it now, need to remove

tests/integration/sandbox.py
sjpotter: so, my thought here was that it might be larger, as we might have multi/exec transactions with more than one append/read pair.  I'm not sure if it gains us anything.  If it doesn't, can remove the loop.

tests/integration/sandbox.py
sjpotter: yes

tests/integration/test_elle.py
sjpotter: i don't see a good way to get rid of key_hash_slot() in this test, I can get rid of the marker calls though.

the idea would be that the instead of doing a key_hash_slot() on the value we get from the marker, we do a hash_key_slot on a literal (which is also set as the marker).  Therefore, it be `slot = key_hash_slot("test")`.

the test itself would validate the that key(s) used by the test itself was/were migrated, and not rely on the cheating "count" mechanism, which would remove the need for the conditional at the bottom.

thoughts?

tests/integration/sandbox.py
tezc: Ok, it is up to you.

tests/integration/test_elle.py
tezc: Ok, `key_hash_slot()` is probably better/required. Just marker calls are a bit ugly inside a test. It'd be nice to avoid them.

tests/integration/test_elle.py
sjpotter: ok, the skipif should be there.  but not the mark

tests/integration/sandbox.py
sjpotter: I'm going to make it configurable, but will default to 1

sjpotter: and no need for random.  random doesn't give us anything here.  so will be configurable to be larger than 1 append/read pair, but will default to 1.

.github/workflows/daily.yml
sjpotter: this I dont quite understand.  need to sit down and dig in.

sjpotter: done

.github/workflows/daily.yml
```diff
@@ -19,6 +19,57 @@ on:
...
+    - uses: actions/checkout@v2
+      with:
+        repository: ${{ env.GITHUB_REPOSITORY }}
+        ref: ${{ env.GITHUB_HEAD_REF }}
+    - name: Install build dependencies
+      run: sudo apt-get install -y build-essential autoconf automake libtool cmake lcov leiningen default-jdk-headless
+    - name: Build
+      run: |
+        mkdir build && cd build
+        cmake .. -DPYTEST_OPTS="--redis-executable=redis/src/redis-server --elle-cli=elle-cli/target/elle-cli-0.1.4-standalone.jar --elle_threads=3 -v"
```
tezc, line 43:
```suggestion
        cmake .. -DPYTEST_OPTS="--redis-executable=redis/src/redis-server --elle-cli=elle-cli/target/elle-cli-0.1.4-standalone.jar --elle-threads=3 -v"
```

Please run daily with the above fix to verify we don't break the build and it actually runs `elle`. 

You can run on Github by using this branch. 
On your fork, `Actions` -> `Daily` -> `Run Workflow` -> `Use workflow from`: select your branch, then fill out the boxes. 

Not necessary right now but to verify we actually run Elle, we can count `success` response count and assert it is greater than zero. Also, counting `asking` and `moved` responses and printing to the log at the end similar to Jepsen would be nice.

tests/integration/test_elle.py
```diff
@@ -0,0 +1,298 @@
...
+
+@pytest.mark.elle_test()
+def test_elle_sanity(cluster_factory):
+    cluster1 = cluster_factory().create(3, raft_args={
+        'sharding': 'yes',
+        'external-sharding': 'yes'})
+    time.sleep(1)
+
+
+@pytest.mark.skipif("not config.getoption('elle_threads')")
```
tezc, line 22:
```suggestion
@pytest.mark.skipif("not config.getoption('elle-threads')")
```

tests/integration/sandbox.py
```diff
@@ -762,6 +770,203 @@ def restart(self):
...
+    def do_ops(self, ops: typing.List[ElleOp]):
+        # needs a way to handle asking
+        good_write = False
+        needs_asking = False
+
+        client = redis.Redis(host="localhost", port=self.clusters[0].leader_node().port)
+        self.elle.log_command(self.id, ops)
+
+        # 3 possible hops in normal scenario
+        # default MOVED -> ASK -> (remote) MOVED (not leader, so ASKING again) -> "remoe leader"
```
tezc, line 908:
```suggestion
        # default MOVED -> ASK -> (remote) MOVED (not leader, so ASKING again) -> "remote leader"
```

tests/integration/test_elle.py
```diff
@@ -0,0 +1,298 @@
...
+
+        cluster2_dbid,
+        1, 3,
+        slot, slot, 1, 0,
+        "{}00000001".format(cluster2_dbid).encode(), cluster2.node(1).address,
+        "{}00000002".format(cluster2_dbid).encode(), cluster2.node(2).address,
+        "{}00000003".format(cluster2_dbid).encode(), cluster2.node(3).address,
+    ) == b'OK'
+
+    # validate that cluster1 has no more eys in slot
```
tezc, line 283:
```suggestion
    # validate that cluster1 has no more entries in slot
```

tezc (approved):

sjpotter:
daily build with elle done on local repo showing it worked - https://github.com/sjpotter/redisraft/actions/runs/3459797904

PR #515, opened 2022-10-23
Handle ASKING in MULTI correctly

sjpotter:
reset only when we go to actually handle the command or upon errors in multi

bug found during elle development

### DIFF
--- a/src/multi.c
+++ b/src/multi.c
@@ -63,12 +63,14 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
         return true;
     } else if (cmd_len == 4 && !strncasecmp(cmd_str, "EXEC", 4)) {
         if (!multiState->active) {
+            setAskingState(rr, ctx, false);
             RedisModule_ReplyWithError(ctx, "ERR EXEC without MULTI");
             return true;
         }
 
         if (multiState->error) {
             MultiStateReset(multiState);
+            setAskingState(rr, ctx, false);
             RedisModule_ReplyWithError(ctx, "EXECABORT Transaction discarded because of previous errors.");
             return true;
         }
@@ -86,10 +88,12 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
     } else if (cmd_len == 7 && !strncasecmp(cmd_str, "DISCARD", 7)) {
         if (!multiState->active) {
             RedisModule_ReplyWithError(ctx, "ERR DISCARD without MULTI");
+            setAskingState(rr, ctx, false);
             return true;
         }
 
         MultiStateReset(multiState);
+        setAskingState(rr, ctx, false);
         RedisModule_ReplyWithSimpleString(ctx, "OK");
 
         return true;
--- a/src/multi.c
+++ b/src/multi.c
@@ -63,14 +63,12 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
         return true;
     } else if (cmd_len == 4 && !strncasecmp(cmd_str, "EXEC", 4)) {
         if (!multiState->active) {
-            setAskingState(rr, ctx, false);
             RedisModule_ReplyWithError(ctx, "ERR EXEC without MULTI");
             return true;
         }
 
         if (multiState->error) {
             MultiStateReset(multiState);
-            setAskingState(rr, ctx, false);
             RedisModule_ReplyWithError(ctx, "EXECABORT Transaction discarded because of previous errors.");
             return true;
         }
@@ -88,12 +86,10 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
     } else if (cmd_len == 7 && !strncasecmp(cmd_str, "DISCARD", 7)) {
         if (!multiState->active) {
             RedisModule_ReplyWithError(ctx, "ERR DISCARD without MULTI");
-            setAskingState(rr, ctx, false);
             return true;
         }
 
         MultiStateReset(multiState);
-        setAskingState(rr, ctx, false);
         RedisModule_ReplyWithSimpleString(ctx, "OK");
 
         return true;
--- a/src/multi.c
+++ b/src/multi.c
@@ -52,7 +52,6 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
             RedisModule_ReplyWithError(ctx, "ERR MULTI calls can not be nested");
         } else {
             multiState->active = 1;
-            multiState->asking = cmds->asking;
 
             /* We put the MULTI as the first command in the array, as we still
              * need to distinguish single-MULTI array from a single command.
@@ -81,7 +80,6 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
         /* Just swap our commands with the EXEC command and proceed. */
         RaftRedisCommandArrayFree(cmds);
         RaftRedisCommandArrayMove(cmds, &multiState->cmds);
-        cmds->asking = multiState->asking;
         MultiStateReset(multiState);
 
         return false;
--- a/src/multi.c
+++ b/src/multi.c
@@ -52,6 +52,7 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
             RedisModule_ReplyWithError(ctx, "ERR MULTI calls can not be nested");
         } else {
             multiState->active = 1;
+            multiState->asking = cmds->asking;
 
             /* We put the MULTI as the first command in the array, as we still
              * need to distinguish single-MULTI array from a single command.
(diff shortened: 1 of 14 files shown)


fadidahanna:
Good catch! can you add a test for the failure scenario?

src/multi.c
```diff
@@ -63,12 +63,14 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
         return true;
     } else if (cmd_len == 4 && !strncasecmp(cmd_str, "EXEC", 4)) {
         if (!multiState->active) {
+            setAskingState(rr, ctx, false);
             RedisModule_ReplyWithError(ctx, "ERR EXEC without MULTI");
             return true;
```
fadidahanna, line 68: I have the feeling we abuse the return value - in some cases, like in `EXEC` - false means - "cmds is ready and we can call`handleRedisCommandAppend()`.  in other cases - false means = "no MULTI-EXEC op, continue to handle the command in `handleRedisCommandAppend()`. True sometimes means Error, or sometimes expected behavior. Resetting the asking state is also not so clear - sometimes on errors we reset it, and sometimes not (like in the case of unsupported commands in the middle of MULTI-EXEC). 
Understand the logic, but it make it hard to follow this function. Suggest rewriting it to be more readable. 🙏

src/redisraft.c
```diff
@@ -731,6 +715,10 @@ static void handleRedisCommand(RedisRaftCtx *rr,
         return;
     }
 
+    /* update cmds array with the client's saved "asking" state */
+    cmds->asking = getAskingState(rr, ctx);
+    setAskingState(rr, ctx, false);
+
```
tezc, line 721: I see we added this line to handle multi case: [link](https://github.com/RedisLabs/redisraft/blame/bef32837e3209df7111a6173040f9071d20b1f1e/src/serialization.c#L52)

So, we copy "asking" state but looks like we override it later. 

ASKING sets client's `asking` =  `true`
MULTI sets `multiState->cmds->asking` = `true` 
GET X sets `multiState->cmds->asking` = `false`


Probably, that line above can be 
```c
target->asking |= source->asking;
```
Once we set it, we don't want to unset it. 
In future, we can improve this logic probably.

----
Btw, I didn't test but IIUC, to match Redis' behavior, we should do in this order:

```c
static void handleRedisCommand(RedisRaftCtx *rr, RedisModuleCtx *ctx, 
                               RaftRedisCommandArray *cmds)
{
    /* update cmds array with the client's saved "asking" state */
    cmds->asking = getAskingState(rr, ctx);
    setAskingState(rr, ctx, false);

    /* Check if MULTI/EXEC bundling is required. */
    if (MultiHandleCommand(rr, ctx, cmds)) {
        return;
    }

    /* Handle intercepted commands. We do this also on non-leader nodes or if we
     * don't have a leader, so it's up to the commands to check these conditions
     * if they have to.
     */
    if (handleInterceptedCommands(rr, ctx, cmds)) {
        return;
    }

    handleRedisCommandAppend(rr, ctx, cmds);
}
```

Redis handles "multi" before other commands, so, I guess we can do in this order if it doesn't cause any other problems.

src/multi.c
sjpotter: I don't see that.  there are only 2 return falses

1) on exec
2) when it has nothing to do with multi (i.e. we didn't do anything at all).  every other case is return true.

tests/integration/test_migrate.py
```diff
@@ -413,3 +413,57 @@ def test_redirect_asking_to_leader(cluster):
...
+    assert conn.read_response() == b'OK'
+    conn.send_command('MULTI')
+    assert conn.read_response() == b'OK'
+    conn.send_command('GET', 'key')
+    assert conn.read_response() == b'QUEUED'
+    conn.send_command('DISCARD')
+    assert conn.read_response() == b'OK'
+    conn.send_command('GET', 'key')
+    with raises(ResponseError, match='MOVED [0-9]+ 3.3.3.3:3333'):
+        conn.read_response()
```
tezc, line 469: Can we execute the command without using the connection object directly? 
It becomes single line instead of two, easier to follow imho. 

Btw in addition to your tests, I added a couple of more tests to be sure we clear "asking" flag after multi:

```suggestion
    node = cluster.leader_node()

    # 1- Verify 'asking' with 'multi'
    node.execute('ASKING')
    node.execute('MULTI')
    node.execute('GET', 'key')
    node.execute('GET', 'key')
    node.execute('GET', 'key')
    assert node.execute('EXEC') == [b'value', b'value', b'value']

    # 2- Verify 'EXEC' clears 'asking' flag
    node.execute('ASKING')
    node.execute('MULTI')
    node.execute('GET', 'key')
    assert node.execute('EXEC') == [b'value']

    with raises(ResponseError, match='MOVED [0-9]+ 3.3.3.3:3333'):
        node.execute('MULTI')
        node.execute('GET', 'key')
        node.execute('EXEC')

    with raises(ResponseError, match='MOVED [0-9]+ 3.3.3.3:3333'):
        node.execute('GET', 'key')

    # 3- Verify 'DISCARD' clears 'asking' flag
    node.execute('ASKING')
    node.execute('MULTI')
    node.execute('GET', 'key')
    node.execute('DISCARD')

    with raises(ResponseError, match='MOVED [0-9]+ 3.3.3.3:3333'):
        node.execute('MULTI')
        node.execute('GET', 'key')
        node.execute('EXEC')

    with raises(ResponseError, match='MOVED [0-9]+ 3.3.3.3:3333'):
        node.execute('GET', 'key')

```

tezc:
Probably, I wasn't clear but my suggestion was fixing `RaftRedisCommandArrayMove()`. We use this function on other places and we rely on moving `asking` flag correctly, so probably we should fix the issue inside `RaftRedisCommandArrayMove()`. For example,  this patch should fix the bug: (I can't create a suggestion here on github) 

```diff
diff --git a/src/serialization.c b/src/serialization.c
index 853957e40..19a4bd508 100644
--- a/src/serialization.c
+++ b/src/serialization.c
@@ -49,7 +49,7 @@ void RaftRedisCommandArrayMove(RaftRedisCommandArray *target, RaftRedisCommandAr
     source->len = 0;
     memset(source->commands, 0, sizeof(RaftRedisCommand *) * source->size);
 
-    target->asking = source->asking;
+    target->asking |= source->asking;
 }
 
 /* Free a RaftRedisCommand */
@@ -90,6 +90,8 @@ void RaftRedisCommandArrayFree(RaftRedisCommandArray *array)
     if (array->acl) {
         RedisModule_FreeString(NULL, array->acl);
     }
+
+    *array = (RaftRedisCommandArray) {0};
 }
 
 static size_t calcSerializedSize(RaftRedisCommand *cmd)
```

The bug is we clear `target->asking` on a subsequent `RaftRedisCommandArrayMove()` call. We can prevent it as in above.

I don't like this solution much but probably this is the simplest for now and we can improve all the logic later, currently our client/multi/asking/raftreq/commandarray handling is a bit messy. All these should be part of a `client` object, then our life will be easier imho.

tests/integration/test_migrate.py
sjpotter: I don't believe that works, as need to use the persistent connection to do these pipelined commands.

sjpotter:
so I thought of that, I'm just not sure it clear semantics from the function name.  i.e. the function name is just "Move the commands array", it has nothing to do with the metadata associated with it.  so I'm wondering if it should be handled separately (and yes, that would mean move shouldnt overwrite asking either now)

tests/integration/test_migrate.py
tezc: Pushed a small change hope that's ok. I guess we can do it in a single line here and use the same approach in other tests when required. Please change the tests if they are not correct.

tezc:
What is your suggestion? Rename? Another helper function like:

```c
void RaftRedisCommandArrayFullCopy(RaftRedisCommandArray *target, RaftRedisCommandArray *source)
{
    RaftRedisCommandArrayMove(target, source);
    
    target->asking = source->asking;
    target->acl = source->acl;
    
    source->asking = false;
    source->acl = NULL;
}
```
Then, we'll call `RaftRedisCommandArrayMove()` when we want to append. (while accumulating for multi) ? 

I don't have a strong opinion as I think we have to refactor all these in future. Just being pragmatic, looking for the smallest fix. 

My concerns: 
- Can we handle this without introducing `asking` flag elsewhere? 
- See the other usages of `RaftRedisCommandArrayMove()`. Currently, we rely on moving `asking` flag on these places. We should not break it.

sjpotter:
ok, I can be convinced.

tezc (approved):

PR #516, opened 2022-10-26
RR-100 adding snapshot info metrics

fadidahanna:


src/redisraft.c
```diff
@@ -1800,9 +1800,14 @@ static void handleInfo(RedisModuleInfoCtx *ctx, int for_crash_report)
     RedisModule_InfoAddFieldULongLong(ctx, "fsync_avg_microseconds", avg);
 
     RedisModule_InfoAddSection(ctx, "snapshot");
-    RedisModule_InfoAddFieldCString(ctx, "snapshot_in_progress", rr->snapshot_in_progress ? "yes" : "no");
-    RedisModule_InfoAddFieldULongLong(ctx, "snapshots_loaded", rr->snapshots_loaded);
+    RedisModule_InfoAddFieldCString(ctx, "snapshot_filename", rr->config.rdb_filename);
+    RedisModule_InfoAddFieldULongLong(ctx, "snapshot_last_idx", rr->raft ? raft_get_snapshot_last_idx(rr->raft) : 0);
+    RedisModule_InfoAddFieldULongLong(ctx, "snapshot_last_term", rr->raft ? raft_get_snapshot_last_term(rr->raft) : 0);
+    RedisModule_InfoAddFieldULongLong(ctx, "snapshot_size", rr->outgoing_snapshot_file.len);
```
tezc, line 1806: Just a note, I guess adding another line for human readable version would be nice. Looks like Redis is missing that API but we can push `RM_InfoAddBytes()` quickly (if there is no objection) , feels like it'd be useful to everybody.

src/redisraft.c
```diff
@@ -1800,9 +1800,14 @@ static void handleInfo(RedisModuleInfoCtx *ctx, int for_crash_report)
...
+    RedisModule_InfoAddFieldULongLong(ctx, "snapshot_time_secs", rr->last_snapshot_time);
```
tezc, line 1807: If I'm not mistaken, this is in `milliseconds` but we say `secs`.

tests/integration/test_snapshots.py
```diff
@@ -490,3 +490,120 @@ def test_snapshot_fork_failure(cluster):
...
+    # Make sure log is compacted
+    assert cluster.node(1).client.execute_command(
+        'RAFT.DEBUG', 'COMPACT') == b'OK'
+
+    info = cluster.node(1).info()
+    assert info['raft_snapshots_created'] == 1
+    assert info['raft_snapshot_last_idx'] == 12
+    assert info['raft_snapshot_last_term'] == 1
+    assert info['raft_snapshot_size'] != 0
+    assert info['raft_snapshot_time_secs'] != 0
```
tezc, line 583: As this metric currently prints `milliseconds` instead of `seconds`, this test passes (Snapshot takes more than a second). If you switch to printing `seconds`, this test might become flaky if snapshot takes less than a second on a fast machine :)

src/redisraft.c
fadidahanna: You're totally right - fixing

src/redisraft.c
fadidahanna: Agree with that - will suggest this API to OSS.

src/redisraft.h
```diff
@@ -848,7 +852,7 @@ RRStatus finalizeSnapshot(RedisRaftCtx *rr, SnapshotResult *sr);
 void cancelSnapshot(RedisRaftCtx *rr, SnapshotResult *sr);
 int pollSnapshotStatus(RedisRaftCtx *rr, SnapshotResult *sr);
 void configRaftFromSnapshotInfo(RedisRaftCtx *rr);
-int raftLoadSnapshot(raft_server_t *raft, void *udata, raft_index_t idx, raft_term_t term);
+int raftLoadSnapshot(raft_server_t *raft, void *udata, raft_term_t term, raft_index_t idx);
```
tezc, line 855: oops, this is my mistake.

tezc (approved):

PR #519, opened 2022-11-01
[RR-116] Do not load snapshot inside the command callback

tezc:
RedisRaft crashes while loading a big snapshot.

RedisRaft calls `rdbLoad()` function from Redis to load the received
snapshot file (RDB). This function is called inside a command callback.
`rdbLoad()` goes to networking and process events if loading the 
snapshot takes a while. If the current client that triggered command
callback disconnects, client struct can be freed before it returns
from `rdbLoad()` and from the callback. After that, a use-after-free
bug occurs and leads to crash.

Previously, there was a similar bug. Redis was trying to process
another message from the same client:
https://github.com/RedisLabs/redisraft/pull/450

Until this issue is fixed, as a workaround, we can delay loading the
snapshot and let the command callback return. Inside the timer callback,
we can load the snapshot. To do that, we need to set state to 
`REDIS_RAFT_LOADING`, so, other commands should be rejected until we
load the snapshot.

### DIFF
--- a/src/snapshot.c
+++ b/src/snapshot.c
@@ -525,35 +525,22 @@ void configRaftFromSnapshotInfo(RedisRaftCtx *rr)
     RedisModule_Assert(raft_get_num_nodes(rr->raft) == added + 1);
 }
 
-/* After a snapshot is received, load it into the Raft library:
- * 1. Replace received snapshot file with the current one.
- * 2. Load rdb file.
- * 3. Configure index/term/etc.
- * 4. Reconfigure nodes based on the snapshot metadata configuration.
- * 5. Create a new snapshot memory map.
- */
-int raftLoadSnapshot(raft_server_t *raft, void *user_data, raft_term_t term, raft_index_t index)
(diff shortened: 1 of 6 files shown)


src/snapshot.c
```diff
@@ -568,19 +551,74 @@ int raftLoadSnapshot(raft_server_t *raft, void *user_data, raft_term_t term, raf
...
+
+    rr->snapshot_load = (SnapshotLoad){
+        .pending = false,
+        .term = -1,
+        .index = -1,
+    };
+}
+
+/* After a snapshot is received, load it into the Raft library:
+ * 1. Replace received snapshot file with the current one.
```
fadidahanna, line 576: Since `raftLoadSnapshot` is not loading any more the snapshot, but only triggers a loading state of RR, I suggest changing the documentation here accordingly  + adding some documentation in `loadPendingSnapshot`

tezc: ok added some docs

fadidahanna (approved):

PR #522, opened 2022-11-02
Use SHUTDOWN command instead of exit()

tezc:
Calling `exit()` from modules is unexpected as Redis won't have a 
chance to do clean up: Closing the sockets, removing pid file etc.
Also, other modules won't get the shutdown callback.

Changing `exit()` to `SHUTDOWN` command to prevent these issues.

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -42,10 +42,6 @@ static void replaceShardGroups(RedisRaftCtx *rr, raft_entry_t *entry);
 
 void shutdownAfterRemoval(RedisRaftCtx *rr)
 {
-    size_t len;
-    const char *err;
-    RedisModuleCallReply *reply;
-
     LOG_NOTICE("*** NODE REMOVED, SHUTTING DOWN.");
 
     if (rr->config.log_filename) {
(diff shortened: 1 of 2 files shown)


fadidahanna:
So for that reason in order to shut down gracefully, we should keep `RAFT_ERR_SHUTDOWN` from libraft...

fadidahanna:
That was not the original behaviour - see `callRaftPeriodic()` and `handleBeforeSleep()`, but agree with your point. 
Besides that, maybe it's a good time to rename the function `shutdownAfterRemoval()`. We call it not only after node removal, see `cmdRaftNodeShutdown()`.

tezc:
It's a bit confusing but we are actually calling `shutdownAfterRemoval()` when we remove the node. `raft.nodeshutdown` will be sent from the leader to a follower to indicate it has been removed. The reason is when the leader appends removal entry of a follower, it stops replicating entries to that node immediately (In raft, config changes take effect on append, rather than apply).  So, follower will never receive its own removal entry. `raft.nodeshutdown` is a best-effort message from the leader to a removed follower to tell "you have been removed".

Still, sure we can rename the function or add some comments about it, open to suggestions.

tezc:
Added comment for `RAFT.NODESHUTDOWN`

fadidahanna (approved):

PR #528, opened 2022-11-05
[RR-117] Fix crash while resolving address

tezc:
RedisRaft does hostname resolution in another thread. If a node is 
removed from the cluster while the server is resolving its address,
its connection object might be freed. This causes use-after-free 
issue just after the address resolution is completed.

1- Connection will be marked for termination here: 
https://github.com/RedisLabs/redisraft/blob/4f0e28b8f4c37a06a7ef0a8e45595ee2fb514c26/src/raft.c#L979

2- Connection will be freed here:
https://github.com/RedisLabs/redisraft/blob/4f0e28b8f4c37a06a7ef0a8e45595ee2fb514c26/src/connection.c#L472

3- Then, it will crash inside `handleResolved()`:
https://github.com/RedisLabs/redisraft/blob/4f0e28b8f4c37a06a7ef0a8e45595ee2fb514c26/src/connection.c#L284

To solve the issue, added a line to prevent `ConnIsIdle()` return `true`
while resolving the address. It will prevent the deallocation. 
Connection handling requires a rewrite. This is just a temporary 
solution.

Introduced by https://github.com/RedisLabs/redisraft/pull/247

### DIFF
--- a/src/connection.c
+++ b/src/connection.c
@@ -419,6 +419,9 @@ void ConnMarkDisconnected(Connection *conn)
  */
 bool ConnIsIdle(Connection *conn)
 {
+    /* If the connection is in the "resolving" state, it means the hostname
+     * resolution is in progress in another thread. Even though it was marked as
+     * "terminating", the connection is not idle yet. */
     return conn->state == CONN_DISCONNECTED ||
            conn->state == CONN_CONNECT_ERROR ||
            (conn->state != CONN_RESOLVING && (conn->flags & CONN_TERMINATING));
--- a/src/connection.c
+++ b/src/connection.c
@@ -417,10 +417,11 @@ void ConnMarkDisconnected(Connection *conn)
 /* An idle state is one that will not transition automatically to another
  * state, unless actively mutated.
  */
-
 bool ConnIsIdle(Connection *conn)
 {
-    return (conn->state == CONN_DISCONNECTED || conn->state == CONN_CONNECT_ERROR || (conn->flags & CONN_TERMINATING));
+    return conn->state == CONN_DISCONNECTED ||
+           conn->state == CONN_CONNECT_ERROR ||
+           (conn->state != CONN_RESOLVING && (conn->flags & CONN_TERMINATING));
 }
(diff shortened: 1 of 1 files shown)


sjpotter:
just a note: I dont think its in "another thread", its on the event loop, but the point is that its still asynchronous so ordering isn't something we control.  Because they are all on the main thread, there are probably other ways of fixing it as well, but this is probably sufficient as is.

fadidahanna:
@tezc , thanks for the fix - please open a bug ticket, with a link to the relevant crash, and add this issue to a connection handling rewrite ticket (if we don't have any, please open one for monitoring all the known issues we currently have).

src/connection.c
```diff
@@ -417,10 +417,11 @@ void ConnMarkDisconnected(Connection *conn)
 /* An idle state is one that will not transition automatically to another
  * state, unless actively mutated.
  */
-
 bool ConnIsIdle(Connection *conn)
 {
-    return (conn->state == CONN_DISCONNECTED || conn->state == CONN_CONNECT_ERROR || (conn->flags & CONN_TERMINATING));
+    return conn->state == CONN_DISCONNECTED ||
+           conn->state == CONN_CONNECT_ERROR ||
+           (conn->state != CONN_RESOLVING && (conn->flags & CONN_TERMINATING));
```
fadidahanna, line 424: Suggest adding here an explanation for the need for `conn->state != CONN_RESOLVING`, and add a JIRA ticket number for a future fix.

tezc: Added a small comment.

tezc:
@sjpotter We do address resolution in another thread: [link](https://github.com/RedisLabs/redisraft/blob/4f0e28b8f4c37a06a7ef0a8e45595ee2fb514c26/src/connection.c#L401)
Hostname resolution might be slow. So, not to block the main thread, it is offloaded to another thread. 

@fadidahanna This becomes visible while testing big snapshots. Associated with the existing ticket.

fadidahanna (approved):

PR #529, opened 2022-11-12
Enforce PEP8 style for Python tests

tezc:
Enforce PEP8 style for Python tests
- Added CI checker.  To run the command on your local:
  ` pycodestyle --show-source tests` 

There were previous attempts like: #364 to cleanup linter warnings.
Back then, I didn't introduce CI style checker for Python tests as 
it might be limiting too much. The problem is when there are lots
of warnings, it is easy to miss serious ones. 

Test fixes are mostly regarding line length limit. There is no logical
change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,4 +100,3 @@ jobs:
       - name: Spellcheck
         run:
           codespell --config=./.github/codespell/config.txt
-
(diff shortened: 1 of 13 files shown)


tezc:
I find the lines longer than 80 chars quite unreadable (generally due to deep nested code or verbose variable & function names) but normally I'm okay with having longer lines if it increases the readability significantly. For example, here: [link](https://github.com/RedisLabs/redisraft/blob/83a76b2cf1f41b83389c6a518c197cb1663a0bc0/src/config.c#L724), I thought longer lines would make it better. So, these are often exceptional cases. That's also why I didn't suggest enforcing this rule at first but considering the fixes here and previous PRs, I don't think longer lines in the project are more readable. (at least most of them). 

Let's try the alternative for now. In Python, it should be easier to follow this rule anyway. (compared to other languages, C etc.)

fadidahanna (approved):

PR #532, opened 2022-11-15
[RR-113] simple test for non quorum reads with scripts

sjpotter:
2 node cluster, set value, pause follower node, script returns value

### DIFF
--- a/tests/integration/test_sanity.py
+++ b/tests/integration/test_sanity.py
@@ -215,6 +215,23 @@ def test_nonquorum_reads(cluster):
     cluster.node(2).client.get('x')
 
 
+def test_nonquorum_read_scripts(cluster):
+    """
+    Test non-quorum reads with scripts
+    """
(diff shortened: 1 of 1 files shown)


tezc (approved):

tezc:
@sjpotter LGTM but is this PR complete or is it going to be only PR for readonly scripts? If so, we should add tests for `evalsha_ro` and `fcall_ro` as well. Also, maybe we should add some write commands inside these scripts and verify it fails. Maybe we can look at Redis tests that may give ideas about what to test. Also, there is ACL + maxmemory. Combination goes big easily as we talked :)

sjpotter:
I added fcall_ro and evalsha_ro, I dont think we need to validate write calls failing (at least as part of this 2 pointer) as that's testing the underlying redis functionality, not redisraft.

tezc (approved):

Issue #367, opened 2022-06-06
Improve RAFT prefix mechanism


yossigo:
Right now, `RAFT` is prefixed to all intercepted commands.
It is defined as a write command which is wrong from the Redis perspective.

Proposal:
* Introduce additional `RAFT` variants (e.g. read-only, unsupported command, etc.)
* Allows read/write to be reflected correctly to Redis
* Optimization: saves the need to lookup the command twice (1st in filter, 2nd in module command)

fadidahanna:
JIRA ticket: https://redislabs.atlassian.net/browse/RR-127

PR #536, opened 2022-11-17
[RR-111] test to ensure followers in oom still apply log changes

sjpotter:
create a 3 node cluster, put followers into oom, sets should still apply and be visible on all nodes

### DIFF
--- a/tests/integration/test_sanity.py
+++ b/tests/integration/test_sanity.py
@@ -654,7 +654,14 @@ def test_followers_in_oom(cluster):
     cluster.wait_for_unanimity()
 
     assert cluster.execute("get", "abc") == b'1234'
-    node2 = cluster.node(2)
-    node3 = cluster.node(3)
-    assert node2.execute("raft.debug", "exec", "get", "abc") == b'1234'
-    assert node3.execute("raft.debug", "exec", "get", "abc") == b'1234'
+    assert cluster.node(2).raft_debug_exec("get", "abc") == b'1234'
+    assert cluster.node(3).raft_debug_exec("get", "abc") == b'1234'
+
+    cluster.execute('EVAL', """
+        redis.call('SET','abc', 5678);
+        return 1234;""", '0')
+    cluster.wait_for_unanimity()
+
+    assert cluster.execute("get", "abc") == b'5678'
+    assert cluster.node(2).raft_debug_exec("get", "abc") == b'5678'
+    assert cluster.node(3).raft_debug_exec("get", "abc") == b'5678'
--- a/tests/integration/test_sanity.py
+++ b/tests/integration/test_sanity.py
@@ -642,3 +642,17 @@ def test_metadata_restore_after_restart(cluster):
     # n2 triggered an election, so n2 should have voted for itself.
     assert n2.info()['raft_current_term'] == 2
     assert n2.info()['raft_voted_for'] == 2
+
+
+def test_followers_in_oom(cluster):
+    cluster.create(3)
+
+    cluster.node(2).execute('config', 'set', 'maxmemory', 1)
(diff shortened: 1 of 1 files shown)


tezc:
We can also test snapshot delivery on OOM. Also, maybe another test for script execution on OOM is needed? Wdyt?

tests/integration/test_sanity.py
```diff
@@ -642,3 +642,19 @@ def test_metadata_restore_after_restart(cluster):
...
+    cluster.node(3).execute('config', 'set', 'maxmemory', 1)
+
+    cluster.execute("set", "abc", 1234)
+    cluster.wait_for_unanimity()
+
+    assert cluster.execute("get", "abc") == b'1234'
+    node2 = cluster.node(2)
+    node3 = cluster.node(3)
+    assert node2.execute("raft.debug", "exec", "get", "abc") == b'1234'
+    assert node3.execute("raft.debug", "exec", "get", "abc") == b'1234'
```
tezc, line 660:
```suggestion
    assert cluster.node(2).raft_debug_exec("get", "abc") == b'1234'
    assert cluster.node(3).raft_debug_exec("get", "abc") == b'1234'
```

Just an alternative to make linter happy. If needed, I guess even we can rename `raft_debug_exec()` -> `debug_exec()`

sjpotter:
> We can also test snapshot delivery on OOM. Also, maybe another test for script execution on OOM is needed? Wdyt?

added a script execution one, unsure snapshot delivery adds anything, as we've already tested all paths through apply.

tezc:
I remember one thing, on OOM, Redis used to reject `raft.ae` as we register it as `write` command: 
https://github.com/RedisLabs/redisraft/blob/8a367d19f2afa5baf08d1edf4c6a99856fa61ebe/src/redisraft.c#L1865

See the issue for it: https://github.com/RedisLabs/redisraft/issues/367

So, looks like this is not happening, not sure why. It deserves an investigation itself probably but my point is there might be some surprises on snapshot delivery as well. e.g. Later, we register `raft.ae` properly and forget `raft.snapshot` to register with correct flags. Then, snapshot delivery will fail on OOM. 

Overall, I guess it'd be nice to have verify for cluster is operational on OOM. Election is happening, leader transfer works, snapshot delivery works, we can add and remove nodes (as there are different messages). We can add a single test for these if you want, either as part of this PR or we can open a ticket to add it later.

sjpotter:
so you're saying you're suprised the test even worked, because you'd have expected it to fail when sending the AE RPC to the followers?

tezc:
Okay, I checked, `raft` and `raft.ae` will not be rejected as their flag is `write`, not `deny-oom`.

Looks like I remembered this one: https://github.com/RedisLabs/redisraft/blob/8a367d19f2afa5baf08d1edf4c6a99856fa61ebe/tests/redis-suite/skip.txt#L76

It used to hit this line if I remember correctly:
https://github.com/redis/redis/blob/203b12e41ff7981f0fae5b23819f072d61594813/src/server.c#L3873

So, it was rejecting some of our commands.

In summary, probably a single test to verify cluster is operational on OOM might protect us from feature changes in Redis.

sjpotter:
> In summary, probably a single test to verify cluster is operational on OOM might protect us from feature changes in Redis.

I'd argue that this test effectively does that.  by putting all followers into OOM

tezc:
> Overall, I guess it'd be nice to have verify for cluster is operational on OOM. Election is happening, leader transfer works, snapshot delivery works, we can add and remove nodes (as there are different messages).

I meant these by saying "cluster is operational" but not necessary as part of this PR anyway. We can add it later if required.

tezc (approved):

PR #538, opened 2022-11-21
Add test for cluster operations on OOM

tezc:
Add tests to verify cluster operations are allowed on OOM. 

Tests do following operations:
- Migrate keys 
- Deliver snapshots
- Add/remove nodes
- Transfer leadership

### DIFF
--- a/tests/integration/test_migrate.py
+++ b/tests/integration/test_migrate.py
@@ -132,6 +132,140 @@ def test_migration_basic(cluster_factory):
         cluster2.node(1).execute('get', '{key2}1')
 
 
+def test_migration_basic_on_oom(cluster_factory):
+    """
+    Simulates full migration process while cluster is OOM. Test generates keys
+    in a cluster and moves half of the keys to another cluster.
+    """
+
+    cluster1 = cluster_factory().create(3, raft_args={
+        'sharding': 'yes',
+        'external-sharding': 'yes'})
+    cluster2 = cluster_factory().create(3, raft_args={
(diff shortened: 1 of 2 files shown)


tezc:
@sjpotter Let's merge this as this is just a copy of existing tests with `maxmemory=1`. We'll know `raft.xx` commands work on OOM.

tests/integration/test_migrate.py
```diff
@@ -132,6 +132,140 @@ def test_migration_basic(cluster_factory):
...
+    assert cluster1.execute('dbsize') == key_count
+    assert cluster1.node(1).execute('get', '{key2}1') == b'value'
+    with raises(ResponseError, match='MOVED 12539'):
+        cluster1.node(1).execute('get', '{key}1')
+
+    # cluster-2 has slots 8001-16383
+    assert cluster2.execute('dbsize') == key_count
+    assert cluster2.node(1).execute('get', '{key}1') == b'value'
+    with raises(ResponseError, match='MOVED 4998'):
+        cluster2.node(1).execute('get', '{key2}1')
```
sjpotter, line 266: wondering if validation should validate that non of the migrated keys are on cluster-1 and they are all on cluster-2

ala the loop above

    for i in range(key_count):
        cluster1.execute('set', '{key}' + str(i), 'value')   # slot 12539

here we are just testing a single key.

tezc: Ok, added validation for all the keys.

PR #539, opened 2022-11-23
[RR-122] add crc value to every entry we write to the log

sjpotter:
+ validate entry crc against stored value on Log Load at startup.
+ update RaftEntry python test class to know about the new element


src/log.c
```diff
@@ -234,29 +186,46 @@ static int writeEntry(Log *log, raft_entry_t *ety)
     return RR_ERROR;
 }
 
-static int writeHeader(Log *log)
+static raft_entry_t *readEntry(Log *log)
```
fadidahanna, line 189: `writeEntry()` also write the crc of the entry, therefore, `readEntry()` should also read the crc value

src/log.c
```diff
@@ -375,7 +352,28 @@ int LogLoadEntries(Log *log)
 
             return RR_OK;
         }
+        /* validate crc for entry by reading and rebuilding it */
+        long read_crc;
```
fadidahanna, line 356: Suggest adding a dedicated function for entry validation.

src/log.c
```diff
@@ -478,6 +476,33 @@ int LogDelete(Log *log, raft_index_t from_idx)
             PANIC("ftruncate failed: %s", strerror(errno));
         }
 
+        /* update running crc value */
+        /* if, we truncated entire log file, its the crc value of the header
```
fadidahanna, line 480:
```suggestion
        /* if we truncated the entire log file, it's the crc value of the header
```

src/log.c
```diff
@@ -206,24 +145,37 @@ static int writeEntry(Log *log, raft_entry_t *ety)
...
+    /* accumulating crc, so add the crc value we just wrote as well */
+    crc = crc16_ccitt(crc, buf, off);
+
     ssize_t ret = FileWrite(&log->idxfile, &offset, sizeof(offset));
     if (ret != sizeof(offset)) {
         goto error;
     }
 
+    /* everything succeeded, so can update accumulated crc */
+    log->current_crc = crc;
```
fadidahanna, line 177: Where do you keep `current_crc` - the crc of the whole log file?

src/snapshot.c
```diff
@@ -271,13 +271,13 @@ RRStatus finalizeSnapshot(RedisRaftCtx *rr, SnapshotResult *sr)
     /* Rewrite any additional log entries beyond the snapshot to a new
      * log file.
      */
-    num_log_entries = LogRewrite(rr, temp_log_filename,
-                                 rr->curr_snapshot_last_idx,
-                                 rr->curr_snapshot_last_term);
+    new_log = LogRewrite(rr, temp_log_filename,
+                         rr->curr_snapshot_last_idx,
+                         rr->curr_snapshot_last_term,
+                         &num_log_entries);
```
fadidahanna, line 277: @tezc, what was the reason for calling to 'LogOpen` here? only for validation? can we omit this check?

tests/integration/test_log.py
```diff
@@ -258,3 +258,31 @@ def test_log_partial_entry(cluster):
     # There will be 9 valid entries + 1 NOOP entry = 10
     assert cluster.node(1).info()['raft_log_entries'] == 10
     assert cluster.execute('get', 'x') == ('b' * 15000).encode('utf-8')
+
+
+def test_log_corrupt_entry(cluster):
```
fadidahanna, line 263: What about corruption in the middle of the file (not the in last entry/ header) or garbage "valid" last entry?

src/snapshot.c
tezc: Not sure about the original intention but we can skip it. It looks totally fine to me.

src/log.c
```diff
@@ -353,11 +322,19 @@ int LogReset(Log *log, raft_index_t index, raft_term_t term)
 int LogLoadEntries(Log *log)
 {
     log->num_entries = 0;
+    long calc_crc = 0;
 
     if (readHeader(log) != RR_OK) {
         return RR_ERROR;
     }
 
+    /* calculate crc to header end */
```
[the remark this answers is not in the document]
sjpotter: so, I'm not sure of this.  

1) readHeader (for some reasn) might be called in other situations, much like readEntry()
2) since readEntry can't update it itself (because it can be called randomly), I'm not sure readHeader() should, even if it could.

I'm unconvinced readHeader() should even do the validation itself, but it was the cleanest way to get it into the code, without modifying other locations.

src/log.c
sjpotter: its extra IO that seems unneccessary (probably not the biggest deal, but didn't serve much point.  we validate it on startup, our use case isn't it getting corrupted during a single runtime (if that would happen, we have worse issues).

src/log.c
sjpotter: ok, can look at that.

src/log.c
sjpotter: we don't, as discussed, I don't think it adds value, but still open to discussion.  It's very possible that "current_crc" we have is overkill and no need to really include the data of the CRC we write to the file, just a chain of the entries.  (so it wont correspond to the file, will just be in a sense a "block chain".)

src/snapshot.c
sjpotter: i.e. fine with what I changed?

tests/integration/test_log.py
sjpotter: I can add a test, but it will behave the same, prune it to that point.

src/log.c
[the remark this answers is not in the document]
fadidahanna: I think we should treat the crc and the entry data as a single entity. I wonder about the validation - I think we can have it optional in the read itself, or have a new RR entry type that contains also the crc, and only read it in the read func, but do a validation after, only when needed. WDYT?

src/log.c
fadidahanna: Yes, after a discussion with Yossi I agree with that.

src/snapshot.c
fadidahanna: @tezc

tests/integration/test_log.py
fadidahanna: right, but it will test that indeed we prune all the irrelevant entries

src/log.c
```diff
@@ -177,24 +113,27 @@ static int readHeader(Log *log)
     log->snapshot_last_idx = snapshot_last_index;
     log->index = snapshot_last_index;
 
+    /* validate CRC, by rebuilding header */
```
fadidahanna, line 116: Suggest doing a validation separately, as in readEntry()

src/log.c
```diff
@@ -206,24 +145,42 @@ static int writeEntry(Log *log, raft_entry_t *ety)
...
 
+    /* crc */
+    /* accumulating crc, so start with current crc */
+    long crc = crc16_ccitt(log->current_crc, buf, off);
+    crc = crc16_ccitt(crc, ety->data, ety->data_len);
+    crc = crc16_ccitt(crc, "\r\n", 2);
+
+    /* write crc as added element */
+    off = 0;
+    off += writeLong(buf + off, sizeof(buf) - off, crc);
```
tezc, line 168:
```suggestion
    off = writeLong(buf, sizeof(buf), crc);
```

Can we simplify a bit like this?

src/log.c
```diff
@@ -234,29 +191,57 @@ static int writeEntry(Log *log, raft_entry_t *ety)
...
+    if (FileRead(&log->file, e->data, length) != length ||
+        FileRead(&log->file, crlf, 2) != 2) {
+        raft_entry_release(e);
+        return NULL;
+    }
+
+    /* crc */
+    long crc;
+    if (!readLong(&log->file, &crc)) {
+        return NULL;
```
tezc, line 234: entry will leak here.

src/log.c
```diff
@@ -478,6 +504,33 @@ int LogDelete(Log *log, raft_index_t from_idx)
...
+            log->current_crc = crc16_ccitt(0, buf, off);
+        } else { /* log entry */
+            /* to regenerate running crc value, need to reread last entry */
+            seekEntry(log, from_idx - 1);
+            raft_entry_t *e = readEntry(log, NULL);
+            raft_entry_release(e);
+
+            /* running crc to the end of this entry */
+            long crc;
+            readLong(&log->file, &crc);
```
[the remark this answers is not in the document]
fadidahanna: Good catch! we definitely need to add test for that flow

src/log.c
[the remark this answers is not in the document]
sjpotter: ok, changed

src/log.c
sjpotter: is that also about simplifying the crc and not recalculating it with the "crc" element.  i.e. just a chain of the data as mentioned?

src/log.c
sjpotter: what do you mean?

src/log.c
```diff
@@ -478,6 +504,33 @@ int LogDelete(Log *log, raft_index_t from_idx)
             PANIC("ftruncate failed: %s", strerror(errno));
         }
 
+        /* update running crc value */
+        /* if we truncated the entire log file, it's the crc value of the header
+         * otherwise, it's the crc from the last entry
+         */
+        if (from_idx - 1 == log->snapshot_last_idx) { /* header */
```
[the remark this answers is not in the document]
sjpotter: ok

src/log.c
sjpotter: ok, the reason I did it the same was to enable copy pasting into other places less error prone, but changed it.

src/log.c
sjpotter: ok, I think I do that now

src/log.c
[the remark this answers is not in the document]
tezc: As this is part of `LogDelete()`, maybe we need to trigger `pop` of log entries and then do a restart. Then, we should verify we don't truncate logs due to miscalculated CRC.

tests/integration/test_log.py
sjpotter: modified the test to add a test case for this

src/log.c
```diff
@@ -284,18 +336,51 @@ int LogReset(Log *log, raft_index_t index, raft_term_t term)
...
+        return RR_ERROR;
+    }
+
+    /* update running crc */
+    len = writeLong(buf, sizeof(buf), calc_crc);
+    log->current_crc = crc16_ccitt(calc_crc, buf, len);
+
     /* Read Entries */
     while (true) {
+        long read_crc;
```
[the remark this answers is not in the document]
sjpotter: I'm not sure it makes a difference, its not reused.  If we go to what I've pondered of just chaining blocks without recalculating based on the element that is the crc, then its even clearer.

CMakeLists.txt
```diff
@@ -182,6 +182,7 @@ add_library(redisraft MODULE
         src/fsync.c
         src/join.c
         src/log.c
+        src/log_utils.c
```
tezc, line 185: We don't need this file anymore (and also log_utils.h).

sjpotter: see a booboo, missed one thing.  fixing

src/log.c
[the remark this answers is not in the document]
tezc: Add test pop + multi restarts

tests/integration/test_log.py
```diff
@@ -258,3 +258,31 @@ def test_log_partial_entry(cluster):
...
+    assert cluster.node(1).info()['raft_log_entries'] == 5
+    cluster.node(1).kill()
+
+    filename = cluster.node(1).raftlog
+    os.open(filename, os.O_RDWR)
+
+    # ----------------------------------------------------
+    # modify CRC32 for last entry
+    file = open(filename, "r+")
+    file.seek(os.path.getsize(filename) - 4)
```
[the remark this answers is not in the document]
tezc: Discussed on meeting to add this for log header + a log entry

tests/integration/test_log.py
```diff
@@ -85,6 +88,21 @@ def test_log_rollback(cluster):
...
+
+    # kill all nodes (1 first, as we don't want any additional log entries)
+    cluster.node(1).terminate()
+    cluster.node(2).terminate()
+    cluster.node(3).terminate()
+
+    # restart 1, won't have an active cluster, but should have loaded log
+    cluster.node(1).start()
+    # need to give it time to read log
+    time.sleep(1)
```
tezc, line 103: It can be flaky on a slow build, maybe we can do `cluster.node(1).wait_for_election()`

tests/integration/raftlog.py
```diff
@@ -32,11 +33,14 @@ def from_file(cls, _file):
         elements = int(mb_line[1:])
 
         args = []
+        locations = []
```
tezc, line 36: This is really good!

tests/integration/test_log.py
sjpotter: by definition in this case, there wont be one (other 2 nodes are down) that's why I did it this way.  I was trying to figure out a different way to do it (i.e. wait on something that says node is up)

tests/integration/raftlog.py
sjpotter: note, its just the "data" part of the multiblk element, not the start of the element. so can only overwrite up to length of element.

tests/integration/test_log.py
tezc: How about `cluster.node(1).wait_for_info_param('raft_state', 'up')`

sjpotter: done

src/log.c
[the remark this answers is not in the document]
sjpotter: done now as part of test_log_rollback

tezc:
@sjpotter Is it ready? Do you want to add something more? 
If not, please trigger a few dailys and a Jepsen test for this branch, then we can merge this PR? Wdyt?

sjpotter:
ok, kicked them off

fadidahanna:
We mentioned in our meeting today that we want to have integration with the CRC32 code, and use it instead of CRC16 - do you plan to do it in this PR or in a separate one?

sjpotter:
done, running it through make test now (a bit more involved as required unsigned char * everywhere, not char *)

deps/common/crc16.c
```diff
@@ -63,11 +63,11 @@ static const unsigned short crc16tab[256]= {
 	0xef1f,0xff3e,0xcf5d,0xdf7c,0xaf9b,0xbfba,0x8fd9,0x9ff8,
 	0x6e17,0x7e36,0x4e55,0x5e74,0x2e93,0x3eb2,0x0ed1,0x1ef0
 };
-  
-unsigned short crc16_ccitt(const char *buf, int len)
+
+unsigned short crc16_ccitt(unsigned short crc_start, const char *buf, int len)
```
tezc, line 67: this change is not necessary anymore, let's revert it

sjpotter: are you sure?  I think its more useful this way, even if not needed for our use case.  (also requries changing all the callers)

tezc: It is just to avoid a commit to open source code. Don't know if we should care or not but my instincts say that if we don't need, we shouldn't touch it :) Probably, we should discuss it.

tezc (approved):

PR #542, opened 2022-11-28
[RR-45] Log and metadata file clean up

tezc:
Changes in this PR:
- Reuse serialization code between the log and the metadata files.
  Added common functions to `util.c`.
- Added `dbid` and `node_id` to the metadata file. 
  These fields are useful for detecting user error on a file mismatch.
  On start-up,  we verify dbid / node id of the metadata file, 
  snapshot file and log file match. 
- Added `MetadataArchive()`
  Similar to the snapshot and the log file, when the node is removed, 
  it will archive metadata file as well. 
- Replaced some `rr->log->dbid` usages with `rr->meta.dbid`.
  In an upcoming PR, we'll move to a multi-logfile scheme. To prevent
  some extra checks, replaced this usage with metadata access as we
  will always have a single metadata instance. 

PR can be reviewed commit by commit.

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -662,7 +662,7 @@ void ShardingInfoRDBLoad(RedisModuleIO *rdb)
         RedisModule_Assert(sg != NULL);
 
         /* set local flag */
-        if (!strncmp(sg->id, rr->log->dbid, RAFT_DBID_LEN)) {
+        if (!strncmp(sg->id, rr->meta.dbid, RAFT_DBID_LEN)) {
             sg->local = true;
         }
 
@@ -1200,7 +1200,7 @@ static int addClusterSlotNodeReply(RedisRaftCtx *rr, RedisModuleCtx *ctx, raft_n
     RedisModule_ReplyWithCString(ctx, addr->host);
     RedisModule_ReplyWithLongLong(ctx, addr->port);
 
-    raftNodeToString(node_id, rr->log->dbid, raft_node);
+    raftNodeToString(node_id, rr->meta.dbid, raft_node);
     RedisModule_ReplyWithCString(ctx, node_id);
 
     return 1;
@@ -1358,11 +1358,11 @@ static void addClusterNodeReplyFromNode(RedisRaftCtx *rr,
     int pong_recv = 0;
 
     char node_id[RAFT_SHARDGROUP_NODEID_LEN + 1];
-    raftNodeToString(node_id, rr->log->dbid, raft_node);
+    raftNodeToString(node_id, rr->meta.dbid, raft_node);
 
     char master_node_id[RAFT_SHARDGROUP_NODEID_LEN + 1];
     if (!leader) {
-        raftNodeIdToString(master_node_id, rr->log->dbid, raft_get_leader_id(rr->raft));
+        raftNodeIdToString(master_node_id, rr->meta.dbid, raft_get_leader_id(rr->raft));
         master = master_node_id;
     }
 
@@ -1580,7 +1580,7 @@ static int addClusterShardsLocalNodeReply(RedisRaftCtx *rr, RedisModuleCtx *ctx,
     }
 
     char node_id[RAFT_SHARDGROUP_NODEID_LEN + 1];
-    raftNodeToString(node_id, rr->log->dbid, raft_node);
+    raftNodeToString(node_id, rr->meta.dbid, raft_node);
 
     addClusterShardsNodeReply(rr, ctx, node_id, addr->port, addr->host, role);
 
@@ -1815,7 +1815,7 @@ void ShardGroupLink(RedisRaftCtx *rr,
 
 void ShardGroupGet(RedisRaftCtx *rr, RedisModuleCtx *ctx)
 {
-    ShardGroup *sg = GetShardGroupById(rr, rr->log->dbid);
+    ShardGroup *sg = GetShardGroupById(rr, rr->meta.dbid);
 
     /* 2 arrays
      * 1. slot ranges -> each element is a 3 element array start/end/type
@@ -1857,10 +1857,10 @@ void ShardGroupGet(RedisRaftCtx *rr, RedisModuleCtx *ctx)
         node_count++;
         RedisModule_ReplyWithArray(ctx, 2);
 
-        char node_id[RAFT_SHARDGROUP_NODEID_LEN + 1];
+        char node_id[128];
         raft_node_id_t id = raft_node_get_id(raft_node);
 
-        snprintf(node_id, sizeof(node_id), "%s%08x", rr->log->dbid, id);
+        snprintf(node_id, sizeof(node_id), "%s%08x", rr->meta.dbid, id);
         RedisModule_ReplyWithStringBuffer(ctx, node_id, strlen(node_id));
 
         char addrstr[512];
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1857,10 +1857,8 @@ void ShardGroupGet(RedisRaftCtx *rr, RedisModuleCtx *ctx)
         node_count++;
         RedisModule_ReplyWithArray(ctx, 2);
 
-        char node_id[RAFT_SHARDGROUP_NODEID_LEN * 2];
-        raft_node_id_t id = raft_node_get_id(raft_node);
-
-        snprintf(node_id, sizeof(node_id), "%s%08x", rr->meta.dbid, id);
+        char node_id[RAFT_SHARDGROUP_NODEID_LEN + 1];
+        raftNodeToString(node_id, rr->meta.dbid, raft_node);
         RedisModule_ReplyWithStringBuffer(ctx, node_id, strlen(node_id));
 
         char addrstr[512];
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1857,7 +1857,7 @@ void ShardGroupGet(RedisRaftCtx *rr, RedisModuleCtx *ctx)
         node_count++;
         RedisModule_ReplyWithArray(ctx, 2);
 
-        char node_id[128];
+        char node_id[RAFT_SHARDGROUP_NODEID_LEN * 2];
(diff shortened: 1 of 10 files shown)


src/cluster.c
```diff
@@ -1857,10 +1857,10 @@ void ShardGroupGet(RedisRaftCtx *rr, RedisModuleCtx *ctx)
         node_count++;
         RedisModule_ReplyWithArray(ctx, 2);
 
-        char node_id[RAFT_SHARDGROUP_NODEID_LEN + 1];
+        char node_id[128];
```
sjpotter, line 1860: why are we going to a magic number here?

src/cluster.c
```diff
@@ -662,7 +662,7 @@ void ShardingInfoRDBLoad(RedisModuleIO *rdb)
         RedisModule_Assert(sg != NULL);
 
         /* set local flag */
-        if (!strncmp(sg->id, rr->log->dbid, RAFT_DBID_LEN)) {
+        if (!strncmp(sg->id, rr->meta.dbid, RAFT_DBID_LEN)) {
```
sjpotter, line 665: I'm confused by adding node/dbid and the like into Metadata when it still needs to be in log for log writing (see writeHeader())

sjpotter: ok, I see, the point is, in the future we might have multiple log files we are juggling, so this makes a single place to find it.  With that said, why would it matter which log object we get it from?  should juggling really matter?

tezc: > I'm confused by adding node/dbid and the like into Metadata when it still needs to be in log for log writing

Added `dbid` and `nodeid` to metadata to verify files (metadata, log and snapshot) match on restart.

Btw, normally, `dbid` and `nodeid` don't have to be in the log header. Previously, metadata was part of the log header. When I moved metadata out of the log file, I didn't delete `dbid` and `nodeid` from the log file. Just keeping `dbid` and `nodeid` in both files to verify we don't mix files of another node.

> why would it matter which log object we get it from?

It doesn't matter which log object we get it. If we leave this unchanged, it still works. 

When we restart, as the first thing, we read metadata file. From that point, we have `dbid` there. So, I'm just trying to centralize `dbid` access, move it to metadata object. If we do that, we don't need to open log file just after we read metadata file: https://github.com/RedisLabs/redisraft/blob/a7dca73ce58997a1176f1944aac31cff3b2a65c3/src/redisraft.c#L1980

Currently, we are opening log file(without reading entries) just to initialize `rr->log->dbid`.

sjpotter: I'm trying to understand the motivation for moving vote and term to this file.

lets forget vote for a moment. what happens if we don't record term.

we are in term X, which should imply we have a log record that says we are in term X, so by replaying log, we will know that.

so I'm thinking it has to do with voter, and not simply what term we are in.  Is it what term we last voted in so that if we restart, we don't vote for a different node in same term?

src/cluster.c
tezc: Compiler generates a false positive warning on my local that buffer size might not be enough. Changed it to `RAFT_SHARDGROUP_NODEID_LEN * 2`, I guess it is better.

src/cluster.c
tezc: > we are in term X, which should imply we have a log record that says we are in term X, so by replaying log, we will know that.

If you are a follower, you may not have an entry with term X. e.g. you restart before receiving the noop entry. Now, this follower will accept RPCs with lower term. Not sure if Raft handles this well on corner cases. 

Just thinking loud, on a split brain, that follower can receive a lower term entry from leader A and add it to its log. Leader A can commit the entry. The leader with the new term, leader B, can override the entry later as it is trying to send noop entry with a higher term for the same index? 

> so I'm thinking it has to do with voter, and not simply what term we are in. Is it what term we last voted in so that if we restart, we don't vote for a different node in same term?

Yes, `vote` is here to prevent double vote on restart.

sjpotter: if we are a follower, didn't we by definition get a noop for term X?  isn't that what makes us a follower?  how would we become a follower without it?  voting makes us a follower?  or does voting make us leaderless on new term (not really a follower, right?) until we get the noop?  so we would reject incoming messages from old leader, as old term?

tezc: When you vote for a node, as you need to increment your term, you are not a follower of the previous leader anymore. 
Now, until you receive NOOP, say whatever you like :) You can say the node is a follower or leaderless. It doesn't matter for Raft as long as you reject older term RPCs. (although, paper indicates that a node becomes a follower when it receives an RPC message with a higher term, requestvote is an RPC).

src/cluster.c
```diff
@@ -1857,10 +1857,10 @@ void ShardGroupGet(RedisRaftCtx *rr, RedisModuleCtx *ctx)
         node_count++;
         RedisModule_ReplyWithArray(ctx, 2);
 
-        char node_id[RAFT_SHARDGROUP_NODEID_LEN + 1];
+        char node_id[RAFT_SHARDGROUP_NODEID_LEN * 2];
         raft_node_id_t id = raft_node_get_id(raft_node);
 
-        snprintf(node_id, sizeof(node_id), "%s%08x", rr->log->dbid, id);
+        snprintf(node_id, sizeof(node_id), "%s%08x", rr->meta.dbid, id);
```
sjpotter, line 1863: elsewhere we use

```
    char node_id[RAFT_SHARDGROUP_NODEID_LEN + 1];
    raftNodeToString(node_id, rr->meta.dbid, raft_node);
```

which basically translates to

```
```

why not use the same code to avoid the complaining issue?

src/log.c
```diff
@@ -236,19 +164,19 @@ static int writeEntry(Log *log, raft_entry_t *ety)
...
+    n += multibulkWriteStr(buf + n, sizeof(buf) - n, RAFTLOG_STR);
+    n += multibulkWriteLong(buf + n, sizeof(buf) - n, RAFTLOG_VERSION);
+    n += multibulkWriteStr(buf + n, sizeof(buf) - n, log->dbid);
+    n += multibulkWriteLong(buf + n, sizeof(buf) - n, log->node_id);
+    n += multibulkWriteLong(buf + n, sizeof(buf) - n, log->snapshot_last_term);
+    n += multibulkWriteLong(buf + n, sizeof(buf) - n, log->snapshot_last_idx);
 
     if (truncateFiles(log, 0, 0) != RR_OK ||
-        FileWrite(&log->file, buf, off) != off ||
+        FileWrite(&log->file, buf, n) != n ||
```
sjpotter, line 179: I don't understand this change.  why did it change from `off` to 0?

src/log.c
```diff
@@ -180,40 +125,23 @@ static int readHeader(Log *log)
...
-    off += writeLength(buf + off, sizeof(buf) - off, '$', ety->data_len);
+    n += multibulkWriteLen(buf + n, sizeof(buf) - n, '*', ENTRY_ELEM_COUNT);
+    n += multibulkWriteStr(buf + n, sizeof(buf) - n, ENTRY_STR);
+    n += multibulkWriteLong(buf + n, sizeof(buf) - n, ety->term);
+    n += multibulkWriteLong(buf + n, sizeof(buf) - n, ety->id);
+    n += multibulkWriteLong(buf + n, sizeof(buf) - n, ety->type);
+    n += multibulkWriteLen(buf + n, sizeof(buf) - n, '$', (int) ety->data_len);
 
-    if (FileWrite(&log->file, buf, off) != (ssize_t) off ||
+    if (FileWrite(&log->file, buf, n) != (ssize_t) n ||
```
sjpotter, line 144: again, dont understand the 0 length here.

sjpotter: never mind, n, not 0.

src/log.c
sjpotter: never mind n, not 0, I think `n` is a bit problematic here, `off` was better imo.

src/cluster.c
tezc: ok, used `raftNodeToString()`.

src/log.c
tezc: I just want to respect 80 column limit when it is possible, that was the reason for that adjustment. I added another commit, line is shorter, let me know if this is better.

src/metadata.c
```diff
@@ -102,9 +131,15 @@ int MetadataRead(Metadata *m, const char *filename)
     RedisModule_Assert(elem_count == METADATA_ELEM_COUNT);
     RedisModule_Assert(strncmp(str, METADATA_STR, strlen(METADATA_STR)) == 0);
     RedisModule_Assert(version == METADATA_VERSION);
+    RedisModule_Assert(strlen(dbid) == RAFT_DBID_LEN);
 
+    strcpy(m->dbid, dbid);
```
fadidahanna, line 147: why not to expand `MetadataConfigure()` to all metadata fields configuration, and call `MetadataConfigure()` here?

src/metadata.c
```diff
@@ -15,45 +15,76 @@
...
+    RedisModule_Free(m->filename);
+}
+
+void MetadataConfigure(Metadata *m, const char *filename, char *dbid,
+                       raft_node_id_t node_id)
+{
+    char buf[PATH_MAX];
+
+    RedisModule_Free(m->filename);
+    metadataFilename(buf, sizeof(buf), filename);
```
fadidahanna, line 47: Why not do it in `MetadataInit()`?

tezc: Until we try to read metadata file, we don't know if the cluster is initialized. Currently, uninitialized RedisRaft allows reconfiguration of the `log_filename`. If we set filename in `MetadataInit()` and the cluster is not initialized, and then if admin changes the `log_filename`, metadata will still have the old name. 

So, I added another function, `MetadataConfigure()` which sets all the config(dbid, node_id, filename), when we know cluster is initialized: either on cluster init or on join. After that point, config cannot be changed anyway. 

The whole point is to allow admin to start an empty RedisRaft and be able to configure it from cli.

fadidahanna (approved):

PR #545, opened 2022-12-07
Bump elle-cli version

tezc:
A new version is released and causing a test failure:
https://github.com/RedisLabs/redisraft/actions/runs/3634573321/jobs/6132789114

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -40,7 +40,7 @@ jobs:
     - name: Build
       run: |
         mkdir build && cd build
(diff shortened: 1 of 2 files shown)


tezc:
After fix: https://github.com/tezc/redisraft/actions/runs/3637751900/jobs/6139122080

fadidahanna:
Suggest changing `Checkout elle-cli` step to take a specific fixed version.

tezc:
@fadidahanna added

fadidahanna (approved):

sjpotter:
missed this, but we should checkout a tag, otherwise this will continue to happen.

PR #549, opened 2022-12-12
[RR-85] Add brute force crash test

tezc:
As part of this commit,
- Added proper checks for CRLF at the end of multibulk elements.
- Added brute force crash test to detect corrupt data
  Loop over bytes of the log file and change a single byte at a time. 
  Then, try to open and load entries.
  
  RedisRaft does not handle corruption (e.g. bit rod) but crash while 
  writing to the log file can create corrupt data on disk. This test 
  verifies on restart, RedisRaft is able to start and discard this partially
  written data.

### DIFF
--- a/tests/unit/test_log.c
+++ b/tests/unit/test_log.c
@@ -37,25 +37,24 @@ static int teardown_log(void **state)
     return 0;
 }
 
-static raft_entry_t *__make_entry_value(int id, const char *value)
+static raft_entry_t *make_entry(int id, const char *value)
 {
+    char buf[64];
+
+    if (!value) {
+        snprintf(buf, sizeof(buf), "value%d\n", id);
+        value = buf;
+    }
+
     raft_entry_t *e = raft_entry_new(strlen(value) + 1);
     e->id = id;
     strcpy(e->data, value);
     return e;
 }
 
-static raft_entry_t *__make_entry(int id)
+static void append_entry(Log *log, int id, const char *value)
 {
-    raft_entry_t *e = raft_entry_new(50);
-    e->id = id;
-    snprintf(e->data, 49, "value%d\n", id);
-    return e;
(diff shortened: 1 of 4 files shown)


tests/unit/test_log.c
```diff
@@ -553,6 +555,103 @@ static void test_crc32c(void **state)
...
+
+    size_t entry_begin = st.st_size;
+
+    log = LogOpen(LOGNAME, 0);
+    LogLoadEntries(log);
+    e = __make_entry_value(6000, "test6000");
+    LogAppend(log, e);
+    raft_entry_release(e);
+    LogFree(log);
+
```
fadidahanna, line 612: suggest adding another entry in order to verify that the last entry (test7000) will be deleted too in corruption.

tezc: done

fadidahanna: it should be added at the file init step too... 🙏

tezc: ok, moved that part into the loop.

Also fixed a naming issue on the way, symbols starting with `__` reserved for C language/compilers. Linter was highlighting it.

tests/unit/test_log.c
```diff
@@ -553,6 +554,99 @@ static void test_crc32c(void **state)
...
+}
+
+/* Loop over an entry's bytes and change one byte at a time.
+ * Verify we detect the corruption when we try to read the file. */
+static void test_corruption_entry(void **state)
+{
+    raft_entry_t *e;
+    struct stat st;
+
+    Log *log = LogCreate(LOGNAME, DBID, 1, 0, 1);
```
fadidahanna, line 595: please add a comment here, to better readability of this part.

fadidahanna (approved):

PR #552, opened 2022-12-13
[RR-125] Add acl/oom permutations to regular readonly cmd tests

sjpotter:


tests/integration/test_sanity.py
```diff
@@ -189,6 +189,32 @@ def test_readonly_commands(cluster):
...
+    # no oom, acl allow
+    assert cluster.execute("get", "abc") == b'1234'
+
+    # oom, acl allow
+    cluster.execute('config', 'set', 'maxmemory', 1)
+    assert cluster.execute("get", "abc") == b'1234'
+    cluster.execute('config', 'set', 'maxmemory', 0)
+
+    cluster.execute('acl', 'setuser', 'default', 'resetkeys')
+    # no oom, acl deny
```
tezc, line 207:
```suggestion
    # no oom, acl deny
    cluster.execute('acl', 'setuser', 'default', 'resetkeys')
```
Should we swap the lines?

tests/integration/test_sanity.py
```diff
@@ -189,6 +189,32 @@ def test_readonly_commands(cluster):
...
+    cluster.execute('set', 'abc', 1234)
+    cluster.wait_for_unanimity()
+
+    # no oom, acl allow
+    assert cluster.execute("get", "abc") == b'1234'
+
+    # oom, acl allow
+    cluster.execute('config', 'set', 'maxmemory', 1)
+    assert cluster.execute("get", "abc") == b'1234'
+    cluster.execute('config', 'set', 'maxmemory', 0)
```
tezc, line 204:
```suggestion
    cluster.config_set('maxmemory', 0)
```
I don't expect anything different but maybe we should use this function to set config on all nodes? Please consider it for other usages as well.

sjpotter: ok

tezc: Are you going to make this change? forgot to push the new commit?

sjpotter: Ive combined this new test (as it was essentially the same as the scrip test into the scrip test, just adding not script test cases.  did this change at same time

tezc (approved):

PR #553, opened 2022-12-20
[RR-146] Migrate/Import issued to not ready cluster crashes

sjpotter:
- add a test that duplicates bug as is
- prevent migrate command and import from attempting to run on a not ready cluster

### DIFF
--- a/src/migrate.c
+++ b/src/migrate.c
@@ -140,6 +140,10 @@ int cmdRaftImport(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
 {
     RedisRaftCtx *rr = &redis_raft;
 
+    if (checkRaftState(rr, ctx) == RR_ERROR) {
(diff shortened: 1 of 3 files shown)


src/migrate.c
```diff
@@ -140,6 +140,10 @@ int cmdRaftImport(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
 {
     RedisRaftCtx *rr = &redis_raft;
 
+    if (checkRaftState(rr, ctx) == RR_ERROR) {
```
tezc, line 143: Do you think we should also check if we are leader here? Something like:

```c
    if (!raft_is_leader(rr->raft)) {
        Node *leader = getLeaderNodeOrReply(rr, ctx);
        if (leader) {
            redirectCommand(rr, ctx, leader);
        }
        return;
    }
```

tezc (approved):

PR #557, opened 2022-12-27
[RR-158] Use redis client-id as raft session.

sjpotter:
* serialize/deserialize it to log file
	- also make decodeInteger/encodeInteger take an unsigned long long instead of size_t
* send it as part of AppendEntry RPC and recieve it correctly
* update unit tests to expect and work with session in raft entry.

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -751,7 +751,7 @@ static int raftSendAppendEntries(raft_server_t *raft, void *user_data,
     for (i = 0; i < msg->n_entries; i++) {
         raft_entry_t *e = msg->entries[i];
         argv[5 + i * 2] = RedisModule_Alloc(64);
-        argvlen[5 + i * 2] = snprintf(argv[5 + i * 2], 63, "%ld:%d:%lld:%d", e->term, e->id, e->session, e->type);
+        argvlen[5 + i * 2] = snprintf(argv[5 + i * 2], 63, "%ld:%d:%llu:%d", e->term, e->id, e->session, e->type);
         argvlen[6 + i * 2] = e->data_len;
         argv[6 + i * 2] = e->data;
     }
--- a/src/raft.c
+++ b/src/raft.c
@@ -543,6 +543,7 @@ static void executeLogEntry(RedisRaftCtx *rr, raft_entry_t *entry, raft_index_t 
                                              entry->data_len) != RR_OK) {
             PANIC("Invalid Raft entry");
         }
+        tmp.client_id = entry->session;
 
         RaftExecuteCommandArray(rr, rr->ctx, NULL, &tmp);
         RaftRedisCommandArrayFree(&tmp);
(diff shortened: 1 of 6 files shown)


tezc:
@sjpotter Why don't we serialize session id as part of the `raft_entry_t` itself? Similar to `term`. This is why I suggested adding it to the `raft_entry_t` struct.

src/raft.c
```diff
@@ -543,6 +543,7 @@ static void executeLogEntry(RedisRaftCtx *rr, raft_entry_t *entry, raft_index_t
                                              entry->data_len) != RR_OK) {
             PANIC("Invalid Raft entry");
         }
+        tmp.client_id = entry->session;
```
tezc, line 546: I think we should stick to a single variable name :) 

We use `client_id`, `session` and `session_id` in the PR. It might be confusing in future. If you think `client_id` is more accurate, we can even change in libraft as well if you want.

sjpotter: I think session is more accurate in the context of libraft, I think client_id might be more accurate in the context of Redis.  i.e. RedisRaft would store client_id as libraft's session.

I'm not stuck to this idea, but that's the reason for it.  we can discuss in our daily.

tezc (dismissed):

tezc:
@sjpotter Could you run MacOS build before merging? `unsigned long long` stuff is generally tricky on MacOS.

src/raft.c
```diff
@@ -750,7 +751,7 @@ static int raftSendAppendEntries(raft_server_t *raft, void *user_data,
     for (i = 0; i < msg->n_entries; i++) {
         raft_entry_t *e = msg->entries[i];
         argv[5 + i * 2] = RedisModule_Alloc(64);
-        argvlen[5 + i * 2] = snprintf(argv[5 + i * 2], 63, "%ld:%d:%d", e->term, e->id, e->type);
+        argvlen[5 + i * 2] = snprintf(argv[5 + i * 2], 63, "%ld:%d:%lld:%d", e->term, e->id, e->session, e->type);
```
tezc, line 754:
```suggestion
        argvlen[5 + i * 2] = snprintf(argv[5 + i * 2], 63, "%ld:%d:%llu:%d", e->term, e->id, e->session, e->type);
```
Should this one be `llu`?

sjpotter:
@tezc passes macos https://github.com/RedisLabs/redisraft/actions/runs/3816927779

fadidahanna (approved):

tezc (approved):

sjpotter:
can't merge because of this new cla test?

src/redisraft.c
```diff
@@ -929,7 +930,7 @@ static int cmdRaftAppendEntries(RedisModuleCtx *ctx, RedisModuleString **argv, i
 
         /* Parse additional entry fields */
         tmpstr = RedisModule_StringPtrLen(argv[5 + 2 * i], &tmplen);
-        if (sscanf(tmpstr, "%ld:%d:%d", &e->term, &e->id, &e->type) != 3) {
+        if (sscanf(tmpstr, "%ld:%d:%lld:%d", &e->term, &e->id, &e->session, &e->type) != 4) {
```
tezc, line 933:
```suggestion
        if (sscanf(tmpstr, "%ld:%d:%llu:%d", &e->term, &e->id, &e->session, &e->type) != 4) {
```

src/raft.c
tezc, line 754:
```suggestion
        argvlen[5 + i * 2] = snprintf(argv[5 + i * 2], 63, "%ld:%d:%llu:%d", e->term, e->id, e->session, e->type);
```

sjpotter: argh, I see, didn't pull the upstream before I rebased.  argh.

fadidahanna (approved):

tezc (approved):

PR #558, opened 2022-12-29
[RR-154] Fix CPU detection for macOS arm64

tezc:
Currently, build fails on arm64 MacOS with the error: 

```
clang: error: argument unused during compilation: ‘-msse4.2’ [-Werror,-Wunused-command-line-argument]
```

This PR adds some more checks to detect CPU correctly.


sjpotter (approved):
seems fine to me if it passes CI and fixes flotilla's issues.

fadidahanna (approved):

PR #561, opened 2023-01-01
[RR-159] setup/teardown sessions on the log

sjpotter:
handle watch/unwatch at a basic level to support sessions on the log
    
doesn't do anything with it besides allocating a void * on session creation and free on session deletion.
    
includes metrics to count number of active sessions for testing purposes

### DIFF
--- a/src/log.c
+++ b/src/log.c
@@ -138,6 +138,7 @@ static size_t generateEntryHeader(raft_entry_t *ety, unsigned char *buf, size_t 
     pos += multibulkWriteLong(pos, end - pos, ety->id);
     pos += multibulkWriteUInt64(pos, end - pos, ety->session);
     pos += multibulkWriteInt(pos, end - pos, ety->type);
+
     pos += multibulkWriteLen(pos, end - pos, '$', (int) ety->data_len);
 
     return pos - buf;
--- a/src/multi.c
+++ b/src/multi.c
@@ -103,6 +103,12 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
 
     /* Are we in MULTI? */
     if (multiState->active) {
+        /* can't call WATCH within a MULTI, but doesn't error out MULTI */
+        if (cmd_len == 5 && !strncasecmp(cmd_str, "WATCH", 5)) {
+            RedisModule_ReplyWithError(ctx, "ERR WATCH inside MULTI is not allowed");
+            return true;
+        }
+
         /* We have to detect commands that are unsupported or must not be
          * intercepted and reject the transaction.
          */
--- a/src/raft.c
+++ b/src/raft.c
@@ -300,6 +300,30 @@ static RRStatus handleSharding(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRedisC
     return validateRaftRedisCommandArray(rr, ctx, cmds, slot);
 }
 
+/* returns the client session object for this CommandArray if applicable
+ * starts/creates it if necessary
+ */
+static void *getClientSession(RedisRaftCtx *rr, RaftRedisCommandArray *cmds)
+{
+    unsigned long long id = cmds->client_id;
+    int nokey;
+
+    void *client_session = RedisModule_DictGetC(rr->client_session_dict, &id, sizeof(id), &nokey);
+
+    if (nokey) {
+        RaftRedisCommand *c = cmds->commands[0];
+        size_t cmd_len;
+        const char *cmd = RedisModule_StringPtrLen(c->argv[0], &cmd_len);
+
+        if (cmd_len == 5 && strncasecmp("watch", cmd, cmd_len) == 0) {
+            client_session = RedisModule_Alloc(sizeof(void *));
+            RedisModule_DictSetC(rr->client_session_dict, &id, sizeof(id), client_session);
+        }
+    }
+
+    return client_session;
+}
+
 /* Execute all commands in a specified RaftRedisCommandArray.
  *
  * If reply_ctx is non-NULL, replies are delivered to it.
@@ -342,22 +366,14 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
         return;
     }
 
+    void *client_session = getClientSession(rr, cmds);
+    if (client_session) {} /* avoid unused variable check */
+
     for (int i = 0; i < cmds->len; i++) {
         RaftRedisCommand *c = cmds->commands[i];
 
         size_t cmdlen;
         const char *cmd = RedisModule_StringPtrLen(c->argv[0], &cmdlen);
-
-        if (strncasecmp("watch", cmd, cmdlen) == 0) {
-            unsigned long long id = cmds->client_id;
-
-            int nokey;
-            RedisModule_DictGetC(rr->session_dict, &id, sizeof(id), &nokey);
-            if (nokey) {
-                void *session = RedisModule_Alloc(sizeof(void *));
-                RedisModule_DictSetC(rr->session_dict, &id, sizeof(id), session);
-            }
-        }
 
         /* We need to handle MULTI as a special case:
         * 1. Skip the command (no need to execute MULTI in a Module context).
@@ -530,23 +546,27 @@ static void unlockDeleteKeys(RedisRaftCtx *rr, raft_entry_t *entry)
     }
 }
 
-static void freeSession(void *session)
+static void freeClientSession(void *client_session)
 {
-    RedisModule_Free(session);
+    RedisModule_Free(client_session);
 }
 
-static void endSession(RedisRaftCtx *rr, raft_entry_t *entry)
+static void endClientSession(RedisRaftCtx *rr, unsigned long long id)
+{
+    void *client_session = NULL;
+    RedisModule_DictDelC(rr->client_session_dict, &id, sizeof(id), &client_session);
+    if (client_session) {
+        freeClientSession(client_session);
+    }
+}
+
+static void handleEndClientSession(RedisRaftCtx *rr, raft_entry_t *entry)
 {
     RedisModule_Assert(entry->type == RAFT_LOGTYPE_END_SESSION);
     RaftReq *req = entry->user_data;
 
     unsigned long long id = entry->session;
-
-    void *session = NULL;
-    RedisModule_DictDelC(rr->session_dict, &id, sizeof(id), &session);
-    if (session) {
-        freeSession(session);
-    }
+    endClientSession(rr, id);
 
     if (req) {
         entryDetachRaftReq(rr, entry);
@@ -927,7 +947,7 @@ static int raftApplyLog(raft_server_t *raft, void *user_data, raft_entry_t *entr
             unlockDeleteKeys(rr, entry);
             break;
         case RAFT_LOGTYPE_END_SESSION:
-            endSession(rr, entry);
+            handleEndClientSession(rr, entry);
         default:
             break;
     }
--- a/src/raft.c
+++ b/src/raft.c
@@ -367,7 +367,7 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
     }
 
     void *client_session = getClientSession(rr, cmds);
-    (void) client_sessions;
+    (void) client_session;
 
     for (int i = 0; i < cmds->len; i++) {
         RaftRedisCommand *c = cmds->commands[i];
--- a/src/raft.c
+++ b/src/raft.c
@@ -948,6 +948,7 @@ static int raftApplyLog(raft_server_t *raft, void *user_data, raft_entry_t *entr
             break;
         case RAFT_LOGTYPE_END_SESSION:
             handleEndClientSession(rr, entry);
+            break;
         default:
             break;
     }
--- a/src/raft.c
+++ b/src/raft.c
@@ -367,7 +367,7 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
     }
 
     void *client_session = getClientSession(rr, cmds);
-    if (client_session) {} /* avoid unused variable check */
+    (void) client_sessions;
 
     for (int i = 0; i < cmds->len; i++) {
         RaftRedisCommand *c = cmds->commands[i];
--- a/src/raft.c
+++ b/src/raft.c
@@ -29,6 +29,7 @@ const char *RaftReqTypeStr[] = {
     [RR_IMPORT_KEYS] = "RR_IMPORT_KEYS",
     [RR_MIGRATE_KEYS] = "RR_MIGRATE_KEYS",
     [RR_DELETE_UNLOCK_KEYS] = "RR_DELETE_UNLOCK_KEYS",
+    [RR_END_SESSION] = "RR_END_SESSION",
 };
 
 /* Forward declarations */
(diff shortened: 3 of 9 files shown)


src/raft.c
```diff
@@ -347,6 +348,17 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
         size_t cmdlen;
         const char *cmd = RedisModule_StringPtrLen(c->argv[0], &cmdlen);
 
+        if (strncasecmp("watch", cmd, cmdlen) == 0) {
```
tezc, line 351: `strncasecmp("watch", "w", 1) == 0` will enter this if check. So, I suggest adding len check first:

```c
if (cmdlen == 5 && strncasecmp("watch", cmd, cmdlen) == 0) {
```

src/redisraft.c
```diff
@@ -555,6 +555,30 @@ static void handleAsking(RedisRaftCtx *rr, RedisModuleCtx *ctx)
     RedisModule_ReplyWithSimpleString(ctx, "OK");
 }
 
+static void handleEndSession(RedisRaftCtx *rr, RedisModuleCtx *ctx, char *reason)
+{
+    RaftReq *req = RaftReqInit(ctx, RR_END_SESSION);
+    if (req == NULL) {
```
tezc, line 561: It won't be null ever. We can skip the check

src/redisraft.c
```diff
@@ -597,6 +621,11 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
         return true;
     }
 
+    if (len == strlen("UNWATCH") && strncasecmp(cmd_str, "UNWATCH", len) == 0) {
```
tezc, line 624: Why do we handle `unwatch` here but handle `watch` in the state machine? Let's send `unwatch` there. We can handle it next to `watch`.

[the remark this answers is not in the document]
tezc: also `WATCH` can be inside of multi-exec?. We need to check Redis' behavior and mimic it here.

src/log.c
```diff
@@ -138,6 +138,7 @@ static size_t generateEntryHeader(raft_entry_t *ety, unsigned char *buf, size_t
     pos += multibulkWriteLong(pos, end - pos, ety->id);
     pos += multibulkWriteUInt64(pos, end - pos, ety->session);
     pos += multibulkWriteInt(pos, end - pos, ety->type);
+
```
tezc, line 141:
```suggestion
```
Is this change intentional? If not, let's get rid of it.

src/raft.c
```diff
@@ -347,6 +348,17 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
...
+            unsigned long long id = cmds->client_id;
+
+            int nokey;
+            RedisModule_DictGetC(rr->session_dict, &id, sizeof(id), &nokey);
+            if (nokey) {
+                void *session = RedisModule_Alloc(sizeof(void *));
+                RedisModule_DictSetC(rr->session_dict, &id, sizeof(id), session);
```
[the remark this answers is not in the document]
tezc: Not sure if this is the same thing but I think, instead of `session` in libraft, we can call it `client_id`.

```c
struct raft_entry_t { 
    ...
    raft_client_id client_id; // Instead of raft_session session
}
```
I think this is easier to understand and also it matches with the function name of redis: `RM_GetClientId()` -> use the return value to set client id in the entry. 

You set the client id of the entry, then do whatever you want to do with it, implement sessions as we do. 

Now it requires libraft change but if we all agree it is better, we can quickly rename it? Please comment.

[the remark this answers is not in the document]
sjpotter: so the argument i made before is that libraft shouldn't be tied to redis semantics.  in redis a "session" is tied to a client id, but it doesn't have to be that way in other systems.  I'm not tied to this interpretation, but I feel "session" should be generic, while the implementation on top of libraft should refer to it whatever makes the most sense for it.

I can change the session variable here to be more understandable.

src/redisraft.c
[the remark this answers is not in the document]
sjpotter: WATCH can't be inside a multiexec (tested previously).  Apparently, UNWATCH can be.  However, it seems to always cause the command to return nil.  the redis semantics seem to be poor here.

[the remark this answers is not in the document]
sjpotter: change that.  nil on dirty is what exec always returns.  so unwatch within a multi/exec dirties the session.  weird.  In this case, I'd think we would have to think to decide what to do, as we aren't (currently) queing the multi/exec on the log, we can know at unwatch time if we are in an multi and just error it out.

src/raft.c
sjpotter: ok, makes sense.

src/redisraft.c
sjpotter: well, this code is about to be rewritten, req can't be created here (no req on disconnect, this actually was the problem I hit earlier that I thought was a redis bug, still might be, but I was doing something bad.  So will have to have a req ==/!= NULL code path, but this isn't it.

src/redisraft.c
sjpotter: confused what you mean that UNWATCH isn't being handled in the state machine?  this appends a raft entry.  no actually unwatch happens here.  The point of handling it here, is to not handle it a "redis command" and append it as a redis command, but as a unique "end session" object.

src/log.c
sjpotter: it sort of was.  the ety-> entries are the raft_entry_t header (id/session/type), the next entry really belong with the data (i.e. we have data_len here, but data is appended separately) which gets appended externally to this, so wanted to make it clear.

src/redisraft.c
tezc: Currently, in `RaftExecuteCommandArray()` you have

```c
if (strncasecmp("watch", cmd, cmdlen) == 0) {
....
    createSession();
...
} 
```

you can add this part:

```c
if (strncasecmp("watch", cmd, cmdlen) == 0) {
...
    createSession();
...
} else if (strncasecmp("unwatch", cmd, cmdlen) == 0) {
...
    endSession();
...
}
```

Then, we don't need to intercept `unwatch` here (on append time), similar to `watch`. So, we handle these commands specially in the state machine. 

Then, we'll use `RAFT_LOGTYPE_END_SESSION` entry just for disconnect events.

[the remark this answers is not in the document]
sjpotter: so an even more update.  unwatch is usable in multi (so current code is problematic), BUT, it doesn't do anything in reality.  i.e. if the watched value in concurrently modified, nil is returned (as multi is dirty), which makes sense as UNWATCH only would execute on exec, but already dirty).  And if its not dirty, you just get an OK back.  It would make more sense to reject it, like it rejects WATCH

(error) ERR WATCH inside MULTI is not allowed

src/raft.c
```diff
@@ -518,6 +530,31 @@ static void unlockDeleteKeys(RedisRaftCtx *rr, raft_entry_t *entry)
...
+
+static void endSession(RedisRaftCtx *rr, raft_entry_t *entry)
+{
+    RedisModule_Assert(entry->type == RAFT_LOGTYPE_END_SESSION);
+    RaftReq *req = entry->user_data;
+
+    unsigned long long id = entry->session;
+
+    void *session = NULL;
+    RedisModule_DictDelC(rr->session_dict, &id, sizeof(id), &session);
```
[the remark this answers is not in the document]
fadidahanna: Agree that the implementation of `RedisModule_DictDelC` do it, but we should stick to the spec, and verify also the return value

src/raft.c
```diff
@@ -347,6 +348,17 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
         size_t cmdlen;
         const char *cmd = RedisModule_StringPtrLen(c->argv[0], &cmdlen);
 
+        if (strncasecmp("watch", cmd, cmdlen) == 0) {
+            unsigned long long id = cmds->client_id;
+
+            int nokey;
+            RedisModule_DictGetC(rr->session_dict, &id, sizeof(id), &nokey);
```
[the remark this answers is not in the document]
sjpotter: having a getSession() that returns the session object / creates it if neccesary (first watch)

src/raft.c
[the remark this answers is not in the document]
sjpotter: I believe this is fine?  session = NULL. after DictDelC, either it will still be NULL (no session set for this id) or it won't be NULL (yes, session).

as we then do

    if (session) {
        freeSession(session);
    }
    
it should all be fine.

src/redisraft.c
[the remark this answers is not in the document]
sjpotter: refactored, so UNWATCH will be handled in normal redis command path so will be enqueued into a MULTI (pointless, but what redis does)

src/raft.c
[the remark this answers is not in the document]
sjpotter: session -> client_session and functions are *ClientSession()

src/raft.c
sjpotter: done

src/redisraft.c
sjpotter: its not just for disconnect, it can also be used for multi abort / discard.

src/log.c
tezc: Change is okay but probably it'd be nicer as part of another PR. As this is the only change in log.c, when you look that this PR later or to the commit history, it'll give you an impression that we changed log.c as part of this PR.

src/raft.c
```diff
@@ -341,6 +366,9 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
         return;
     }
 
+    void *client_session = getClientSession(rr, cmds);
+    if (client_session) {} /* avoid unused variable check */
```
tezc, line 370:
```suggestion
    (void) client_sessions;
```
This should also work, just fyi.

src/raft.c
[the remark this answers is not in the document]
tezc: I thought we would change variable name in libraft, no?

To be honest, I don't like session->client_session change. In general I don't like verbose naming as it makes code larger/longer. `session` was perfectly clear to me, "client_session" doesn't make it more readable imho. 

Just to be clear, my point was referring `client_id` <-> `session` <-> `session_id` interchangeably in the code. I find this a bit confusing / harder to understand. I think confusion comes from `raft_entry_t` as it has `session` variable (it is an id) and we also have "session" struct/logic in redisraft. To fix this issue, we can rename `raft_entry_t`'s `session` to `client_id`. Then, it'll be easier to understand, at least this was my confusion / problem.

Anyway, I'm not even sure we discuss same thing :) Maybe I misunderstood initial comment in this thread.

src/redisraft.c
tezc: We can directly write `UNWATCH`, `DISCARD` commands to the log and process differently (by destroying related session) in the state machine. 

It'll be same functionality-wise but a step closer to the "real sessions". Send commands/events to the state machine and process there.

Then, for the disconnect event, we create that entry(`RAFT_LOGTYPE_CLIENT_DISCONNECT`)  and write to the log. So, these two commands (unwatch, discard) and disconnect event will do the same thing currently: destroy the session. In future, maybe they won't. (depends on what we implement).

sjpotter: the reason I didn't do it that way, is I dislike the text parsing of UNWATCH/DISCARD at apply time.  It just bothers me when dealing with the state machine to work based on string comparisons.  if we want to do it that way, I can implement it, just rus me the wrong way.

tezc: ignoring for now as this is not critical

src/raft.c
```diff
@@ -889,6 +946,8 @@ static int raftApplyLog(raft_server_t *raft, void *user_data, raft_entry_t *entr
         case RAFT_LOGTYPE_DELETE_UNLOCK_KEYS:
             unlockDeleteKeys(rr, entry);
             break;
+        case RAFT_LOGTYPE_END_SESSION:
+            handleEndClientSession(rr, entry);
```
tezc, line 950:
```suggestion
            handleEndClientSession(rr, entry);
            break;
```

Let's add this not to see a surprise later

src/redisraft.h
```diff
@@ -418,9 +418,10 @@ typedef struct RedisRaftCtx {
     unsigned long snapshotreq_received;          /* Number of received snapshotreq messages */
     unsigned long exec_throttled;                /* Number of command executions throttled due to slow execution */
 
-    int entered_eval;             /* handling a lua script */
-    RedisModuleDict *locked_keys; /* keys thar have been locked for migration */
-    RedisModuleDict *acl_dict;    /* maps acl strings to RedisModuleUser * objects */
+    int entered_eval;                     /* handling a lua script */
+    RedisModuleDict *locked_keys;         /* keys thar have been locked for migration */
```
tezc, line 422:
```suggestion
    RedisModuleDict *locked_keys;         /* keys that have been locked for migration */
```
let's fix this on the way

src/redisraft.c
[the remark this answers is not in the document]
tezc: Okay, are we going to prevent WATCH inside multi in this PR or in another PR? 
Btw, I guess we can even just return OK to UNWATCH without adding it to the queue.

src/redisraft.c
```diff
@@ -621,6 +663,10 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
         return;
     }
 
+    if (handleUnwatch(rr, ctx, cmds)) {
```
tezc, line 666: I'm scared to ask this question but hope ACL does not apply to unwatch :) 
dry-run is a bit down below, we skip ACL check currently, maybe we can move this down a bit.

src/raft.c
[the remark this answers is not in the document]
sjpotter: @fadidahanna ?

src/redisraft.c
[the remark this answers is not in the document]
sjpotter: prevent watch within MULTI.

src/redisraft.c
```diff
@@ -555,6 +555,28 @@ static void handleAsking(RedisRaftCtx *rr, RedisModuleCtx *ctx)
     RedisModule_ReplyWithSimpleString(ctx, "OK");
 }
 
+static void handleEndClientSession(RedisRaftCtx *rr, RaftReq *req, unsigned long long id, char *reason)
```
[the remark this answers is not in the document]
sjpotter: ok applied

src/redisraft.c
sjpotter: it can.  so, we should.  moving.

src/multi.c
```diff
@@ -103,6 +103,12 @@ bool MultiHandleCommand(RedisRaftCtx *rr,
 
     /* Are we in MULTI? */
     if (multiState->active) {
+        /* can't call WATCH within a MULTI, but doesn't error out MULTI */
+        if (cmd_len == 5 && !strncasecmp(cmd_str, "WATCH", 5)) {
+            RedisModule_ReplyWithError(ctx, "ERR WATCH inside MULTI is not allowed");
+            return true;
```
tezc, line 109:
```suggestion
            multiState->error = true;
            return true;
```

Should we fail the transaction?

src/redisraft.c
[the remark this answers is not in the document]
tezc: pff, horrible. I hope we can reject unwatch as well

src/raft.c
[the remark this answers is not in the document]
fadidahanna: We agreed in our daily to change raft session into session id, and then to solve the ambiguity, but we can do it later.

PR #564, opened 2023-01-04
[RR-161] Support ending sessions on client disconnect 

sjpotter:
On client disconnect, append a log entry ending any session belonging to this client.

Only append the log entry for clients that have attempted regular redis commands through the RedisRaft mechanism.  For instance, clients belonging to peer nodes should not have this appended.


src/redisraft.c
```diff
@@ -622,6 +628,21 @@ static bool handleInterceptedCommands(RedisRaftCtx *rr,
     return false;
 }
 
+/* TODO: used to only add entry for clients that have actually watched,
+ *       needed to avoid testing inconsistencies for now
+ */
+static void handleWatch(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRedisCommandArray *cmds)
```
fadidahanna, line 634: Do we have to define `watch` as a client state? We already keep the information in watched sessions in `client_session_dict`, and we should use this dict in order to call `UNWATCH`, even in disconnection - otherwise, Redis will not update its watched_keys dict.

sjpotter: this is to fix the testing issue of having possible inconsistent number of log entries in our tests.  Its not "needed", as we could always send the log entry and it wouldn't do harm to a real system, it just breaks tests that expect the number of entries to always be the same.

We can't rely on the client_session_dict because it can disconnect after append but before apply and then a client_session will be created and leak.  With that said, its possible we could be "ok" with said leak (will be cleaned up at election time, so not a permanent leak), and then we could just rely on the client_session_dict without this new entry.

fadidahanna (approved):

sjpotter:
In going to commit this, I'm wondering if the logic of when to issue end session/disconnect can be simplified.

We currently say 2 things have to be true to send it

1) we have to know that its a real user client (sent a regular redis command that raft would append to log

2) we then added that it had to have used WATCH at some point.

in practice 2 includes 1.  So perhaps we can simplify it.  The reason to not to simplify it, is that it makes it clear what's going on.

1 is there to make sure we dont send disconnect requests for inter-node cluster clients (i.e. what they use to communicate with each other)

2 is there to make tests consistent.  While 2 contains 1, if we ever decide to remove 2, we could forget about the need for 1.

On the flip side, I'm not sure anymore if it matters if we send a "disconnect" log entry for the "inter-node" clients.  i.e. its a waste of a log entry, but shouldn't cause any harm.

Before I merge it, figured should discuss.

tezc:
@sjpotter If we are going to implement read-only multis, one way or another we need to figure out if there is a session, so I guess we'll revisit this issue later. I feel like we cannot delete "2" without making it better / more precise. Personally, I think we should be okay without "1" right now.

fadidahanna (approved):

PR #565, opened 2023-01-05
[RR-181] clear sessions on NO_OP

sjpotter:
Upon apply of a NO_OP, clear all active sessions.

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -575,6 +575,19 @@ static void handleEndClientSession(RedisRaftCtx *rr, raft_entry_t *entry)
     }
 }
 
+static void clearClientSessions(RedisRaftCtx *rr)
+{
+    void *client_session;
+    RedisModuleDictIter *iter = RedisModule_DictIteratorStartC(rr->client_session_dict, "^", NULL, 0);
+
+    while (RedisModule_DictNextC(iter, NULL, (void **) &client_session) != NULL) {
+        freeClientSession(client_session);
+    }
+    RedisModule_DictIteratorStop(iter);
+    RedisModule_FreeDict(rr->ctx, rr->client_session_dict);
+    rr->client_session_dict = RedisModule_CreateDict(rr->ctx);
+}
+
 /*
  * Execution of Raft log on the local instance.
  *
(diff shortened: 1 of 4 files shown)


src/raft.c
```diff
@@ -575,6 +575,19 @@ static void handleEndClientSession(RedisRaftCtx *rr, raft_entry_t *entry)
...
+{
+    void *client_session;
+    RedisModuleDictIter *iter = RedisModule_DictIteratorStartC(rr->client_session_dict, "^", NULL, 0);
+
+    while (RedisModule_DictNextC(iter, NULL, (void **) &client_session) != NULL) {
+        freeClientSession(client_session);
+    }
+    RedisModule_DictIteratorStop(iter);
+    RedisModule_FreeDict(rr->ctx, rr->client_session_dict);
+    rr->client_session_dict = RedisModule_CreateDict(rr->ctx);
```
fadidahanna, line 588: We don't have to free and alloc the whole dictionary here, but we should make sure that we free it at `RedisRaftCtxClear()` (+ please make sure that we actually free there the inner objs memory, for other dicts as well).

src/redisraft.c
```diff
@@ -2031,7 +2031,7 @@ RRStatus RedisRaftCtxInit(RedisRaftCtx *rr, RedisModuleCtx *ctx)
     rr->locked_keys = RedisModule_CreateDict(rr->ctx);
 
     /* acl -> user dictionary */
-    rr->acl_dict = RedisModule_CreateDict(ctx);
+    rr->acl_dict = RedisModule_CreateDict(rr->ctx);
 
     rr->client_session_dict = RedisModule_CreateDict(ctx);
```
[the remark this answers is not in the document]
sjpotter: I think I made this change in the other PR.

src/raft.c
sjpotter: So, I wasn't sure if we could delete items while iterating without breaking the iteration.

And, I wasn't sure if its actually less performance to do many individual deletes from the dict, vs just deleting and recreating the dict.

that's why I went this way, if we think its better in other way, that's ok too.

fadidahanna: k, checked now on Redis - indeed it seems like your practice is the right one, so let's keep it.

fadidahanna (approved):

PR #567, opened 2023-01-09
[RR-184] Force close clients that exist while clearing sessions

sjpotter:
In order to prevent clients with sessions from persisting if leadership transfers away and then returns, we force close them when clearing the session.

If we did not do this, a client that sent a WATCH could believe that its WATCH is still valid.  terminating the client connection will force the client to reconnect.

To not require all clients to do this, we only close those that are in the middle of a session.

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -368,7 +368,7 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
     }
 
     ClientSession *client_session = getClientSession(rr, cmds);
-    if (reply_ctx) {
+    if (client_session && reply_ctx) {
         client_session->local = true;
     }
     (void) client_session;
--- a/src/raft.c
+++ b/src/raft.c
@@ -308,7 +308,7 @@ static void *getClientSession(RedisRaftCtx *rr, RaftRedisCommandArray *cmds)
     unsigned long long id = cmds->client_id;
     int nokey;
 
-    void *client_session = RedisModule_DictGetC(rr->client_session_dict, &id, sizeof(id), &nokey);
+    ClientSession *client_session = RedisModule_DictGetC(rr->client_session_dict, &id, sizeof(id), &nokey);
 
     if (nokey) {
         RaftRedisCommand *c = cmds->commands[0];
@@ -316,7 +316,9 @@ static void *getClientSession(RedisRaftCtx *rr, RaftRedisCommandArray *cmds)
         const char *cmd = RedisModule_StringPtrLen(c->argv[0], &cmd_len);
 
         if (cmd_len == 5 && strncasecmp("watch", cmd, cmd_len) == 0) {
-            client_session = RedisModule_Alloc(sizeof(void *));
+            client_session = RedisModule_Alloc(sizeof(ClientSession));
+            client_session->client_id = id;
+            client_session->session_term = raft_get_current_term(rr->raft);
             RedisModule_DictSetC(rr->client_session_dict, &id, sizeof(id), client_session);
         }
     }
@@ -577,10 +579,10 @@ static void handleEndClientSession(RedisRaftCtx *rr, raft_entry_t *entry)
 
 static void clearClientSessions(RedisRaftCtx *rr)
 {
-    void *client_session;
+    ClientSession *client_session;
     RedisModuleDictIter *iter = RedisModule_DictIteratorStartC(rr->client_session_dict, "^", NULL, 0);
-
     while (RedisModule_DictNextC(iter, NULL, (void **) &client_session) != NULL) {
+        RedisModule_DeauthenticateAndCloseClient(rr->ctx, client_session->client_id);
         freeClientSession(client_session);
     }
     RedisModule_DictIteratorStop(iter);
--- a/src/raft.c
+++ b/src/raft.c
@@ -318,7 +318,6 @@ static void *getClientSession(RedisRaftCtx *rr, RaftRedisCommandArray *cmds)
         if (cmd_len == 5 && strncasecmp("watch", cmd, cmd_len) == 0) {
             client_session = RedisModule_Alloc(sizeof(ClientSession));
             client_session->client_id = id;
-            client_session->session_term = raft_get_current_term(rr->raft);
             RedisModule_DictSetC(rr->client_session_dict, &id, sizeof(id), client_session);
         }
     }
@@ -368,7 +367,10 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
         return;
     }
 
-    void *client_session = getClientSession(rr, cmds);
+    ClientSession *client_session = getClientSession(rr, cmds);
+    if (reply_ctx) {
+        client_session->local = true;
+    }
     (void) client_session;
 
     for (int i = 0; i < cmds->len; i++) {
@@ -582,7 +584,9 @@ static void clearClientSessions(RedisRaftCtx *rr)
     ClientSession *client_session;
     RedisModuleDictIter *iter = RedisModule_DictIteratorStartC(rr->client_session_dict, "^", NULL, 0);
     while (RedisModule_DictNextC(iter, NULL, (void **) &client_session) != NULL) {
-        RedisModule_DeauthenticateAndCloseClient(rr->ctx, client_session->client_id);
+        if (client_session->local) {
+            RedisModule_DeauthenticateAndCloseClient(rr->ctx, client_session->client_id);
+        }
         freeClientSession(client_session);
     }
     RedisModule_DictIteratorStop(iter);
--- a/src/raft.c
+++ b/src/raft.c
@@ -318,6 +318,7 @@ static void *getClientSession(RedisRaftCtx *rr, RaftRedisCommandArray *cmds)
         if (cmd_len == 5 && strncasecmp("watch", cmd, cmd_len) == 0) {
             client_session = RedisModule_Alloc(sizeof(ClientSession));
             client_session->client_id = id;
+            client_session->local = false;
             RedisModule_DictSetC(rr->client_session_dict, &id, sizeof(id), client_session);
         }
     }
--- a/src/redisraft.h
+++ b/src/redisraft.h
@@ -728,8 +728,8 @@ typedef struct ClientState {
 } ClientState;
 
 typedef struct ClientSession {
-    raft_term_t session_term;
     unsigned long long client_id;
+    bool local;
 } ClientSession;
 
 /* common.c */
--- a/src/redisraft.h
+++ b/src/redisraft.h
@@ -727,6 +727,11 @@ typedef struct ClientState {
     bool asking;
 } ClientState;
 
+typedef struct ClientSession {
+    raft_term_t session_term;
+    unsigned long long client_id;
+} ClientSession;
+
 /* common.c */
 void joinLinkIdleCallback(Connection *conn);
 void joinLinkFreeCallback(void *privdata);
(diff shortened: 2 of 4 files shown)


src/raft.c
```diff
@@ -308,15 +308,17 @@ static void *getClientSession(RedisRaftCtx *rr, RaftRedisCommandArray *cmds)
...
     if (nokey) {
         RaftRedisCommand *c = cmds->commands[0];
         size_t cmd_len;
         const char *cmd = RedisModule_StringPtrLen(c->argv[0], &cmd_len);
 
         if (cmd_len == 5 && strncasecmp("watch", cmd, cmd_len) == 0) {
-            client_session = RedisModule_Alloc(sizeof(void *));
+            client_session = RedisModule_Alloc(sizeof(ClientSession));
+            client_session->client_id = id;
+            client_session->session_term = raft_get_current_term(rr->raft);
```
tezc, line 321: `term` will be different on log replay, so this is something not deterministic. Does this cause a problem somewhere? Where are we going to use this variable?

src/raft.c
```diff
@@ -577,10 +579,10 @@ static void handleEndClientSession(RedisRaftCtx *rr, raft_entry_t *entry)
 
 static void clearClientSessions(RedisRaftCtx *rr)
 {
-    void *client_session;
+    ClientSession *client_session;
     RedisModuleDictIter *iter = RedisModule_DictIteratorStartC(rr->client_session_dict, "^", NULL, 0);
-
     while (RedisModule_DictNextC(iter, NULL, (void **) &client_session) != NULL) {
+        RedisModule_DeauthenticateAndCloseClient(rr->ctx, client_session->client_id);
```
tezc, line 585: I think we should consider log replay and free clients only if we are in the current term? Otherwise, we may free unrelated clients, no?

tezc:
Btw, I realize one other scenario. We may directly receive a snapshot and don't get NOOP. Then, we still need to free clients that have sessions? 

We may leave it as it is but seems a bit fragile:
1- The node is the leader
2- Client issues "watch"
3- The node receives a snapshot and becomes a follower
--  4- Client issues "exec" -> rejected with redirect
--  4- The node becomes a leader first, we append client's multi-exec before applying NOOP <-- Currently, we don't allow appending before NOOP but we become dependend on this behavior.

Maybe there are some other cases as well? Wdyt?

src/raft.c
sjpotter: you might be right that its not needed.  I think I was originally thinking we need to compare it, then realized we didn't.

src/raft.c
tezc: Also, we may disconnect unrelated clients on followers this way. 
Maybe similar to you did, we need to store term in the session object and disconnect clients if the `NOOP term = session_term + 1`.

sjpotter: so I think its impossible to have a "client_session" (i.e. client_id -> client_session" mapping) that will be in the current term.

why?

the way we create a client_session mapping is a WATCH command on the log being applied.  

Any WATCH commands appended before the NO_OP will be on the old term.
Any WATCH commands appended after the NO_OP will be on the new term.

Any sessions created by old WATCH commands before the NO_OP will be freed/closed at NO_OP time.  This won't impact clients that issued them after the NO_OP was appended (but perhaps before it was applied) because they will come after in the log. 

Basically, I think its impossible for their to be client_sessions (at NO_OP) time that exist for the new term.  

The only way this would be wrong is if its possible for their to be log entries for a the NO_OP's term before the NO_OP, which as I said, I think is impossible (i.e. would break raft guarantees).

sjpotter:
> Btw, I realize one other scenario. We may directly receive a snapshot and don't get NOOP. Then, we still need to free clients that have sessions ?(ClientState objects that we maintain for append time stuff)
> 
> We may leave it as it is but seems a bit fragile: 1- Node A is the leader 2- Client X issues "watch" 3- Node A receives a snapshot that contains latest NOOP and becomes a follower. Node A will never recieve NOOP. -- 4- Client X issues "multi-exec" to Node A-> rejected with redirect (not leader) -- 4- Node A becomes the leader again, Client X issues multi-exec(client still thinks its watch is valid) ------ We append client's multi-exec before applying NOOP <-- Currently, we don't allow appending before applying NOOP but we become dependent on this behavior (we've changed this recently).
> 
> Maybe there are some other cases as well? Wdyt?
> 
> Edit: Thinking again, not sure if there is a problem here but probably deserves some more thinking.

don't undestand the issue.  If a node included a NO_OP in their snapshot (say as last log entry).  Then it won't include any client_sessions (the NO_OP will have cleaned them up locally, and hence wont be in snapshot).

Basically, client_sessions (i.e. WATCHes) can't persist past a NO_OP.

With that said, I did realize an issue with this PR.

old leader had client session with client_id 1.  new leader has a client connect to it that also gets client_id 1.  When new leader applies NO_OP, it will kill its "new" client that is client_id 1.

A thought I have, is that at WATCH time we know if this client is local or not.  (i.e. req).  In client_session we store a bool "local".  If req exists, we set to true, otherwise we set to false.  On snapshot we don't store it, as by definition on snapshot restore, it will always be false.

Therefore, at NO_OP time, we can make a decision if we force close the client, if its local or not.

sjpotter:
I'd like to add a test that tests this new "local" functionality.

i.e. create an environment where one has clients on say node1 (current leader) and node2 with the same client id. client on node1 issues a watch, we then transfer leader to node2.  This should just have node1 force close its client, but client with same id on node2 should persist and be usable.

the trouble is ensuring that we get the same idea reliably.  might require setting up and tearing down a number of clients till we get it + a debug api that returns us our client id.

so on node1, we would do somtehing like 

conn = None
for i in range(100):
    if conn is not None:
        conn.disconnect()
    conn = createConn()

id = conn.getId()

then for node2, we do something like

conn2 = None
conn2_id = -1
while conn2_id < id:
    if conn2 is not None:
       conn2.disconnect()
    conn2 = createConn()

assert id == conn2_id

and then can do the rest of the test as described.

tezc:
Just to recap, what we want to do is: whenever a node realizes there is a new term, it should fail existing client transactions.

There are two steps for it:
1- Destroying sessions in the state machine
2- Failing transactions on append time (this PR does it by forcing disconnection). 

Right now, tricky part is 2 as we can see duplicate client ids. 

----

One problematic scenario might be something like "ABA problem":

1- Node A is leader. Client with id 100 sends WATCH and it's applied.
2- Node B becomes leader, does bunch of things and takes a snapshot (includes the NOOP, no session in the snapshot).
3- Node A realizes Node B is the new leader and Node A receives that snapshot.
4- Node A becomes the leader again.
5- Client with id 103 is still connected to Node A. Client thinks WATCH is still valid but it is not. (No session after step 2 in the state machine.)
6- Client with id 103 issues multi-exec, it will be executed without WATCH. 

Can this scenario happen currently? Maybe there are some more corner cases, not sure right now. Wdyt?

sjpotter:
Ok, I see the issue you are saying.  Basically, "on load of snapshot" we have to clear the client_session_dict.  Yes, I didn't think about that.  But the client_session_dict on loading from snapshot should just depend on snapshot state.

sjpotter:
see latest commit I just pushed to RR-185 as I think your Q is answered by it.

Need to figure out how to write a test for it.

tezc:
>  Basically, "on load of snapshot" we have to clear the client_session_dict

That's not what I'm saying but looks like this is another problem :) Before we load a snapshot, we need to clear all the maps/dicts/arrays in the state machine, sure. 

What I am saying in other words: 
1- Client A executes "watch" and we create a ClientSession in the state machine.
2- Leader becomes a follower and gets a snapshot with no ClientSessions in it. 
3- Client A doesn't know about leader change, still connected to old leader.
4- Are we going to disconnect our client from the old leader? Our client still thinks there is a ClientSession for it. 

Now, if this follower node (old leader) becomes the leader again, client will never learn its session does not exist anymore. It can send multi-exec and we'll execute it. Client thinks it executed transaction with "watch" but in reality, we will execute it without "watch". 

So, basically, leader becomes follower and becomes the leader again without force disconnecting the client. This seems possible currently, no?

sjpotter:
when old leader loads snapshot, it will clear ClientSession dict and close all sessioned clients.  You can't get to step 3, as the loading the snapshot will kill the client.

tezc:
ok, when you say clear sessions, it will also disconnect clients. So, looks like we are talking about the same thing and I see you fixed in the other PR.

sjpotter:
so are we on the same page that this is fine.  But yes, when I say clear session it means not just clean up state machine state, but also everything else that would go with that.  It's good you brought it up though, might have skipped it.

We might want to put a comment at the top of the load snapshot function to remind us that we have to make sure that when we load snapshot, no state machine derived state from before persists, all state machine state should be derived from the snapshot, and that can include "dependencies" on state outside the state machine (in this case, connected clients) that have to be cleaned up.  In practice I think I handled this nicely by having a clearClientState function that cleaned up the state machine and ensured that if it referred to a "local" client, it was "cleaned up" (i.e. closed) as well.

sjpotter:
not an issue in ShardingInfoRDBLoad as we handle it

```
   /* If we have something to load, we need to reset ShardingInfo first.
     * We also need to be sure we're in cluster mode, i.e. that si was
     * initialized.
     */
    RedisModule_Assert(si != NULL);
    ShardingInfoReset(rr->ctx, si);
```

but is an issue in lockedKeysRDBLoad

```
static void lockedKeysRDBLoad(RedisModuleIO *rdb)
{
    size_t count = RedisModule_LoadUnsigned(rdb);

    for (size_t i = 0; i < count; i++) {
        RedisModuleString *key = RedisModule_LoadString(rdb);
        RedisModule_DictSet(rr->locked_keys, key, NULL);
        RedisModule_FreeString(NULL, key);
    }
}
```

need to reset the dict.

tezc:
btw if you are going to fix this in other PR, I guess you want to merge this? 
(just a note, a test also would be nice for the snapshot issue if it is easy. maybe as part of other PR)

tezc (approved):

sjpotter:
I can't figure out how to test it.  i.e. to get old leader to load a snapshot without restarting (as if we restart, can't have any state anyways).  help would be appreciated for that if you have an idea.

tezc:
I didn't try but maybe something like this one:

```python
def test_watch_failure(cluster):
    """
    Describe the issue
    """
    cluster.create(1)
    cluster.node(1).execute('set', 'key1', 1)

    conn = RawConnection(cluster.node(1).client)
    conn.execute('watch', 'x')

    cluster.add_node()
    cluster.add_node()
    cluster.wait_for_unanimity()

    assert cluster.node(1).transfer_leader(2) == b'OK'
    cluster.update_leader()
    cluster.node(1).pause()

    # Write some log entries and create a snapshot
    cluster.execute('set', 'z', 100)
    cluster.execute('set', 'z', 101)
    cluster.execute('set', 'z', 102)
    cluster.node(2).execute('raft.debug', 'compact')

    # node-1 will receive the snapshot
    cluster.node(1).resume()
    cluster.wait_for_unanimity()
    cluster.node(1).wait_for_info_param('raft_snapshots_received', 1)
    
    # Make node-1 leader again
    assert cluster.node(2).transfer_leader(1) == b'OK'

    # Multi-exec should not work
    with raises(RedisError):
        conn.execute('multi')
        conn.execute('set', 'x', 10)
        conn.execute('exec')

```

sjpotter:
ok, that's reasonable, will apply that to the other PR.  didn't know about pause/resume.

PR #568, opened 2023-01-09
[RR-185] Basic snapshotting

sjpotter:
as we don't have redis support yet, this just snapshots/restores the minimal session state
    
adds a test to make sure it works across multiple node restats.

RR-184 need to be merged first.


src/snapshot.c
```diff
@@ -714,6 +734,22 @@ static void lockedKeysRDBSave(RedisModuleIO *rdb)
...
+    RedisRaftCtx *rr = &redis_raft;
+    RedisModuleDict *dict = rr->client_session_dict;
+
+    RedisModule_SaveUnsigned(rdb, RedisModule_DictSize(dict));
+
+    RedisModuleDictIter *iter = RedisModule_DictIteratorStartC(dict, "^", NULL, 0);
+
+    ClientSession *clientSession;
+    while (RedisModule_DictNextC(iter, NULL, (void **) &clientSession) != NULL) {
+        RedisModule_SaveUnsigned(rdb, clientSession->client_id);
```
tezc, line 748:
```suggestion
    ClientSession *client_session;
    while (RedisModule_DictNextC(iter, NULL, (void **) &client_session) != NULL) {
        RedisModule_SaveUnsigned(rdb, client_session->client_id);
```
nitpick

tests/integration/test_snapshots.py
```diff
@@ -605,3 +608,97 @@ def test_snapshot_state_on_failure(cluster):
...
+
+
+def test_snapshot_sessions(cluster):
+    cluster.create(3)
+
+    @retry(delay=1, tries=10)
+    def assert_num_sessions(val):
+        assert cluster.node(1).info()["raft_num_sessions"] == val
+        assert cluster.node(2).info()["raft_num_sessions"] == val
+        assert cluster.node(3).info()["raft_num_sessions"] == val
```
tezc, line 620: Btw, we have `wait_for_info_param()` function to retry a few times. Is this `retry` package doing something else? 
Maybe we can also add `wait_for_info_param()` to `Cluster` object. Then, all we have to do is: 

```python
cluster.wait_for_info_param("raft_num_sessions", val);
```

tests/integration/test_snapshots.py
```diff
@@ -605,3 +608,97 @@ def test_snapshot_state_on_failure(cluster):
...
+
+    conn2 = RawConnection(cluster.leader_node().client)
+    conn2.execute("WATCH", "X")
+    cluster.wait_for_unanimity()
+
+    assert_num_sessions(2)
+
+    n2 = cluster.node(2)
+
+    assert n2.client.execute_command('RAFT.DEBUG', 'COMPACT') == b'OK'
```
tezc, line 638:
```suggestion
    assert n2.execute('RAFT.DEBUG', 'COMPACT') == b'OK'
```
Just fyi

tests/integration/test_snapshots.py
```diff
@@ -605,3 +608,97 @@ def test_snapshot_state_on_failure(cluster):
...
+    assert_num_sessions(2)
+
+    cluster.execute("set", "a", "123")
+
+    conn1.execute("UNWATCH")
+    cluster.wait_for_unanimity()
+    assert_num_sessions(1)
+
+    n2 = cluster.node(2)
+    assert n2.client.execute_command('RAFT.DEBUG', 'COMPACT') == b'OK'
```
tezc, line 652:
```suggestion
    assert n2.execute('RAFT.DEBUG', 'COMPACT') == b'OK'
```
just fyi

tests/integration/test_snapshots.py
```diff
@@ -605,3 +608,97 @@ def test_snapshot_state_on_failure(cluster):
...
+    assert_num_sessions(1)
+
+    cluster.execute("set", "b", "123")
+
+    conn2.execute("UNWATCH")
+    cluster.wait_for_unanimity()
+    assert_num_sessions(0)
+
+    n2 = cluster.node(2)
+    assert n2.client.execute_command('RAFT.DEBUG', 'COMPACT') == b'OK'
```
tezc, line 665:
```suggestion
    assert n2.execute('RAFT.DEBUG', 'COMPACT') == b'OK'
```
just fyi

tests/integration/test_snapshots.py
```diff
@@ -605,3 +608,97 @@ def test_snapshot_state_on_failure(cluster):
...
+    assert n2.info()['raft_log_entries'] == 0
+    n2.restart()
+    n2.wait_for_node_voting()
+    assert_num_sessions(0)
+
+
+def test_session_cleaned_on_load(cluster):
```
tezc, line 672: Test description would be nice here

tezc:
@sjpotter LGTM. Just have a few minor comments & nitpicks.

tests/integration/test_snapshots.py
sjpotter: if wait wont assert after some timeout, then its probably not good enough.  this is an assert.  I guess the test will eventually timeout, but I'd rather it be a quicker failure

tezc: What is special about assert?

Just to be clear, `wait_for_info_param()` already exists in `RedisRaft` class: https://github.com/RedisLabs/redisraft/blob/49b292c4f54cb681fda737bed360ea5e5ca35345/tests/integration/sandbox.py#L541

It retries a few times and throws an exception. 

So, we can also add this function to `Cluster` class, something like:

```python
    def wait_for_info_param(self, param, value, exclude=None):
        for _id, node in self.nodes.items():
            if exclude is not None and int(_id) in exclude:
                continue
            node.wait_for_info_param(param, value)
```

Then, we can use it in a single line:

```python
cluster.wait_for_info_param("raft_num_sessions", 2);
```

tests/integration/test_snapshots.py
```diff
@@ -605,3 +608,97 @@ def test_snapshot_state_on_failure(cluster):
     assert info['raft_snapshot_last_term'] == 0
     assert info['raft_snapshot_size'] == 0
     assert info['raft_snapshot_time_secs'] == -1
+
+
+def test_snapshot_sessions(cluster):
```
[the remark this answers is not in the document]
fadidahanna: @sjpotter ^^^

fadidahanna:
@sjpotter, can you please go over @tezc comments? Have no other comments from my side.

fadidahanna:
@sjpotter ^^^

tezc (approved):

PR #569, opened 2023-01-10
Fix deployment doc

LLiuJJ:
fix deployment doc raft node startup script


sjpotter:
Edited PR to remove the [RR-#] from the title, we use those for connections to JIRA that we use internally at Redis to plan and track the work Redis employees are doing.

fadidahanna (approved):

tezc:
@LLiuJJ thank you very much for the fix.

LLiuJJ:
> 

You're welcome

PR #574, opened 2023-01-17
[RR-145] Fix migration test

tezc:
This PR introduces a bug fix for migration scenario: 
- If `raft_recv_entry()` returns while appending "unlock" entry, 
  we reply twice and break the protocol. It's fixed by this commit. 

- `migrate` command can fail. 
   (e.g. leader becomes a follower while migrating). 
   
   In this case, admin should retry the operation. It was missing in 
   the tests, added failure handling and retry. 
   Important point is, current `cursor` value will be invalid, admin 
   should set `cursor` to zero and start from the beginning. 

Also, as part of this PR:
- Used `raft_get_error_str()` for raft library error strings.


src/raft.c
```diff
@@ -485,10 +485,6 @@ static void lockKeys(RedisRaftCtx *rr, raft_entry_t *entry)
 
     for (size_t i = 0; i < num_keys; i++) {
         if (RedisModule_KeyExists(rr->ctx, keys[i])) {
-            size_t str_len;
-            const char *str = RedisModule_StringPtrLen(keys[i], &str_len);
-            LOG_VERBOSE("locking %.*s", (int) str_len, str);
```
fadidahanna, line 495: These logs can help us in debugging issues, we can wrap these lines with a condition in case you care about performance.

tests/integration/test_migrate.py
```diff
@@ -81,8 +84,18 @@ def test_migration_basic(cluster_factory):
             for key in keys:
                 key_names.append(key[0].decode('utf-8'))
 
-            assert cluster1.execute('migrate', '', '', '', '', '', 'keys',
-                                    *key_names) == b'OK'
+            # Try to handle 'migrate' error response gracefully. In the process
+            # of migration, leader node can lose leadership or connection to
+            # other nodes. In this case, we just log the error response and
+            # retry the operation.
```
fadidahanna, line 90: where do we retry the migrate command?

src/util.c
```diff
@@ -541,37 +541,23 @@ const char *getStateStr(RedisRaftCtx *rr)
...
-            break;
+    if (err == RAFT_ERR_SHUTDOWN) {
+        LOG_WARNING("Raft requires immediate shutdown!");
+        shutdownServer(rr);
+    }
+
+    if (msg) {
+        snprintf(buf, sizeof(buf), "ERR %s (%s)", msg, raft_get_error_str(err));
+    } else {
+        snprintf(buf, sizeof(buf), "ERR %s", raft_get_error_str(err));
```
fadidahanna, line 557: For `RAFT_ERR_NOMEM` error we should add `OOM` prefix, not `ERR`.

src/raft.c
tezc: No, this is not about performance (I think it will also have some performance impact).

Printing log for each key makes log file bigger and makes it hard to analyze.  See the log file here: https://redislabs.atlassian.net/browse/RR-145

It can get worse if we have millions of keys. 

So, if you think it is useful, maybe we should introduce a trace level for it:
https://github.com/RedisLabs/redisraft/blob/52d28196dca6738609825ce08e176db6db2e49ee/src/config.c#L40

Maybe something like TRACE_MIGRATION. Then, probably, it'd be nice to print logs not only when we lock, also when we unlock, import etc maybe?

CMakeLists.txt
```diff
@@ -214,7 +214,6 @@ add_library(redisraft MODULE
         src/clientstate.c
         src/cluster.c
         src/commands.c
-        src/common.c
```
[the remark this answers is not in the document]
tezc: Tbh, I always felt like these two files are same :) We used both for anything. 
I'll revert this change, let's try to follow your definition from now on, we'll have better separation between these two.

tests/integration/test_migrate.py
tezc: We just continue to loop, it solves the issue. Probably, we can do it more explicit, I'll update it.

src/util.c
tezc: We won't ever get `RAFT_ERR_NOMEM`, that's why I ignored it. Do you think we can use it somehow? Currently, libraft does not return it. Even if we fix it there and return this, our malloc impl won't ever fail. So, I thought we can just ignore it.

src/raft.c
tezc: https://github.com/RedisLabs/redisraft/pull/575

tests/integration/test_migrate.py
```diff
@@ -3,11 +3,53 @@
...
+        except (redis.ResponseError, redis.TimeoutError) as e:
+            logging.info('raft.scan command failed with error: ' + str(e))
+
+            max_retries -= 1
+            if max_retries == 0:
+                raise RedisRaftTimeout("migrate command was not successful")
+
+            # In case of error, we just update the leader node and start from
+            # beginning. If leader has changed, our current 'cursor' value
+            # is not valid on the new leader.
```
tezc, line 49: I realized one thing and I think we didn't discuss this before. On leader change, admin should reset `cursor` value. (it should start from beginning). 

`cursor` iteration should be valid on the same node only IIUC. Not sure if ngc was aware of this as well. 

@sjpotter @fadidahanna wdyt?

fadidahanna (approved):

PR #578, opened 2023-02-26
small refactor for acls, extract user creation/get into function

sjpotter:


fadidahanna (approved):

fadidahanna (changes requested):
Needs to fix 2 small style issues (see CI failure)

fadidahanna (approved):

PR #584, opened 2023-03-08
[RR-227] Blocking Commands Support

sjpotter:
Add blocking command support through the new RM_Call Blocking/Promise infrastructure in Redis.

Blocking commands are similar to regular commands, except for the fact that they don't execute right away and can possibly timeout.

To support the blocking nature, we use Redis's new RM_Call blocking/promise infrastructure

To support timing out, the leader enforces timeouts via a timer on itself and will append a timeout log entry if the timer is triggered.

If a new leader comes around, all pending blocked commands that haven't executed yet, are automatically aborted.

Pending blocked commands are serialized to the snapshot and reexecuted upon snapshot load in the order that they were initially executed.

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -642,7 +642,8 @@ static void timeoutBlockedCommand(RedisRaftCtx *rr, raft_entry_t *entry)
     raft_index_t idx;
     bool error; /* not used yet */
 
-    RedisModule_Assert(RaftRedisDeserializeTimeout(entry->data, entry->data_len, &idx, &error) != RR_OK);
+    int ret = RaftRedisDeserializeTimeout(entry->data, entry->data_len, &idx, &error);
+    RedisModule_Assert(ret == RR_OK);
 
     BlockedCommand *bc = getBlockedCommand(idx);
     if (!bc) {
@@ -651,7 +652,8 @@ static void timeoutBlockedCommand(RedisRaftCtx *rr, raft_entry_t *entry)
     }
 
     /* In future might have to handle abort failing, but for plain redis commands this is a panic condition */
-    RedisModule_Assert(RedisModule_CallReplyPromiseAbort(bc->reply, NULL) == REDISMODULE_OK);
+    ret = RedisModule_CallReplyPromiseAbort(bc->reply, NULL);
+    RedisModule_Assert(ret == REDISMODULE_OK);
 
     if (bc->req) {
         if ((strncasecmp(bc->command, "bzpopmin", 8) == 0) ||
--- a/src/serialization.c
+++ b/src/serialization.c
@@ -473,6 +473,7 @@ raft_entry_t *RaftRedisSerializeTimeout(raft_index_t idx, bool error)
 
     size_t sz = calcIntSerializedLen(idx); /* idx we are timing out */
     sz += calcIntSerializedLen(err_val);   /* encoding the error bool */
+    sz++;
 
     raft_entry_t *ety = raft_entry_new(sz);
     ety->type = RAFT_LOGTYPE_TIMEOUT_BLOCKED;
@@ -517,7 +518,7 @@ RRStatus RaftRedisDeserializeTimeout(const void *buf, size_t buf_size, raft_inde
     buf_size -= n;
     *error = (tmp == 1);
 
-    RedisModule_Assert(buf_size == 0);
+    RedisModule_Assert(buf_size == 1); /* should only have the final '\0' at the end of the data */
 
     return RR_OK;
 }
--- a/src/serialization_utils.c
+++ b/src/serialization_utils.c
@@ -93,7 +93,7 @@ int encodeInteger(char prefix, char *ptr, size_t sz, unsigned long val)
 {
     int n = snprintf(ptr, sz, "%c%lu\n", prefix, val);
 
-    if (n > (int) sz) {
+    if (n >= (int) sz) {
         return -1;
     }
     return n;


src/util.c
```diff
@@ -520,3 +520,52 @@ int syncRename(const char *oldname, const char *newname)
...
+{
+    RedisModule_Assert(cmds->len == 1);
+
+    RaftRedisCommand *cmd = cmds->commands[0];
+    if (cmd->argc < 2) {
+        return REDISMODULE_ERR;
+    }
+
+    if (zero == NULL) {
+        zero = RedisModule_CreateString(NULL, "0", 1);
```
tezc, line 536: Maybe we can create a struct of pointers in RedisRaftCtx and init them RedisRaftCtxInit() then we can use like
```
rr->shared.zero;
rr->shared.one;
```
similar to redis.

src/util.c
```diff
@@ -520,3 +520,52 @@ int syncRename(const char *oldname, const char *newname)
...
+    }
+
+    size_t cmd_len;
+    const char *cmd_str = RedisModule_StringPtrLen(cmd->argv[0], &cmd_len);
+    if (strncasecmp(cmd_str, "brpop", 5) == 0 ||
+        strncasecmp(cmd_str, "blpop", 5) == 0 ||
+        strncasecmp(cmd_str, "bzpopmin", 5) == 0 ||
+        strncasecmp(cmd_str, "bzpopmax", 5) == 0 ||
```
tezc, line 544:
```suggestion
        strncasecmp(cmd_str, "bzpopmin", 8) == 0 ||
        strncasecmp(cmd_str, "bzpopmax", 8) == 0 ||
```

src/redisraft.c
```diff
@@ -757,8 +760,28 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
...
+             */
+            RedisModule_ReplyWithError(ctx, "ERR timeout is not a float or out of range");
+            return;
+        }
+        /* redis commands take timeout in seconds, timer takes time in milliseconds) */
+        timeout = timeout * 1000;
+
+        req = RaftReqInitBlocking(ctx, timeout);
+        RaftRedisCommandArrayMove(&req->r.redis.cmds, cmds);
+        req->r.redis.cmds.blocking = true;
```
tezc, line 780: Are you going to use this variable anywhere?

src/redisraft.c
```diff
@@ -12,6 +12,7 @@
 #include "common/sc_crc32.h"
 
 #include <dlfcn.h>
+#include <execinfo.h>
```
tezc, line 15:
```suggestion
```

I assume this is a leftover?

src/blocked.c
```diff
@@ -0,0 +1,149 @@
...
+ * the Server Side Public License v1 (SSPLv1).
+ */
+
+#include "redisraft.h"
+
+#include <string.h>
+
+/* A list of all current blocked commands */
+
+BlockedCommand *addBlockedCommand(raft_index_t idx, unsigned long long session, const char *data, size_t data_len, RaftReq *req, RedisModuleCallReply *reply)
```
tezc, line 13:
```suggestion
BlockedCommand *addBlockedCommand(raft_index_t idx, unsigned long long session, const char *data, size_t data_len, RaftReq *req)
```

Looks like you don't use reply here

src/blocked.c
```diff
@@ -0,0 +1,149 @@
...
+}
+
+
+void freeBlockedCommand(BlockedCommand *bc)
+{
+    if (bc->data) {
+        RedisModule_Free(bc->data);
+        bc->data = NULL;
+    }
+    RedisModule_Free(bc);
```
tezc, line 51:
```suggestion
    RedisModule_Free(bc->data);
    RedisModule_Free(bc);
```
You can avoid null check if you want.

src/redisraft.h
```diff
@@ -859,6 +881,9 @@ int fsyncFile(int fd);
 int fsyncFileAt(const char *path);
 void fsyncDir(const char *path);
 int syncRename(const char *oldname, const char *newname);
+RaftRedisCommandArray *RaftRedisBlockToNonBlockCmds(RaftRedisCommandArray *cmds);
+RedisModuleString **RaftRedisExtractBlockedKeys(RaftRedisCommandArray *cmds, size_t *count);
```
tezc, line 885:
```suggestion
RaftRedisCommandArray *RaftRedisBlockToNonBlockCmds(RaftRedisCommandArray *cmds);
RedisModuleString **RaftRedisExtractBlockedKeys(RaftRedisCommandArray *cmds, size_t *count);
```
seems like these are leftovers

src/raft.c
```diff
@@ -1632,9 +1715,33 @@ void RaftReqFree(RaftReq *req)
...
+    entry->type = RAFT_LOGTYPE_TIMEOUT_BLOCKED;
+    memcpy(entry->data, &req->raft_idx, entry->data_len);
+    entryAttachRaftReq(rr, entry, req);
+
+    int e = raft_recv_entry(rr->raft, entry, NULL);
+    if (req && e != 0) {
+        /* TODO: unsure if we should reply to user here, like in other case.
+         * I'd think in this case, it would be the timer went off and we aren't leader anymore,
+         * which will be more cleanly handled upon new term
+         */
```
tezc, line 1738: It might be better to reply here. If we can't elect a new leader, client will stay blocked forever.
Still, we can append successfully and lose leadership before replicating/applying the entry. Then, client will stay blocked.

Not sure if this is important but maybe something to think about. Wdyt?

src/redisraft.c
```diff
@@ -1225,7 +1249,7 @@ static int cmdRaftCluster(RedisModuleCtx *ctx, RedisModuleString **argv, int arg
 
         clusterInit(cluster_id);
 
-        char reply[RAFT_DBID_LEN + 256];
+        char reply[RAFT_DBID_LEN + 260];
```
tezc, line 1252: did you suspect overflow around here :)

src/redisraft.c
```diff
@@ -742,7 +743,9 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
      * we're still a leader. Otherwise, just process the reads. */
     if (cmd_flags & CMD_SPEC_READONLY && !(cmd_flags & CMD_SPEC_WRITE)) {
         if (!rr->config.quorum_reads) {
-            RaftExecuteCommandArray(rr, ctx, ctx, cmds);
+            RaftReq req = {0};
+            req.ctx = ctx;
```
tezc, line 747:
```suggestion
            RaftReq req = {
                .ctx = ctx
            };
```
I don't know which one you like more but you can use this version if you want, just fyi.

src/util.c
```diff
@@ -520,3 +520,76 @@ int syncRename(const char *oldname, const char *newname)
...
 }
+
+int findTimeoutIndex(RaftRedisCommand *cmd) {
+    size_t cmd_len;
+    const char *cmd_str = RedisModule_StringPtrLen(cmd->argv[0], &cmd_len);
+
+    if (strncasecmp(cmd_str, "brpop", 5) == 0 ||
+        strncasecmp(cmd_str, "blpop", 5) == 0 ||
+        strncasecmp(cmd_str, "bzpopmin", 5) == 0 ||
+        strncasecmp(cmd_str, "bzpopmax", 5) == 0 ||
```
tezc, line 531:
```suggestion
        strncasecmp(cmd_str, "bzpopmin", 8) == 0 ||
        strncasecmp(cmd_str, "bzpopmax", 8) == 0 ||
```

src/blocked.c
sjpotter: use it now

src/blocked.c
sjpotter: I think I'd rather have it in case in future its possible to have a BlockedCommand struct with a null data element

src/redisraft.h
```diff
@@ -650,6 +659,16 @@ typedef struct RaftReq {
     } r;
 } RaftReq;
 
+typedef struct BlockedCommand {
+    raft_index_t idx;
+    unsigned long long session;
```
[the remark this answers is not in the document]
sjpotter: done

src/redisraft.h
sjpotter: must have removed already

src/raft.c
sjpotter: so I think its actually wrong to reply here, and updated the commen.

If we can't append, three things could happen
 
this entry has been comitted, so is pending at timeout time.
1) has in practice executed before term change, therefore doesn't matter
2) hasn't executed at term change, will be aborted by term change
 
this entry hasn't been comitted
 
3) entry will be popped and aborted via pop of log.

src/redisraft.c
sjpotter: I was having compiler complaining, wasn't sure why, but you had same issue.

src/redisraft.h
```diff
@@ -515,8 +519,10 @@ typedef struct {
     bool asking;                  /* if this command array is in an asking mode */
     int size;                     /* Size of allocated array */
     int len;                      /* Number of elements in array */
+    bool blocking;                /* this is a converted blocking command */
     RaftRedisCommand **commands;
     RedisModuleString *acl;
+    bool timed_out;
```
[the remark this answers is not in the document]
sjpotter: yes, removed

src/raft.c
```diff
@@ -1654,6 +1761,21 @@ RaftReq *RaftReqInit(RedisModuleCtx *ctx, enum RaftReqType type)
     return req;
 }
 
+RaftReq *RaftReqInitBlocking(RedisModuleCtx *ctx, size_t timeout)
```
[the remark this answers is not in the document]
sjpotter: so I did this. but I question it now, makes things ugly.  I'm leaning towards a 

RaftReqInit/RaftReqInitBlocked that both call a RaftReqInitCore

src/redisraft.c
```diff
@@ -757,9 +782,28 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
         return;
     }
 
-    RaftReq *req = RaftReqInit(ctx, RR_REDISCOMMAND);
+    long long timeout = 0;
+    bool blocking = false;
+    if (cmd_flags & CMD_SPEC_BLOCKING) { /* protect against blocking commands in a MULTI above */
+        blocking = true;
+        long double tmp;
+        if (RaftRedisExtractBlockingTimeout(cmds, &tmp) != REDISMODULE_OK) {
```
fadidahanna, line 790: Suggest combining `RaftRedisExtractBlockingTimeout` with `validTimeout`, same as redis does at `getTimeoutFromObjectOrReply`. Anyway, if you still want to keep them both,`validTimeout` should be renamed, because it also replies to the client.

src/raft.c
```diff
@@ -1640,14 +1756,48 @@ void RaftReqFree(RaftReq *req)
...
+RaftReq *RaftReqInit(RedisModuleCtx *ctx, enum RaftReqType type, long long timeout)
+{
+    bool blocking = timeout >= 0 ? true : false;
+
     RaftReq *req = RedisModule_Calloc(1, sizeof(RaftReq));
     if (ctx != NULL) {
-        req->client = RedisModule_BlockClient(ctx, NULL, NULL, NULL, 0);
+        req->client_id = RedisModule_GetClientId(ctx);
+        if (blocking) {
+            req->client = RedisModule_BlockClient(ctx, NULL, NULL, NULL, 0);
```
fadidahanna, line 1797: There is no different in the behavior between the two cases...

src/raft.c
```diff
@@ -1656,6 +1806,10 @@ RaftReq *RaftReqInit(RedisModuleCtx *ctx, enum RaftReqType type)
         redis_raft.migrate_req = req;
     }
 
+    if (timeout > 0) {
```
fadidahanna, line 1809: Suggest keeping `RaftReqInit()` as a constructor, and creating the timer outside of this function.

src/util.c
```diff
@@ -520,3 +521,89 @@ int syncRename(const char *oldname, const char *newname)
     fsyncDir(oldname);
     return RR_OK;
 }
+
+int findTimeoutIndex(RaftRedisCommand *cmd)
```
fadidahanna, line 525: I'm not sure this function should be under `utils`. For now, it's relevant only for blocking commands. suggest moving it to `blocked.c`, rename it so it will be easy to understand that this function handles blocking commands only, and abort for unknown commands.

src/redisraft.h
```diff
@@ -430,6 +430,11 @@ typedef struct RedisRaftCtx {
     RedisModuleDict *locked_keys;         /* keys that have been locked for migration */
     RedisModuleDict *acl_dict;            /* maps acl strings to RedisModuleUser * objects */
     RedisModuleDict *client_session_dict; /* maps session IDs to Session Objects */
+
+    /* we use a dict and an intrusive list to reproduce java's LinkedHashMap, fast lookup with order maintenance */
+    struct sc_list blocked_command_list;   /* list of blocked commands in order of them blocking */
+    RedisModuleDict *blocked_command_dict; /* raft entry id -> blocked command mapping, for fast lookup */
+    RedisModuleDict *blocked_clients;      /* mapping blocked clients to blocked commands to timeout on disconnect or client unblock */
```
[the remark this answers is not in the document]
sjpotter: not needed at moment, it was from people playing around with trying to timeout on disconnect (and also client unblock)

src/raft.c
```diff
@@ -343,21 +343,46 @@ RedisModuleUser *RaftGetACLUser(RedisModuleCtx *ctx, RedisRaftCtx *rr, RaftRedis
...
 
+void handleUnblock(RedisModuleCtx *ctx, RedisModuleCallReply *reply, void *private_data)
+{
+    UNUSED(ctx);
+    BlockedCommand *bc = (BlockedCommand *) private_data;
+    /*
+     * the below lookup is duplicative, could have allocated memory just for idx, but that's a waste,
+     * so just use existing allocated object
+     */
+    bc = getBlockedCommand(bc->idx);
```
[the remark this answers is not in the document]
sjpotter: this was an artifact and we i had a single getAndDelete function.  when I separated them, this was no longer needed.

src/raft.c
```diff
@@ -408,31 +444,47 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
         /* Explanation:
          * When we have an ACL, we will have a user set on the context, so need "C"
          */
-        char *resp_call_fmt = cmds->acl ? "CE0v" : "E0v";
+        char *resp_call_fmt;
+        if (cmds->len > 1) {
```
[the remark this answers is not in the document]
sjpotter: it would be something like cmds->commands[0].argv[0] == MULTI, can look into add something.

src/raft.c
```diff
@@ -366,13 +391,24 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
...
-    ClientSession *client_session = getClientSession(rr, cmds, reply_ctx != NULL);
+    ClientSession *client_session = getClientSession(rr, cmds, req != NULL);
     (void) client_session; /* unused for now */
 
+    if (cmds->cmd_flags & CMD_SPEC_BLOCKING) {
+        // LOG_WARNING("replacing timeout with zero at apply time");
+        if (RaftRedisReplaceBlockingTimeout(cmds) != REDISMODULE_OK) {
+            // LOG_WARNING("failed to replace timeout with zero at apply time");
+            if (req) {
+                RedisModule_ReplyWithError(req->ctx, "Failed to replace blocking command timeout with zero at apply time");
```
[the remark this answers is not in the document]
sjpotter: we actually probably want to panic, as at apply time, any changes in behavior is bad, so panic.

src/raft.c
sjpotter: as of now there isn't, it depends if we need a timeout callback for client unblock.

src/raft.c
sjpotter: so I didn't like that as I felt there should be symmetry between RaftReqInit and RaftReqFree.  One setups the timer, the other can tear it down.

src/blocked.c
```diff
@@ -0,0 +1,145 @@
...
+    BlockedCommand *blocked = NULL;
+    RedisModule_DictDelC(redis_raft.blocked_command_dict, &idx, sizeof(idx), &blocked);
+    if (blocked == NULL) {
+        return;
+    }
+
+    sc_list_del(&redis_raft.blocked_command_list, &blocked->blocked_list);
+}
+
+BlockedCommand *getBlockedCommand(raft_index_t idx)
```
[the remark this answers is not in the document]
sjpotter: not sure this is necessary at the moment.

src/redisraft.c
```diff
@@ -757,9 +782,28 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
...
+        if (validTimeout(ctx, tmp, &timeout) != RR_OK) {
+            return;
+        }
+    }
+
+    RaftReq *req = RaftReqInit(ctx, RR_REDISCOMMAND, timeout);
     RaftRedisCommandArrayMove(&req->r.redis.cmds, cmds);
 
+    if (blocking) {
+        req->null_reply_type = getNullReplyType(&req->r.redis.cmds);
```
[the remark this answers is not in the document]
sjpotter: mostly because we don't have the command list at that point in a usable form, its just in a serialized form.

src/raft.c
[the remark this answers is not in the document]
sjpotter: added a CMDSPEC of type multi that we can check

src/raft.c
sjpotter: I removed it for now with my just pushed change for RaftReqInit()

src/redisraft.c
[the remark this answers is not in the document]
sjpotter: if you want it there, we can probably save the command as part of the bc.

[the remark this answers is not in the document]
sjpotter: ok, did that.  saved the command.

src/blocked.c
```diff
@@ -0,0 +1,253 @@
...
+#include <math.h>
+#include <string.h>
+#include <strings.h>
+
+/* A list of all current blocked commands */
+
+BlockedCommand *allocBlockedCommand(const char *cmd_name, raft_index_t idx, raft_session_t session, const char *data, size_t data_len, RaftReq *req, RedisModuleCallReply *reply)
+{
+    BlockedCommand *bc = RedisModule_Calloc(1, sizeof(BlockedCommand));
+    bc->command = RedisModule_Strdup(cmd_name);
```
tezc, line 18: Is this guaranteed to be null terminated? As far as I can see you get it from RedisModuleString which may not have null terminator.

src/blocked.c
```diff
@@ -0,0 +1,253 @@
...
+void addBlockedCommand(BlockedCommand *bc)
+{
+    sc_list_init(&bc->blocked_list);
+    sc_list_add_tail(&redis_raft.blocked_command_list, &bc->blocked_list);
+    RedisModule_DictSetC(redis_raft.blocked_command_dict, &bc->idx, sizeof(bc->idx), bc);
+}
+
+void deleteBlockedCommand(raft_index_t idx)
+{
+    BlockedCommand *blocked = NULL;
```
tezc, line 39:
```suggestion
    BlockedCommand *blocked = NULL;

```
Please leave one empty line after variable declarations

src/blocked.c
```diff
@@ -0,0 +1,253 @@
...
+}
+
+BlockedCommand *getBlockedCommand(raft_index_t idx)
+{
+    return RedisModule_DictGetC(redis_raft.blocked_command_dict, &idx, sizeof(idx), NULL);
+}
+
+void clearAllBlockCommands()
+{
+    struct sc_list *elem;
```
tezc, line 72:
```suggestion
    struct sc_list *elem;

```

src/blocked.c
sjpotter: so my experience is that command names are actually null terminated C strings, its things that can be binary (i.e. keys/values) that are not C strings and hence length matters

src/blocked.c
```diff
@@ -0,0 +1,253 @@
...
+    bc->data_len = data_len;
+    bc->req = req;
+    bc->reply = reply;
+
+    return bc;
+}
+
+void addBlockedCommand(BlockedCommand *bc)
+{
+    sc_list_init(&bc->blocked_list);
```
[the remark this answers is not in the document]
sjpotter: agree

src/blocked.c
```diff
@@ -0,0 +1,253 @@
...
+        raft_session_t session = RedisModule_LoadUnsigned(rdb);
+        size_t data_len;
+        char *data = RedisModule_LoadStringBuffer(rdb, &data_len);
+
+        /* execute blocking command again */
+        RaftRedisCommandArray tmp = {0};
+        if (RaftRedisCommandArrayDeserialize(&tmp, data, data_len) != RR_OK) {
+            PANIC("Invalid Raft entry");
+        }
+        tmp.client_id = session;
```
tezc, line 134:
```suggestion
        RaftRedisCommandArray tmp = {
            .client_id = session,
        };
        if (RaftRedisCommandArrayDeserialize(&tmp, data, data_len) != RR_OK) {
            PANIC("Invalid Raft entry");
        }
```
You can move it into initializer if you want

src/blocked.c
tezc, line 39:
```suggestion
    BlockedCommand *blocked = NULL;

```

src/raft.c
```diff
@@ -343,21 +343,41 @@ RedisModuleUser *RaftGetACLUser(RedisModuleCtx *ctx, RedisRaftCtx *rr, RaftRedis
...
+
+    BlockedCommand *bc = private_data;
+    if (bc->req) {
+        RedisModule_ReplyWithCallReply(bc->req->ctx, reply);
+        RaftReqFree(bc->req);
+    }
+
+    RedisModule_FreeCallReply(reply);
+    deleteBlockedCommand(bc->idx);
+    freeBlockedCommand(bc);
```
tezc, line 358: seems like you can combine these two

src/raft.c
```diff
@@ -343,21 +343,41 @@ RedisModuleUser *RaftGetACLUser(RedisModuleCtx *ctx, RedisRaftCtx *rr, RaftRedis
...
+}
+
 /* Execute all commands in a specified RaftRedisCommandArray.
  *
  * If reply_ctx is non-NULL, replies are delivered to it.
  * Otherwise, no replies are delivered.
+ *
+ * Now handles blacking commands (ex: brpop) as well.
```
tezc, line 366:
```suggestion
 * Now handles blocking commands (ex: brpop) as well.
```

src/raft.c
```diff
@@ -343,21 +343,41 @@ RedisModuleUser *RaftGetACLUser(RedisModuleCtx *ctx, RedisRaftCtx *rr, RaftRedis
...
+ * If a blocking command is issued, we execute normally, but now use the return value of RM_Call()
+ * to indicate if we areblocking (i.e. should the req not be freed).  If RM_Call returns a promise, we
```
tezc, line 368:
```suggestion
 * to indicate if we are blocking (i.e. should the req not be freed). If RM_Call returns a promise, we
```

src/raft.c
```diff
@@ -408,31 +432,47 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
...
+            /* don't block inside a MULTI */
+            resp_call_fmt = cmds->acl ? "CE0v" : "E0v";
+        } else {
+            resp_call_fmt = cmds->acl ? "KCE0v" : "KE0v";
+        }
 
         enterRedisModuleCall();
         RedisModule_SetContextUser(ctx, user);
         reply = RedisModule_Call(ctx, cmd, resp_call_fmt, &c->argv[1], c->argc - 1);
+        RedisModule_SetContextUser(ctx, NULL);
```
tezc, line 446: Why do we need this?

src/raft.c
```diff
@@ -599,6 +637,40 @@ static void clearClientSessions(RedisRaftCtx *rr)
     rr->client_session_dict = RedisModule_CreateDict(rr->ctx);
 }
 
+static void timeoutBlockedCommand(RedisRaftCtx *rr, raft_entry_t *entry)
+{
+    RedisModule_Assert(entry->data_len == sizeof(raft_index_t));
+    raft_index_t idx;
+    memcpy(&idx, entry->data, sizeof(idx));
```
tezc, line 644:
```suggestion
    RedisModule_Assert(entry->data_len == sizeof(raft_index_t));

    raft_index_t idx;
    memcpy(&idx, entry->data, sizeof(idx));
```
memcpy looks like a variable to my eyes without a single empty line

src/raft.c
tezc: I think I agree with Hanna here RaftReqInit() as a constructor is good. Generally, I prefer single init function when it is possible, I think it makes reading easier when you have a single place that you set variables. 

I guess you all didn't like timer set in RaftReqInit()? I thought it was good.

src/redisraft.c
tezc: yeah it was false positive

src/redisraft.h
```diff
@@ -625,6 +631,9 @@ typedef struct RaftReq {
     int type;
     RedisModuleBlockedClient *client;
     RedisModuleCtx *ctx;
+    RedisModuleTimerID timeout_timer;
+    raft_index_t raft_idx;
+    unsigned long long client_id;
```
tezc, line 636:
```suggestion
    raft_session_t client_id;
```

src/serialization.c
```diff
@@ -135,6 +139,12 @@ raft_entry_t *RaftRedisCommandArraySerialize(const RaftRedisCommandArray *source
     p += n;
     sz -= n;
 
+    /* Encode cmd_flags */
+    n = encodeInteger('*', p, sz, source->cmd_flags);
```
tezc, line 143: Actually, serialization here was using multi-bulk format with `\n` termination rather than `\r\n`. 
`*` indicates elem count in the list, so this is not correct but I see we already broke it. 
I guess we can revisit this later and fix it. Maybe we can combine some flags and make it more compact as well.

src/raft.c
```diff
@@ -366,13 +386,17 @@ void RaftExecuteCommandArray(RedisRaftCtx *rr,
...
+    if (handleSharding(rr, req ? req->ctx : NULL, cmds) != RR_OK) {
+        return NULL; /* sharding error, so even if blocking command, don't */
     }
 
-    ClientSession *client_session = getClientSession(rr, cmds, reply_ctx != NULL);
+    ClientSession *client_session = getClientSession(rr, cmds, req != NULL);
     (void) client_session; /* unused for now */
 
+    if (cmds->cmd_flags & CMD_SPEC_BLOCKING) {
+        RedisModule_Assert(RaftRedisReplaceBlockingTimeout(cmds) == REDISMODULE_OK);
```
tezc, line 397: As far as I can see, you don't use return value of this function. So, maybe we can assert things inside of it. 
I

tests/integration/test_blocking.py
```diff
@@ -0,0 +1,304 @@
+from _pytest.python_api import raises
+
+from redis.exceptions import ConnectionError
+from .sandbox import RawConnection
+
+
+def test_brpop(cluster):
```
tezc, line 7: maybe you can have some tests for bad timeout values, does redis-suite run those tests? Still, I think we can have a one test here.

src/redisraft.h
```diff
@@ -519,6 +523,7 @@ typedef struct {
     bool asking;                  /* if this command array is in an asking mode */
     int size;                     /* Size of allocated array */
     int len;                      /* Number of elements in array */
+    size_t cmd_flags;             /* the calculated cmd_flags for all commands in this array */
```
tezc, line 526: I think `size_t` gives wrong impression here for flags (feels like it is the length of an array, number of items etc.).
maybe we can use `unsigned long` or `unsigned int`.

src/raft.c
sjpotter: I have 2 raftreqinit's now, one with timer / one without.  the problem is, 0 is a valid timeout, so need to pass -1 to non blocking ones to distinguish them if want to do anything different.  I think it makes their calls look ugly as timeout means nothing to them.

I'm content the way it is

src/blocked.c
sjpotter: ok

src/raft.c
sjpotter: I wanted to make them parallel alloc/add

src/raft.c
sjpotter: after executing the commnd, we want to reset state to be safe.

src/raft.c
tezc: No, my point was if this node won't see a new term (cluster is down, this node becomes isolated etc), it will never pop the entry and it won't unblock client. 

I guess clients should be prepared for these anyway. Probably either user will set a timeout or client will apply it is own operation timeout on its local.

src/raft.c
sjpotter: ok, I wanted them next to each other as they go together, its basically idx being initalized.  perhaps one could say something like 

`raft_idx_t idx = *(raft_idx_t *) entry->data`

will that (or something like it) work.

or perhaps just 

`raft_index_t *idx = entry->data`

and then just use `*idx` elsewhere in the code?

tests/integration/test_blocking.py
sjpotter: redis-suite does those tests, so that's why I skipped it.  but I guess.

src/redisraft.h
sjpotter: ok

src/raft.c
```diff
@@ -599,6 +637,40 @@ static void clearClientSessions(RedisRaftCtx *rr)
...
+
+    if (bc->req) {
+        switch (getNullReplyType(bc->command)) {
+            case REDISRAFT_NULL_ARRAY:
+                RedisModule_ReplyWithNullArray(bc->req->ctx);
+                break;
+            default:
+                RedisModule_ReplyWithNull(bc->req->ctx);
+                break;
+        }
```
tezc, line 666:
```suggestion
        if ((strncasecmp(bc->command, "bzpopmin", 8) == 0) ||
            (strncasecmp(bc->command, "bzpopmax", 8) == 0) ||
            (strncasecmp(bc->command, "bzmpop", 6) == 0)) {
            RedisModule_ReplyWithNullArray(bc->req->ctx);
        } else {
            RedisModule_ReplyWithNull(bc->req->ctx);
        }

```

I think we can move that function here to keep logic local. Also, this way we can avoid defining things this way.

src/raft.c
```diff
@@ -1640,18 +1741,49 @@ void RaftReqFree(RaftReq *req)
...
+void blockedTimedOut(RedisModuleCtx *ctx, void *data)
+{
+    RedisRaftCtx *rr = &redis_raft;
+    RaftReq *req = data;
+
+    /* don't need to attach the req to this entry, as the req is part of the BlockedClient object */
+    raft_entry_t *entry = raft_entry_new(sizeof(raft_index_t));
+    entry->id = rand();
+    entry->type = RAFT_LOGTYPE_TIMEOUT_BLOCKED;
+    memcpy(entry->data, &req->raft_idx, entry->data_len);
```
tezc, line 1757: One thing that slightly bothers me here is we use binary serialization and it won't be readable on file. 
We use string serialization normally (only exception is RaftCfgChange unfortunately). 

Here it looks easier to switch to string serialization as it is just a number (probably you can use `decodeInteger()/encodeInteger()`. Maybe we should stay consistent and use string serialization here and we can do it for RaftCfgChange later?

src/raft.c
```diff
@@ -599,6 +637,40 @@ static void clearClientSessions(RedisRaftCtx *rr)
...
+    memcpy(&idx, entry->data, sizeof(idx));
+
+    BlockedCommand *bc = getBlockedCommand(idx);
+    if (!bc) {
+        /* unblock handler called before timeout was applied */
+        return;
+    }
+
+    if (RedisModule_CallReplyPromiseAbort(bc->reply, NULL) != REDISMODULE_OK) {
+        /* shouldn't happen with normal redis commands */
```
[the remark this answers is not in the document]
sjpotter: it's part of the API, and it could happen in future with module based blocking commands (ala what we do in redisraft, wouldn't be able to abort us), but I think you're right in this case that it should be an assert.

src/raft.c
sjpotter: ok

src/raft.c
sjpotter: my point is that we shouldn't have 2 code paths that WILL be executed that respond to the client.  While we could respond here, it means we have to code to prevent the other 2 possibilities from happening which is ugly.

src/blocked.c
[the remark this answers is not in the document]
sjpotter: done

src/blocked.c
sjpotter: done

src/blocked.c
sjpotter: done

src/blocked.c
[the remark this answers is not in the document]
tezc: I think you didn't update the Pr

src/raft.c
sjpotter: done

src/raft.c
sjpotter: done

src/serialization.c
tezc: Opened a ticket, https://redislabs.atlassian.net/browse/RR-232
We can revisit it later.

src/raft.c
tezc: ok

src/redisraft.h
sjpotter: not needed yet, so just removing for now

src/redisraft.c
```diff
@@ -757,7 +764,14 @@ static void handleRedisCommandAppend(RedisRaftCtx *rr,
...
 
-    RaftReq *req = RaftReqInit(ctx, RR_REDISCOMMAND);
+    long long timeout = 0;
+    if (cmd_flags & CMD_SPEC_BLOCKING) { /* protect against blocking commands in a MULTI above */
+        if (RaftRedisExtractBlockingTimeout(ctx, cmds, &timeout) != RR_OK) {
+            return;
+        }
+    }
+
+    RaftReq *req = RaftReqInitBlocking(ctx, RR_REDISCOMMAND, timeout);
```
fadidahanna, line 774: Why no to run `RaftReqInitBlocking()` only for `CMD_SPEC_BLOCKING` case?

tests/integration/test_blocking.py
```diff
@@ -0,0 +1,304 @@
...
+    cluster.execute("lpush", "x", 3)
+
+    cluster.wait_for_unanimity()
+
+    for i in range(1, 3):
+        val = cluster.node(i).raft_debug_exec("lrange", "x", 0, -1)
+        assert val == [b'3', b'2', b'1']
+
+
+def test_blocking_with_key_reaname(cluster):
```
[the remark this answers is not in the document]
sjpotter: done

src/raft.c
sjpotter: ok, adjusted

tests/integration/test_blocking.py
sjpotter: done, copied what redis tests do.

src/redisraft.c
sjpotter: that's a mistake on my part.

sjpotter: code changed to 

```
    RaftReq *req;
    if (cmd_flags & CMD_SPEC_BLOCKING) { /* protect against blocking commands in a MULTI above */
        long long timeout = 0;
        if (RaftRedisExtractBlockingTimeout(ctx, cmds, &timeout) != RR_OK) {
            return;
        }
        req = RaftReqInitBlocking(ctx, RR_REDISCOMMAND, timeout);
    } else {
        req = RaftReqInit(ctx, RR_REDISCOMMAND);
    }
    RaftRedisCommandArrayMove(&req->r.redis.cmds, cmds);
```

src/blocked.c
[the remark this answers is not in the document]
sjpotter: correct, its in my tree, will push all the small changes together

src/raft.c
sjpotter: ok, changing this to be relative to your comment re serialized below.  I'm not going to get the multibulk right, so could use feedback on that.

src/raft.c
sjpotter: so, the reason I didn't go to this right now is that I'm thinking of a general metadata parsing, so that we don't have to strncasecmp all the time for them.  With that said, can do that later.

src/blocked.c
```diff
@@ -0,0 +1,253 @@
...
+    }
+    if (bc->data) {
+        RedisModule_Free(bc->data);
+        bc->data = NULL;
+    }
+    if (bc->reply) {
+        RedisModule_FreeCallReply(bc->reply);
+        bc->reply = NULL;
+    }
+    RedisModule_Free(bc);
```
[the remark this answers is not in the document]
sjpotter: ok, done

src/raft.c
tezc: I think we can fix multibulk format later. For now, let's just try to serialize things as strings.

src/redisraft.h
```diff
@@ -649,6 +658,20 @@ typedef struct RaftReq {
     } r;
 } RaftReq;
 
+#define REDISRAFT_NULL_NONE  1
+#define REDISRAFT_NULL_ARRAY 2
```
fadidahanna, line 662: No use - please remove

src/redisraft.h
```diff
@@ -965,4 +995,16 @@ void ClientStateFree(RedisRaftCtx *rr, unsigned long long client_id);
...
+BlockedCommand *allocBlockedCommand(const char *cmd_name, raft_index_t idx, raft_session_t session, const char *data, size_t data_len, RaftReq *req, RedisModuleCallReply *reply);
+void addBlockedCommand(BlockedCommand *bc);
+void freeBlockedCommand(BlockedCommand *bc);
+void deleteBlockedCommand(raft_index_t idx);
+BlockedCommand *getBlockedCommand(raft_index_t idx);
+void blockedCommandsSave(RedisModuleIO *rdb);
+void blockedCommandsLoad(RedisModuleIO *rdb);
+void clearAllBlockCommands();
+int RaftRedisExtractBlockingTimeout(RedisModuleCtx *ctx, RaftRedisCommandArray *cmds, long long *timeout);
+void RaftRedisReplaceBlockingTimeout(RaftRedisCommandArray *cmds);
```
fadidahanna, line 1008: Not really sure why we need RaftRedis prefix for these two commands.

sjpotter: ok

src/raft.c
```diff
@@ -599,6 +637,37 @@ static void clearClientSessions(RedisRaftCtx *rr)
     rr->client_session_dict = RedisModule_CreateDict(rr->ctx);
 }
 
+static void timeoutBlockedCommand(RedisRaftCtx *rr, raft_entry_t *entry)
+{
+    raft_index_t idx;
+    bool error; /* not used yet */
+
+    RedisModule_Assert(RaftRedisDeserializeTimeout(entry->data, entry->data_len, &idx, &error) != RR_OK);
```
tezc, line 645: I'm confused by this ` != RR_OK`. Seems like we expect this to fail? 

Btw, could you move this function outside of assert. If something is in assert, it gives impression that it is just there to validate things.

sjpotter: you're right.  ok.  I'll assert on the response.  that's probably in a few places.

tezc (approved):

fadidahanna:
Please change the header and add a description

fadidahanna (approved):

PR #587, opened 2023-03-11
[RR-209] Snapshot metrics

tezc:
This PR can be reviewed commit by commit.

- First commit makes debug configs actual configs that are set by 
`raft.config` command. IIRC this was @fadidahanna's idea, I now realize
 that we can use `REDISMODULE_CONFIG_HIDDEN` for these configs, so they
 won't be listed when we call `config get raft*`. 

- Second commit adds last index and last term of ongoing snapshot 
process to the info string.

### DIFF
--- a/src/redisraft.c
+++ b/src/redisraft.c
@@ -1386,7 +1386,7 @@ static int cmdRaftDebug(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
 
         if (argc > 2) {
             if (RedisModule_StringToLongLong(argv[2], &async) != REDISMODULE_OK) {
-                RedisModule_ReplyWithError(ctx, "ERR invalid compact fail value");
+                RedisModule_ReplyWithError(ctx, "ERR invalid compact async value");
                 return REDISMODULE_OK;
             }
         }
--- a/src/redisraft.c
+++ b/src/redisraft.c
@@ -1342,14 +1342,10 @@ static int cmdRaftShardGroup(RedisModuleCtx *ctx, RedisModuleString **argv, int 
     }
 }
 
-/* RAFT.DEBUG COMPACT [delay] [fail] [async]
+/* RAFT.DEBUG COMPACT [async]
  *     Initiate the snapshot.
- *     If [delay] is specified, introduce an artificial delay of [delay] seconds
- *        in the background child process.
- *     If [fail] is non-zero, snapshot will fail.
  *     If [async] is non-zero, snapshot will be triggered and client will get
- *        the reply without waiting the snapshot result. [delay] ahd [fail]
- *        parameters will be ignored.
+ *        the reply without waiting the snapshot result.
  * Reply:
  *   +OK
  *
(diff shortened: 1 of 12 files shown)


src/snapshot.c
```diff
@@ -367,7 +354,7 @@ RRStatus initiateSnapshot(RedisRaftCtx *rr)
     rr->curr_snapshot_last_idx = rr->snapshot_info.last_applied_idx;
     rr->curr_snapshot_last_term = rr->snapshot_info.last_applied_term;
     rr->snapshot_in_progress = true;
-    rr->curr_snapshot_start_time = RedisModule_Milliseconds();
+    rr->curr_snapshot_start_time = RedisModule_MonotonicMicroseconds();
```
tezc, line 357: moved to monotonic timestamp, just to prevent possible negative diff on wall clock adjustments

src/redisraft.h
```diff
@@ -388,27 +396,21 @@ typedef struct RedisRaftCtx {
                                                     it will be renamed to the original rdb file */
     SnapshotLoad snapshot_load;          /* Indicates we've received a snapshot waiting to be loaded */
     bool snapshot_in_progress;           /* Indicates we're creating a snapshot in the background */
-    mstime_t curr_snapshot_start_time;   /* start time of the snapshot operation currently in progress */
     raft_index_t curr_snapshot_last_idx; /* Last included idx of the snapshot operation currently in progress */
     raft_term_t curr_snapshot_last_term; /* Last included term of the snapshot operation currently in progress */
-    mstime_t last_snapshot_time;         /* Total time (ms) of a last local snapshot operation */
+    uint64_t curr_snapshot_start_time;   /* Start time of the snapshot operation currently in progress */
```
tezc, line 401: As this is a monotonic timestamp now, its type has to be uint64_t

src/snapshot.c
```diff
@@ -883,3 +866,21 @@ void archiveSnapshot(RedisRaftCtx *rr)
...
+
+    /* Delete if there is a partial snapshot file from the previous run */
+    int ret = unlink(rr->incoming_snapshot_file);
+    if (ret != 0 && errno != ENOENT) {
+        PANIC("Unlink file: %s, error: %s \n", rr->incoming_snapshot_file,
+              strerror(errno));
+    }
+
+    resetSnapshotState(rr);
+}
```
tezc, line 886: moved a few pieces here, this function is to be called on module load once.

src/redisraft.c
```diff
@@ -1908,14 +1795,18 @@ static void handleInfo(RedisModuleInfoCtx *ctx, int for_crash_report)
...
+    if (rr->snapshots_created != 0) {
+        snapshot_time = (long long) rr->last_snapshot_time;
     }
     RedisModule_InfoAddFieldLongLong(ctx, "snapshot_time_secs", snapshot_time);
+
     RedisModule_InfoAddFieldULongLong(ctx, "snapshots_created", rr->snapshots_created);
     RedisModule_InfoAddFieldULongLong(ctx, "snapshots_received", rr->snapshots_received);
     RedisModule_InfoAddFieldCString(ctx, "snapshot_in_progress", rr->snapshot_in_progress ? "yes" : "no");
+    RedisModule_InfoAddFieldLongLong(ctx, "snapshot_in_progress_last_idx", rr->curr_snapshot_last_idx);
+    RedisModule_InfoAddFieldLongLong(ctx, "snapshot_in_progress_last_term", rr->curr_snapshot_last_term);
```
tezc, line 1809: please comment about field names, maybe there are better names for these?

src/snapshot.c
fadidahanna: I wonder why - can you elaborate a little bit?

tezc: https://github.com/RedisLabs/redisraft/pull/587/files#diff-d31a5eb8cc6de9b16ed710fd09c8a8516aa375c97f345fa98adf28ab4bf5b65eL2071

We initialize these metrics with `-1` there in `RedisRaftCtxInit()`. We also have `resetSnapshotState()`. So, I thought we may want a single init function. Metrics will be initialized/cleared by only `resetSnapshotState()`

Also, we have `initSnapshotTransferData()` function, so we can move that one here as well and call once on module load. My intention was to have a single place to init snapshot related variables.

fadidahanna (approved):

PR #591, opened 2023-03-20
track count of raft.import errors in info raft

LLiuJJ:


src/migrate.c
```diff
@@ -108,6 +108,7 @@ void importKeys(RedisRaftCtx *rr, raft_entry_t *entry)
         RedisModule_Assert(reply != NULL);
 
         if (RedisModule_CallReplyType(reply) == REDISMODULE_REPLY_ERROR) {
+            rr->importkeys_restore_failed_count++;
```
tezc, line 111: There is an assert on line 115, so it will abort anyway once we are here. 
fyi, `RedisModule_Assert()` is a runtime check, compiler won't strip it out like regular `assert()`. 

Btw I think this is an important error, it should not ever happen. Have you hit this error somehow? I guess it can only happen if there is redis version mismatch.

LLiuJJ: I just took a look at the implementation of the restore command. When a very serious data error is encountered, an error will be returned. You are right, it is necessary to interrupt the program.

tezc: Yes, because importkeys command will run on all nodes in the destination cluster. If one node succeeds and other fails, nodes can diverge. So, we cannot allow this and that's why we assert here.

PR #592, opened 2023-03-25
[RR-225] - extend command.c for subcommands of command containers

sjpotter:
currentlly we only support specifying top level commands, but if we filter on "client", we can't pass through "client id" while filtering "client unblock". This allows us to distinguish them.


sjpotter:
Note: it probably be nice for initialize to have better support for sub commands, but that wasn' for now.

src/commands.c
```diff
@@ -348,22 +375,52 @@ RRStatus CommandSpecTableSetC(CommandSpecTable *cmd_spec_table, void *key, size_
 /* Look up the specified command in the command spec table and return the
  * CommandSpec associated with it, or NULL.
  */
-const CommandSpec *CommandSpecTableGet(CommandSpecTable *cmd_spec_table, const RedisModuleString *cmd)
+int CommandSpecTableGetFlags(CommandSpecTable *cmd_spec_table, RedisModuleDict *sub_command_tables, const RedisModuleString *cmd, const RedisModuleString *subcmd)
```
fadidahanna, line 378: Why not to keep `CommandSpecTableGet` and return the relevant cs?

src/commands.c
```diff
@@ -348,22 +375,52 @@ RRStatus CommandSpecTableSetC(CommandSpecTable *cmd_spec_table, void *key, size_
...
+        return -1;
+    }
+
+    int flags = cs->flags;
+
+    if (!(cs->flags & CMD_SPEC_SUBCOMMAND) || subcmd == NULL) {
+        return flags;
+    }
+
+    cmd_spec_table = RedisModule_DictGetC(sub_command_tables, cs->name, strlen(cs->name), NULL);
```
fadidahanna, line 392: please create a local var

src/commands.c
```diff
@@ -348,22 +375,52 @@ RRStatus CommandSpecTableSetC(CommandSpecTable *cmd_spec_table, void *key, size_
...
+    if (cmd_spec_table == NULL) {
+        return flags;
+    }
+
+    /* flags are now for subcommand, so mask it out of flags */
+    flags &= ~CMD_SPEC_SUBCOMMAND;
```
fadidahanna, line 398: What is the difference between this case and the one above? in both cases we get a command with a subcommand.

src/commands.c
sjpotter: we can't return the cs, as my change returns a modified flags (mask out the subcommand flag).  And didn't have a use for the whole cs struct, it was only ever used for the flags from this call.

src/commands.c
sjpotter: done

src/commands.c
fadidahanna: Yes, and I wonder about `CMD_SPEC_SUBCOMMAND` too - see my other comment

src/commands.c
sjpotter:
```
    if (!(cs->flags & CMD_SPEC_SUBCOMMAND) || subcmd == NULL) {
        return flags;
    }
```

either the command we got is not a subcommand container or we didn't get a subcommand for it, therefore return its flags as is.  This will be the very common case

```
    CommandSpecTable *sub_cmd_spec_table = RedisModule_DictGetC(sub_command_tables, cs->name, strlen(cs->name), NULL);
    if (sub_cmd_spec_table == NULL) {
        return flags;
    }
```

it is a subcommand container, and has a subcommand, but, we didn't create a subcommand table for it, therefore return its flags as is (though perhaps can return it masked, as it is a subcommand, that would make sense to me

```
    /* flags are now for subcommand, so mask it out of flags */
    flags &= ~CMD_SPEC_SUBCOMMAND;
```

for subcommands (i.e. my note above), we mask out the the SUBCOMMND part, the rest is the default values returned for subcommands not specified

```
    CommandSpec *sub_cs = getOrCreateCommandSpec(sub_cmd_spec_table, subcmd, false);
    if (sub_cs == NULL) {
        return flags;
    }
```

look up the subcommand, if its not specified, returned default (now masked) value.

```
    return sub_cs->flags;
```
otherwise return value defined for the subcommand.

sjpotter: added the above notes as comments in the func

fadidahanna:
@sjpotter - test-style failed

fadidahanna (approved):

PR #593, opened 2023-03-26
[RR-225] add "CLIENT UNBLOCK" and timeout blocking when disconnect

sjpotter:
this add support for the CLIENT UNBLOCK command, so that we can send a timeout log entry when the user issues it, and will reply to the user correctly if we were able to unblock or not

in addition, it provides a best effort attempt at timing out entries when the blocked client disconnects.  However, it is possible that the blocked entry will execute before the timeout entry is applied, as is expected.


sjpotter:
passes redis test suite with fewer skipped tests

https://github.com/RedisLabs/redisraft/actions/runs/4524381347

src/redisraft.c
```diff
@@ -555,6 +555,66 @@ static void handleAsking(RedisRaftCtx *rr, RedisModuleCtx *ctx)
     RedisModule_ReplyWithSimpleString(ctx, "OK");
 }
 
+static void handleClientUnblock(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRedisCommand *cmd)
+{
+    raft_session_t id;
+    bool error = false;
+
+    if (cmd->argc != 3 && cmd->argc != 4) {
+        RedisModule_ReplyWithError(ctx, "ERR syntax error");
```
tezc, line 564: Can we use `RM_WrongArity()` here.

src/clientstate.c
```diff
@@ -46,4 +60,4 @@ void MultiStateReset(MultiState *multi_state)
 void ClientStateReset(ClientState *client_state)
 {
     MultiStateReset(&client_state->multi_state);
-}
+}
```
tezc, line 63:
```suggestion
}

```

src/redisraft.c
```diff
@@ -555,6 +555,66 @@ static void handleAsking(RedisRaftCtx *rr, RedisModuleCtx *ctx)
...
+        return;
+    }
+
+    if (cmd->argc == 4) {
+        size_t str_len;
+        const char *str = RedisModule_StringPtrLen(cmd->argv[3], &str_len);
+
+        if (!strcasecmp(str, "timeout")) {
+            error = false;
+        } else if (!strcasecmp(str, "error")) {
```
tezc, line 580: probably we should use strncasecmp for these

src/redisraft.c
```diff
@@ -555,6 +555,66 @@ static void handleAsking(RedisRaftCtx *rr, RedisModuleCtx *ctx)
...
+    }
+
+    raft_entry_t *entry = RaftRedisSerializeTimeout(cs->blocked_req->raft_idx, error);
+    RaftReq *req = RaftReqInit(ctx, RR_CLIENT_UNBLOCK);
+    entryAttachRaftReq(rr, entry, req);
+
+    int e = raft_recv_entry(rr->raft, entry, NULL);
+    if (e != 0) {
+        RedisModule_ReplyWithLongLong(req->ctx, 0);
+        RaftReqFree(req);
```
tezc, line 601: Did you test this path? I think this will crash if you don't call `entryDetachRaftReq()`.because `raft_entry_release()` will call `entryFreeAttachedRaftReq()` and it will try to use `req` if I'm not mistaken.

src/raft.c
```diff
@@ -639,35 +640,57 @@ static void clearClientSessions(RedisRaftCtx *rr)
...
-    bool error; /* not used yet */
+    bool error;
 
     int ret = RaftRedisDeserializeTimeout(entry->data, entry->data_len, &idx, &error);
     RedisModule_Assert(ret == RR_OK);
 
     BlockedCommand *bc = getBlockedCommand(idx);
     if (!bc) {
         /* unblock handler called before timeout was applied */
+        RedisModule_ReplyWithLongLong(req->ctx, 0);
```
tezc, line 654: If I'm not mistaken, we won't have req on replay? A test would be nice for these race conditions if possible.

src/raft.c
```diff
@@ -639,35 +640,57 @@ static void clearClientSessions(RedisRaftCtx *rr)
...
-    RedisModule_Assert(ret == REDISMODULE_OK);
 
-    if (bc->req) {
-        if ((strncasecmp(bc->command, "bzpopmin", 8) == 0) ||
-            (strncasecmp(bc->command, "bzpopmax", 8) == 0) ||
-            (strncasecmp(bc->command, "bzmpop", 6) == 0)) {
-            RedisModule_ReplyWithNullArray(bc->req->ctx);
-        } else {
-            RedisModule_ReplyWithNull(bc->req->ctx);
+    if (ret == REDISMODULE_OK) {
```
tezc, line 662: So, can it return REDISMODULE_ERR now?

src/redisraft.c
```diff
@@ -1677,8 +1743,15 @@ void handleClientEvent(RedisModuleCtx *ctx, RedisModuleEvent eid,
                  * We can only append if we are leader, if we're not leader anymore,
                  * then this would have already been cleaned up.
                  */
-                if (cs && cs->watched && rr->raft && raft_is_leader(rr->raft)) {
-                    appendEndClientSession(rr, NULL, ci->id, "disconnect");
+                if (cs) {
+                    if (cs->watched && rr->raft && raft_is_leader(rr->raft)) {
+                        appendEndClientSession(rr, NULL, ci->id, "disconnect");
+                    }
+                    if (cs->blocked_req && rr->raft && raft_is_leader(rr->raft)) {
```
tezc, line 1750: `rr->raft && raft_is_leader(rr->raft)` seems common in these if checks and you can move back to the top if check if you want.

src/redisraft.c
sjpotter: its not what clientCommand returns.  It actually calls addReplySubcommandSyntaxError() which is what I was hinting to myself there, need to fix it up to be

"unknown subcommand or wrong number of arguments for '%.128s'. Try %s HELP"

where I should make "ERR unknown subcommand or wrong number of arguments for 'unblock'.  Try client HELP"

src/redisraft.c
sjpotter: yes

src/redisraft.c
sjpotter: yes, there should be a return after the RaftReqFree()

src/raft.c
sjpotter: yes, it needs to be protected with an if (req) {}  (this is the client unblock req)

src/raft.c
sjpotter: no real difference from before, just trying to be accurate to API.

src/raft.c
```diff
@@ -639,35 +640,57 @@ static void clearClientSessions(RedisRaftCtx *rr)
...
         }
-        RaftReqFree(bc->req);
+
+        deleteBlockedCommand(bc->idx);
+        freeBlockedCommand(bc);
     }
 
-    deleteBlockedCommand(bc->idx);
-    freeBlockedCommand(bc);
+    if (req) {
```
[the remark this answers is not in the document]
sjpotter: ?  this is that the req exists or not, don't reply if it doesn't exist (i.e. replay)  type shouldn't matter (have a case I missed that ozan mentioned above)

src/redisraft.c
sjpotter: well, redis does strcasecmp().

src/raft.c
sjpotter: this test would be to client unblock, and then restart the node to test that replay works as expected and doesn't crash.  (possibly to push also after restart and validate that it didn't get popped)

sjpotter: I'm actually unsure how to test get this race condition to work. i.e. to have the blocked command execute before the timeout.  

Only way I can say forcing it is via a debug command that issues 2 entries in succession (basically atomically to redis), first entry is an lpush, second entry is the timeout log entry.  since redis couldn't process them (apply) in between each other, its the closest to simulating the race condition.

src/redisraft.c
tezc: I don't think this is enough, please take a look at other usages. Probably, it is better to "return" in error case, something like: 

```c
if (e != 0) {
    RedisModule_ReplyWithLongLong(req->ctx, 0);
    raft_entry_release(entry);
    RaftReqFree(req);
    return;
}

```

src/redisraft.c
```diff
@@ -555,6 +555,67 @@ static void handleAsking(RedisRaftCtx *rr, RedisModuleCtx *ctx)
     RedisModule_ReplyWithSimpleString(ctx, "OK");
 }
 
+static void handleClientUnblock(RedisRaftCtx *rr, RedisModuleCtx *ctx, RaftRedisCommand *cmd)
+{
+    raft_session_t id;
+    bool error = false;
+
+    if (cmd->argc != 3 && cmd->argc != 4) {
+        RedisModule_ReplyWithError(ctx, "ERR unknown subcommand or wrong number of arguments for '%.128s'. Try %s HELP");
```
tezc, line 564: you forgot to replace format identifiers

src/raft.c
tezc: > I'm actually unsure how to test get this race condition to work. i.e. to have the blocked command execute before the timeout.

Maybe you can use 3 seconds timeout for the blocking command and then submit a lpush from another client

sjpotter:
redis suite - https://github.com/RedisLabs/redisraft/actions/runs/4544049415

src/common.c
```diff
@@ -27,6 +27,14 @@ const char *getStateStr(RedisRaftCtx *rr)
     }
 }
 
+void replyZero(RedisModuleCtx *ctx, const char *msg, int err)
+{
+    UNUSED(msg);
+    UNUSED(err);
+
+    RedisModule_ReplyWithLongLong(ctx, 0);
+}
```
tezc, line 36: not needed anymore?

src/raft.c
tezc: How about 
- blpop 3 seconds
- config set raft.log-disable-apply yes
- lpush
- wait at least 3 seconds so timeout entry will be appended
- config set raft.log-disable-apply no

src/common.c
sjpotter: done

src/common.c
```diff
@@ -281,3 +289,21 @@ void shutdownServer(RedisRaftCtx *rr)
...
+    if (e != 0) {
+        if (req) {
+            entryDetachRaftReq(rr, entry);
+        }
+    }
+
+    raft_entry_release(entry);
+
+    return e;
+}
```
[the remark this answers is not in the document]
sjpotter: done

src/raft.c
sjpotter: oh, didnt know you could do that.  disable apply that is :)  that would work, basically the same as what I pushed already.  perhaps a one off debug command isn't what we want, so can move to your approach.

tests/integration/test_blocking.py
```diff
@@ -315,3 +318,100 @@ def test_blocking_with_invalid_timeout(cluster):
...
+    c1.send_command("client", "id")
+    id = c1.read_response()
+
+    c1.send_command("brpop", "x", 0)
+
+    cluster.wait_for_unanimity()
+    time.sleep(1)
+
+    for i in range(1, 3):
+        cluster.node(i).execute("config", "set", "raft.log-disable-apply", "yes")
```
tezc, line 401:
```suggestion
    cluster.config_set("raft.log-disable-apply", "yes")
```

tests/integration/test_blocking.py
```diff
@@ -315,3 +318,100 @@ def test_blocking_with_invalid_timeout(cluster):
...
+
+    c2.send_command("lpush", "x", 1)
+    c3.send_command("client", "unblock", id)
+
+    for i in range(1, 3):
+        cluster.node(i).execute("config", "set", "raft.log-disable-apply", "no")
```
tezc, line 407:
```suggestion
    cluster.config_set("raft.log-disable-apply", "no")
```

tezc (approved):

PR #594, opened 2023-03-26
remove cluster-mode flag from redis suite run

sjpotter:
skip less tests, but have to skip one eval oom test because of different semantics in rediraft


sjpotter:
this might depend on my client unblock test, as only tested with it at the moment, will run redis suite in ci now to see if it works without it as well

sjpotter:
passes redis suite - https://github.com/RedisLabs/redisraft/actions/runs/4524954048

tezc (approved):

PR #597, opened 2023-03-29
Do not allow node config change before log replay

tezc:
On restart, RedisRaft should not allow node add/remove before replaying
all the log entries. Because, there might be add/remove log entry in the
log already and before replay, node does not know if those entries are
committed and applied in the previous run. So, node may reject new
add/remove operations with "one voting change only" error, it might be
confusing for the users. Instead, it should return -CLUSTERDOWN.

We already do this for regular operations and reject them before log 
replay but that check was missing for node add/remove operations.

----
This is the check in libraft to prevent multiple voting changes at  the same time:
https://github.com/RedisLabs/redisraft/blob/7cdad3a1bfb1b7ccd23942820d6f6edafa55c9a2/deps/raft/src/raft_server.c#L1131

This is where we clear voting change flag (when we apply the entry):
https://github.com/RedisLabs/redisraft/blob/7cdad3a1bfb1b7ccd23942820d6f6edafa55c9a2/deps/raft/src/raft_server.c#L1263

### DIFF
--- a/src/redisraft.c
+++ b/src/redisraft.c
@@ -103,6 +103,11 @@ static int cmdRaftNode(RedisModuleCtx *ctx, RedisModuleString **argv, int argc)
         return REDISMODULE_OK;
     }
 
+    if (raft_get_current_term(rr->raft) != raft_get_last_applied_term(rr->raft)) {
+        replyClusterDown(ctx);
+        return REDISMODULE_OK;
+    }
+
     size_t cmd_len;
     const char *cmd = RedisModule_StringPtrLen(argv[1], &cmd_len);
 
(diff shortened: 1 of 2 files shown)


sjpotter:
I'm wondering if this is correct.  i.e. the test is 

```
if (raft_get_current_term(rr->raft) != raft_get_last_applied_term(rr->raft))
```

But couldn't this mean we could get into a degenerate case where we can never accept `raft.node` commands (i.e. our latest entry is never the same term as the last applied one)

it doesn't seem to functionally change things in terms of what can be done, before we errored out with only one in progress, now we are erroring out with clusterdown.  Is there a better way to just prevent it from accepting messages during log replay?

perhaps this is me just not understanding something.

fadidahanna:
@sjpotter , any objections after Ozan's last comment?

sjpotter:
I just dont understand the point.  I dont understand why RedisRaft would accept any commands before the log is replayed.

i.e. the log is loaded while RedisRaft is in the REDIS_RAFT_LOADING state, not in REDIS_RAFT_UP state.

If that's that case, why do we have to be concerned about new entries?

PR #599, opened 2023-04-02
[RR-224] Add tests with big snapshots

tezc:
This PR adds two tests with large data-sets (~1.7 Gb on ram, ~800 mb RDB file on disk)
Goal is to test stability when dealing with large data-sets. 
Both tests have 10 clients running operations in the background. 

Tests added:
- test_fuzzing_with_large_snapshots: randomly add/remove/restart nodes
- test_stability_with_snapshot_delivery: Same leader node delivers 800 mb snapshot to the follower in a loop

As these tests are slow, only added to daily to run with `--runslow` pytest option.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -31,7 +31,7 @@ def pytest_addoption(parser):
         '--work-dir', default='tests/tmp',
         help='Working directory for tests temporary files.')
     parser.addoption(
-        '--redis-up-timeout', default=3,
+        '--redis-up-timeout', default=120,
         help='Seconds to wait for Redis to start before timing out.')
     parser.addoption(
         '--valgrind', default=False, action='store_true',
--- a/tests/integration/sandbox.py
+++ b/tests/integration/sandbox.py
@@ -708,10 +708,13 @@ class Cluster(object):
         return self.nodes[self.leader]
 
     def follower_node(self):
-        for _id, node in self.nodes.items():
+        if len(self.nodes) <= 1:
+            raise RedisRaftError('No followers in the cluster')
+
+        while True:
+            _id = random.choice(list(self.nodes.keys()))
             if _id != self.leader:
-                return node
-        raise RedisRaftError('No followers in the cluster')
+                return self.node(_id)
 
     def pause_leader(self) -> int:
         old_leader = self.leader
--- a/tests/integration/sandbox.py
+++ b/tests/integration/sandbox.py
@@ -299,7 +299,7 @@ class RedisRaft(object):
             try:
                 self.client.ping()
                 return
-            except redis.exceptions.ConnectionError:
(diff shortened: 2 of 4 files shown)


sjpotter:
is this just adding tests?  it seems its replacing tests from the diff?

tezc:
Somehow github shows larger diff.. I moved deleted test to another commit, reviewing commit by commit might be easier. 

Currently, we don't run tests marked with `python.mark.slow`. 
There are two tests with these marks:
- test_stability_with_snapshots_and_restarts: We'll still keep it, made minor changes. 
- test_proxy_stability_under_load: We already have a similar test: `test_fuzzing_with_proxy_multi_and_restarts`. I deleted this one.

fadidahanna:
@tezc I'm not sure why `test_fuzzing_with_proxy_multi_and_restarts` and `test_proxy_stability_under_load` are similar - in `test_fuzzing_with_proxy_multi_and_restarts` we randomly restart nodes and verify correctness, while `test_proxy_stability_under_load` only test stability.

tests/integration/conftest.py
```diff
@@ -31,7 +31,7 @@ def pytest_addoption(parser):
         '--work-dir', default='tests/tmp',
         help='Working directory for tests temporary files.')
     parser.addoption(
-        '--redis-up-timeout', default=3,
+        '--redis-up-timeout', default=120,
```
fadidahanna, line 34: why not to keep the default and change it only for the slow run?

tests/integration/sandbox.py
```diff
@@ -704,6 +707,12 @@ def random_node(self):
     def leader_node(self):
         return self.nodes[self.leader]
 
+    def follower_node(self):
+        for _id, node in self.nodes.items():
```
fadidahanna, line 711: maybe we should add randomness?

tezc: done

tests/integration/conftest.py
tezc: I think it is better to use a config that works for every test suite. Then, it will be easier to reproduce issues on our local. (No need to remember to pass extra config parameters)

tezc:
@fadidahanna I think the problem with `test_proxy_stability_under_load`, it is archaic. It does one extra thing from the other test is this check: `commit_index >= prev_commit_index`. I think this doesn't make any sense these days, sure we have monotonic commit index. So, I think if it doesn't crash, it is stable enough. The other test is doing a similar thing, sending a big request, that's why I thought this is unnecessary.  

I reverted this change, keeping these tests. Maybe we can revisit them later, I see there are some other tests that requires a fix or better name/explanation in this file. (like `test_stale_reads_on_restarts`)

fadidahanna (approved):

PR #601, opened 2023-04-04
[RR-53] Add pubsub support

tezc:
This PR adds pubsub support to RedisRaft.

Differences from Redis Cluster:
- In Redis Cluster, clients can subscribe and publish to any node in the cluster. Publish messages will be forwarded to other shards and to the other nodes in the shard. 
In RedisRaft, we'll only allow publish/subscribe on the leader node. Messages will not be forwarded to the other shards. 

- Normally, Redis does not return -MOVED response to PUBLISH/SUBSCRIBE commands. RedisRaft will return -MOVED as it doesn't support subscribe/publish on the followers. 

Implementation details:
- If the node is the leader, allow SUBSCRIBE messages (do not intercept them). If not the leader, reject them with -MOVED or -CLUSTERDOWN response. 
- Process PUBLISH as a read-only command.
- PUBLISH can be inside of multi or lua script and might be written to the log. On log replay, we may process publish message again and if we have subscribers, they might receive the message. To prevent that, we won't allow subscribers until log replay is completed. 
- If the leader node becomes a follower, it will kill subscriber clients. Assumption is that those clients will try to connect back and they will get -MOVED or -CLUSTERDOWN response.

### DIFF
--- a/src/redisraft.c
+++ b/src/redisraft.c
@@ -1563,18 +1563,12 @@ static int cmdRaftRejectSubscribe(RedisModuleCtx *ctx, RedisModuleString **argv,
 
     RedisRaftCtx *rr = &redis_raft;
 
-    if (checkRaftState(rr, ctx) != RR_OK) {
+    if (checkRaftState(rr, ctx) != RR_OK ||
+        checkLeader(rr, ctx, NULL) != RR_OK) {
         return REDISMODULE_OK;
     }
 
-    if (!raft_is_leader(rr->raft)) {
-        Node *leader = getLeaderNodeOrReply(rr, ctx);
-        if (leader) {
-            redirectCommand(rr, ctx, leader);
-        }
-        return REDISMODULE_OK;
-    }
-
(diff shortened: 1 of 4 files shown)


src/redisraft.c
```diff
@@ -1613,6 +1613,44 @@ static int cmdRaftRandom(RedisModuleCtx *ctx,
...
+
+    if (!raft_is_leader(rr->raft)) {
+        Node *leader = getLeaderNodeOrReply(rr, ctx);
+        if (leader) {
+            redirectCommand(rr, ctx, leader);
+        }
+        return REDISMODULE_OK;
+    }
+
+    replyClusterDown(ctx);
```
fadidahanna, line 1650: maybe worth mentioning in a comment the case here - log replay is not competed yet

fadidahanna (dismissed):

src/redisraft.c
tezc: done

fadidahanna (approved):

PR #602, opened 2023-04-05
Fix big

tezc:


tests/integration/conftest.py
```diff
@@ -31,7 +31,7 @@ def pytest_addoption(parser):
         '--work-dir', default='tests/tmp',
         help='Working directory for tests temporary files.')
     parser.addoption(
-        '--redis-up-timeout', default=120,
+        '--redis-up-timeout', default=3,
```
tezc, line 34: Revert the change we made in https://github.com/RedisLabs/redisraft/pull/599

Some tests expect a node "down" after start-up. So, it is better to keep this value small, otherwise those tests take a long time.

tests/integration/test_fuzzing.py
```diff
@@ -365,12 +365,12 @@ def test_stability_with_snapshot_delivery(cluster, workload):
...
     # Add followers
-    cluster.add_node()
-    cluster.add_node()
+    cluster.add_node(use_cluster_args=True)
+    cluster.add_node(use_cluster_args=True)
 
     cluster.wait_for_info_param("raft_num_voting_nodes", 3, timeout=240)
     cluster.wait_for_unanimity(timeout=240)
-    cluster.config_set("raft.log-max-file-size", "16mb")
+    cluster.config_set("raft.log-max-file-size", "4mb")
```
tezc, line 373: Trigger snapshot more frequently to lower waiting time.

fadidahanna (approved):

Issue #608, opened 2023-04-13
Redis 7.0.10 crashed by signal: 11, si_code: 1


jerrytesting:
We found one crash while running redis-7.0.10 with the redisraft-cd52ee4. The following is the bug report:

```
2023-04-13 07:56:48 Jepsen starting  /opt/redis/redis-server --protected-mode no --bind 0.0.0.0 --dbfilename redis.rdb --loadmodule /opt/redis/redisraft.so --raft.loglevel debug --raft.log-filename raftlog.db --raft.log-max-file-size 32000 --raft.log-max-cache-size 1000000 --raft.follower-proxy yes
3438:C 13 Apr 2023 07:56:48.130 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
3438:C 13 Apr 2023 07:56:48.131 # Redis version=7.0.10, bits=64, commit=6523d019, modified=0, pid=3438, just started
3438:C 13 Apr 2023 07:56:48.131 # Configuration loaded
3438:M 13 Apr 2023 07:56:48.136 * Increased maximum number of open files to 10032 (it was originally set to 1024).
3438:M 13 Apr 2023 07:56:48.136 * monotonic clock: POSIX clock_gettime
3438:M 13 Apr 2023 07:56:48.139 * Running mode=standalone, port=6379.
3438:M 13 Apr 2023 07:56:48.139 # Server initialized
3438:M 13 Apr 2023 07:56:48.139 # WARNING Memory overcommit must be enabled! Without it, a background save or replication may fail under low memory condition. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect.
3438:M 13 Apr 2023 07:56:48.140 * <raft> RedisRaft version 255.255.255 [5b53a293]
3438:M 13 Apr 2023 07:56:48.177 * <raft> Raft module loaded, state is 'uninitialized'
3438:M 13 Apr 2023 07:56:48.177 * Module 'raft' loaded from /opt/redis/redisraft.so
3438:M 13 Apr 2023 07:56:48.178 * Ready to accept connections
3438:M 13 Apr 2023 07:56:49.138 * <raft> State change: Node is now a leader, term 1
3438:M 13 Apr 2023 07:56:49.138 * <raft> Cluster Membership: term:1 index:1 nodes: id=1958283344,voting=0,active=1,addr=-
[...]

PR #609, opened 2023-04-13
Better redis version check

tezc:
Check existence of the latest function that was added to redismodule.h to detect compability better.

Hopefully, it will print helpful warnings for the cases like: https://github.com/RedisLabs/redisraft/issues/608

### DIFF
--- a/src/redisraft.c
+++ b/src/redisraft.c
@@ -2171,10 +2171,11 @@ __attribute__((__unused__)) int RedisModule_OnLoad(RedisModuleCtx *ctx, RedisMod
                     REDISRAFT_VERSION, REDISRAFT_GIT_SHA1);
 
     const int MIN_SUPPORTED_REDIS_VERSION = 0x00070000;
-    if (!RMAPI_FUNC_SUPPORTED(RedisModule_GetServerVersion) ||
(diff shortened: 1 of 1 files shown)


src/redisraft.c
```diff
@@ -2171,10 +2171,11 @@ __attribute__((__unused__)) int RedisModule_OnLoad(RedisModuleCtx *ctx, RedisMod
                     REDISRAFT_VERSION, REDISRAFT_GIT_SHA1);
 
     const int MIN_SUPPORTED_REDIS_VERSION = 0x00070000;
```
fadidahanna, line 2173: We can update `MIN_SUPPORTED_REDIS_VERSION` too - to `0x00070200`

tezc: 7.2 is still RC. Looks like we don't update version numbers in RCs. See [link](https://github.com/redis/redis/blob/7.2/src/version.h)
It is still 7 bla bla. 

Maybe we should ignore this for now?

fadidahanna: k, good to know, in that case - sure, we can update later (just open a ticket so we will do it in the future, thanks)

fadidahanna (approved):

PR #613, opened 2023-04-29
Print logs on configuration change

tezc:
Currently, we don't print a log line on a configuration change on the leader node. When there is an error regarding config change, it is not possible to understand what/when happened from the logs. So, I think this is an important event that we should print cluster configuration after a change.

### DIFF
--- a/src/raft.c
+++ b/src/raft.c
@@ -1121,7 +1121,7 @@ static int raftNodeHasSufficientLogs(raft_server_t *raft, void *user_data, raft_
     Node *node = raft_node_get_udata(raft_node);
     RedisModule_Assert(node != NULL);
 
-    LOG_DEBUG("node:%d has sufficient logs, adding as voting node.", node->id);
+    LOG_NOTICE("node:%d has sufficient logs, adding as voting node.", node->id);
 
     raft_entry_req_t *entry = raft_entry_new(sizeof(RaftCfgChange));
     entry->id = rand();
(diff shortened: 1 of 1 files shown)


src/raft.c
```diff
@@ -1137,6 +1137,37 @@ static int raftNodeHasSufficientLogs(raft_server_t *raft, void *user_data, raft_
...
+        buf = catsnprintf(buf, &buflen, " id=%d,voting=%d,active=%d,addr=%s",
+                          raft_node_get_id(rn),
+                          raft_node_is_voting(rn),
+                          raft_node_is_active(rn),
+                          addr);
+    }
+
+    return buf;
+}
+
```
tezc, line 1170: I've just moved this function above to avoid writing a prototype.

src/raft.c
```diff
@@ -1121,7 +1121,7 @@ static int raftNodeHasSufficientLogs(raft_server_t *raft, void *user_data, raft_
     Node *node = raft_node_get_udata(raft_node);
     RedisModule_Assert(node != NULL);
 
-    LOG_DEBUG("node:%d has sufficient logs, adding as voting node.", node->id);
+    LOG_NOTICE("node:%d has sufficient logs, adding as voting node.", node->id);
```
tezc, line 1124: Default loglevel is "notice", let's print this line by default.

fadidahanna (approved):

PR #614, opened 2023-05-04
Delete cmocka

tezc:
Although we have cmocka in the project, currently we don't use its features.
Main thing that bothers me with cmocka is it swallows asserts/crashes to keep running other tests. So, when you run the test in IDE, debugger wouldn't stop at the point assert fails as cmocka swallows it. This makes debugging harder. 

This PR removes cmocka and basically uses regular `assert()` instead of cmocka's assert macros.


tests/unit/test_log.c
```diff
@@ -61,169 +42,198 @@ static void append_entry(Log *log, int id, const char *value)
...
-    assert_int_equal(LogAppend(log, e), RR_OK);
+    assert(LogAppend(log, e) == RR_OK);
     raft_entry_release(e);
 }
 
-static void test_log_random_access(void **state)
+static void test_log_random_access()
 {
-    Log *log = (Log *) *state;
+    Log log;
```
fadidahanna, line 51: The initialization and teardown sections are one of the advantages of using cmocka.... but in our case, since we don't use any other capability (although we can in other tests, and maybe in the future) I guess we can remove it for now.

fadidahanna (approved):

tests/unit/test_log.c
tezc: It's easy to add this if we want. Just a macro to call a function pointer before and after the test. 

```c
test_run(before_fn, test_fn, after_fn) \
    before_fn();
    test_fn();
    after_fn();
```

That's actually what cmocka does. 
Question is do we want it? I personally like when a test case has all the code it runs (no magic). That's why I didn't add it.

PR #623, opened 2023-05-16
[RR-261] Free applied entries from the entry cache

tezc:
Currently, we delete entries from the entry cache if memory usage goes over configured size. 
This is happening inside periodic(). 

We may even delete entries that are waiting to be applied. When that happens, if an entry has a RaftReq entry associated with it, we'll reply with -TIMEOUT response when we remove the entry from the cache and destroy it:
https://github.com/RedisLabs/redisraft/blob/103c9d5035ff74425b51ed983ff7deb4533c57d6/src/raft.c#L81

This is a bad user experience. Even though we'll apply the entry in a moment, we return -TIMEOUT reply.
This PR changes this behavior, adds a check to prevent deleting entries that are waiting to be applied.


fadidahanna:
@tezc what should be the expected behavior in case the memory that was freed is not enough?

tezc:
We'll be over the configured limit but this risk is already there, we call periodic() every 100 milliseconds, in case of a burst of commands before compacting the cache, we will be over the configured limit. I think cache compaction is a best-effort thing (soft-limit) rather than a hard limit. As we don't have control on all memory allocations in the system, trying to be strict is not possible.

If we are constantly way over the configured limit, it might be a problem. e.g. we have 1000 clients, each one of them are sending 5 mb entries non-stop, we may keep these entries in memory until they are applied and it will consume 5 gb. 
On the other hand, if we don't fix this, these requests will consume 5 gb memory until periodic() is called and they will get `-TIMEOUT` responses. 

I think this is a backpressure issue (missing in our impl). We should not allow new commands if you don't have capacity to handle them. 

These heavy load cases aside, probably the more common case is handling occasional big requests. e.g. entry cache size is 64 mb, you get lots of small requests but one request with 32mb size. If that one makes you go over the limit, then some entries will be evicted from the cache and they will get -TIMEOUT reply. This is how I discovered this issue.

fadidahanna (approved):

PR #626, opened 2023-05-25
[RR-271] Bug fixes and improvements for File implementation

tezc:
This PR includes three changes and can be reviewed commit by commit

1- Bug fix when readv() returns with partial data. 
    This is a bit hard to reproduce but it happens when a large buffer read from file. 
    Conditions to reproduce:
     - There must be some data in the internal buffer of FILE object.
     - readv() should return with partial read.

 2- An improvement. Previously, File impl used internal buffer in the struct. This is fine as long as you don't copy/move the struct. 
     If you do that, pointers that point to that buffer will be invalid after moving. (e.g. `wpos`, `rpos`, `rend` ). Switching to heap allocated buffer to prevent future confusion. 

3- Looks like `readv()` and `writev()` with larger than INT_MAX buffer sizes don't work on macOS. So, adding a change to use regular `read()` and `write()` on macOS.

### DIFF
--- a/src/file.c
+++ b/src/file.c
@@ -14,11 +14,16 @@
 #include <sys/uio.h>
 #include <unistd.h>
 
+#define FILE_BUFSIZE 4096
+
 void FileInit(File *file)
 {
+    char *buf = RedisModule_Alloc(FILE_BUFSIZE);
+
     *file = (File){
         .fd = -1,
-        .wpos = file->buf,
+        .buf = buf,
+        .wpos = buf,
     };
 }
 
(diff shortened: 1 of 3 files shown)


src/file.c
```diff
@@ -28,17 +33,25 @@ int FileTerm(File *file)
...
-        return RR_ERROR;
+        rc = RR_ERROR;
+        goto out;
     }
 
+out:
+    RedisModule_Free(file->buf);
+    *file = (File){
+        .fd = -1,
+        .buf = NULL,
```
fadidahanna, line 52: `wpos` will continue to point to an invalid address.

tezc: (File) { .fd = -1} syntax sets rest of the struct variables to zero. So, other pointers will be NULL. `.buf = NULL` was also unnecessary here but I guess l left it for documentation purposes (easier to see on a quick look).

fadidahanna: right, missed the assignment here to a new obj...

fadidahanna (approved):

PR #631, opened 2023-06-07
[RR-299] Fix flaky test_entry_cache_compaction

tezc:
There are occasional failures on macOS: [link](https://github.com/RedisLabs/redisraft/actions/runs/5183100862/jobs/9340595893)

Somehow, followers start an election and choose another leader. So, previous leader node loses leadership and fails to execute the command.

Logs indicates no other issue. I tried to run tests with more debug logs enabled, I can't see any other issue. Probably, slow instances are causing the problem as the test uses a large entry to be replicated. 

As as workaround, creating one less node for that test. 
Ran the test 200 times after the fix and all passed on MacOS: [link](https://github.com/tezc/redisraft/actions/runs/5195571619)


fadidahanna:
really strange behaviour...

fadidahanna (approved):

PR #640, opened 2023-07-13
Fix delayed blocked client handling

tezc:


fadidahanna:
OSS PR is now on hold. We currently have an issue with calling RM_Call during rdbload (see `blockedCommandsLoad`) since no execution unit is started (and not supposed to). One solution can be moving this part out of rdbload to `callRaftPeriodic()'.

